Issue #21, opened 2024-02-13
Use `cargo check-external-types` to avoid breaking changes in public APIs


lquerel:
`cargo-check-external-types` is a static analysis tool for Rust library authors to set and verify which types from other libraries are allowed to be are exposed in their public API. This is useful for ensuring that a breaking change to a dependency doesn't force a breaking change in the library that's using it.

https://crates.io/crates/cargo-check-external-types

Define a `allowed-external-types.toml` file in each public crate. Update the script in the build process to check the presence of this file in each crate.

lquerel:
See #23

PR #23, opened 2024-02-14
chore(build): add workspace and public API controls to ensure compliance with project policies.

lquerel:
Two controls are implemented in this PR.
- Public APIs are controlled with cargo-check-external-types (see #21).
- Each crate in the workspace must complies with the following set of project policies.
  - Crate name starting with `weaver_`
  - Presence of a `README.md` file
  - Presence of a `allowed-external-types.toml`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -115,6 +115,24 @@ jobs:
       - uses: Swatinem/rust-cache@v2
       - name: Check formatting
         run: cargo fmt --all -- --check
+  check-workspace:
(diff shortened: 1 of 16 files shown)


jsuereth (approved):
Is it too early to enforce non breaking changes?

Overall looks good

lquerel:
> Is it too early to enforce non breaking changes?
> 
> Overall looks good

Yes, it's too early. Once we have an approved Application Telemetry Schema (conceptually, components and resolved parts), then we should consider enforcing non-breaking changes.

PR #25, opened 2024-02-15
Integration tests to check the resolution process of the official semconv registry.

lquerel:
This PR tests the resolution process of the official semantic registry. This integration test doesn't intend to precisely check that each reference and 'extends' clause are properly resolved. The goal is to ensure that no errors are raised during the loading and resolution of the official registry.

Success criteria:
- All semconv files downloaded from the official semconv repo.
- The parsing process should not fail.
- The resolution process should not fail.
- No warn or error messages should be reported by the logger.

This PR serves as an example of a first integration test. More should follow.

Note: some refactoring has also been performed to use the term 'semconv registry' instead of 'semconv catalog' to be more aligned with OpenTelemetry wording.

### DIFF
--- a/crates/weaver_resolver/src/registry.rs
+++ b/crates/weaver_resolver/src/registry.rs
@@ -9,7 +9,7 @@ use weaver_resolved_schema::lineage::{FieldId, FieldLineage, GroupLineage, Resol
 use weaver_resolved_schema::registry::{Group, Registry, TypedGroup};
 use weaver_semconv::attribute::AttributeSpec;
 use weaver_semconv::group::ConvTypeSpec;
-use weaver_semconv::{GroupSpecWithProvenance, SemConvSpecs};
+use weaver_semconv::{GroupSpecWithProvenance, SemConvRegistry};
 
 use crate::attribute::AttributeCatalog;
 use crate::constraint::resolve_constraints;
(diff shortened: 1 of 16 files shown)


crates/weaver_resolver/src/registry.rs
```diff
@@ -380,9 +376,10 @@ mod tests {
             }
 
             let mut attr_catalog = AttributeCatalog::default();
-            let observed_registry = resolve_registry(
-                unresolved_registry_from_specs("https://semconv-registry.com", &sc_specs),
+            let observed_registry = resolve_semconv_registry(
                 &mut attr_catalog,
+                "https://semconv-registry.com",
```
jsuereth, line 381: Is there an http address or IP address we can use with guaranteed failure that won't ever be domain-squatted here?

tests/resolution_process.rs
```diff
@@ -0,0 +1,68 @@
...
+                panic!("Failed to resolve the official semantic convention registry, error: {e}");
+            });
+
+    // The number of semconv groups is fluctuating, so we can't check for a
+    // specific number, but we can check if there are any groups at all.
+    assert!(!resolved_registry.groups.is_empty());
+
+    // Check if the logger has reported any warnings or errors.
+    assert_eq!(log.warn_count(), 0);
+    assert_eq!(log.error_count(), 0);
```
jsuereth, line 60: Do you think warn/error is enough to catch all potential errors?

Should we spot check 1-2 "stable" semconv here?   If warn/error count is enough to catch all potential serialization issues, then I think it's better not to hardcode any specific item here, we can focus attribute resolution logic unit tests on that.

jsuereth (approved):
Overall looks good, two nits

crates/weaver_resolver/src/registry.rs
lquerel: For this unit test, I will use a local IP address. Thanks

tests/resolution_process.rs
lquerel: Multiple explicit panic! calls will be triggered in the test if the API returns any explicit errors, so the test does not solely rely on logs.

```rust
let semconv_specs = SchemaResolver::load_semconv_registry(
        SEMCONV_REGISTRY_URL.to_string(),
        Some(SEMCONV_REGISTRY_MODEL.to_string()),
        &cache,
        log.clone(),
    )
    .unwrap_or_else(|e| {
        panic!("Failed to load and parse the official semantic convention registry, error: {e}");
    });
```

The unwrap_or_else is where the error handling is done.

Issue #26, opened 2024-02-15
`weaver check registry` subcommand


lquerel:
This subcommand must load, parse, resolve, and run additional tests to validate a semantic convention registry. This command could be utilized in a CI/CD pipeline to ensure the registry's validity.

PR #30, opened 2024-02-17
Registry sub-commands

lquerel:
This PR sets up the registry command and defines a list of sub-commands (i.e., check, generate, resolve, search, stats). Only the `check` command is implemented (see #26).

All other existing commands have been placed behind the `experimental` feature flag so as not to appear by default.

A brief description of the semantic convention registry resolution process has been added to the `README.md` of the `weaver_resolver` crate. Unimplemented elements are also identified (i.e., `any_of`, `include`). The completion and full testing of the registry resolution process will be the subject of a future PR.


src/registry/mod.rs
```diff
@@ -0,0 +1,65 @@
...
+mod generate;
+mod resolve;
+mod search;
+mod stats;
+
+/// Parameters for the `registry` command
+#[derive(Debug, Args)]
+pub struct RegistryCommand {
+    /// Define the sub-commands for the `registry` command
+    #[clap(subcommand)]
```
jsuereth, line 26: Haven't interacted with Clap much, but this is really nice.

src/registry/search.rs
```diff
@@ -0,0 +1,20 @@
+// SPDX-License-Identifier: Apache-2.0
+
+//! Search a semantic convention registry.
+
+use clap::Args;
+
+/// Parameters for the `registry search` sub-command
+#[derive(Debug, Args)]
+pub struct SearchRegistry {
+    /// Git URL of the semantic convention registry
```
jsuereth, line 10: We should probably uniformly allow "git url" & "local path" or some denoatation of acceptable "location" arguments.  I've opened #31 for that.

src/registry/stats.rs
```diff
@@ -0,0 +1,54 @@
...
+use weaver_resolver::SchemaResolver;
+
+/// Parameters for the `registry stats` sub-command
+#[derive(Debug, Args)]
+pub struct StatsRegistry {
+    /// Local path or Git URL of the semantic convention registry.
+    pub registry: String,
+
+    /// Optional path in the Git repository where the semantic convention
+    /// registry is located
```
jsuereth, line 19: For my understanding, this is generally the `model` directory?

jsuereth (approved):

src/registry/mod.rs
lquerel: Yes Clap is really nice. There is also a support various shell completions and few others things that we need to enable.

src/registry/search.rs
lquerel: Thanks

src/registry/stats.rs
lquerel: I left this part configurable because in the future we could allow the creation of vendor or enterprise oriented registries.

lquerel: #32 created

Issue #8, opened 2024-02-10
Add more tests to validate the semconv registry resolution process


lquerel:
Right now it's not obvious that the existing tests cover all the various configurations observed in the existing official semconv registry.

The current tests are located in the `weaver_resolver` crate and more specifically in the registry module. 
The new data model used to represent the resolved registry is present in the `weaver_resolved_schema` crate.

Note: there is a previous attempt, in the code base, for this resolution process trying to resolve both the semconv registry and the telemetry schema. This part of the code will be removed soon.

PR #39, opened 2024-02-27
[Feature] Implement `any_of` and `include` constraints support

lquerel:
This PR finalizes the resolution process for a semantic convention registry. More specifically, the following resolution steps have been implemented:
- Resolution of the `include` constraints: attributes and `any_of` constraints of the group pointed by the `include` reference are imported into the group containing the reference.
- Resolution of the `any_of` constraints: all `any_of` constraints are checked.

Some tests have been implemented, however more exhaustive tests will be implemented in the future (see #8).

### DIFF
--- a/crates/weaver_resolved_schema/src/catalog.rs
+++ b/crates/weaver_resolved_schema/src/catalog.rs
@@ -3,7 +3,7 @@
 //! Defines the catalog of attributes, metrics, and other telemetry items
 //! that are shared across multiple signals in the Resolved Telemetry Schema.
 
-use crate::attribute::Attribute;
+use crate::attribute::{Attribute, AttributeRef};
 use crate::metric::Metric;
 use serde::{Deserialize, Serialize};
 use std::fmt::Debug;
@@ -30,4 +30,12 @@ pub enum Stability {
     Experimental,
     /// A stable definition.
     Stable,
+}
+
+impl Catalog {
+    /// Returns the attribute name from an attribute ref if it exists
+    /// in the catalog or None if it does not exist.
+    pub fn attribute_name(&self, attribute_ref: &AttributeRef) -> Option<String> {
+        self.attributes.get(attribute_ref.0 as usize).map(|attr| attr.name.clone())
+    }
 }
--- a/crates/weaver_resolved_schema/src/catalog.rs
+++ b/crates/weaver_resolved_schema/src/catalog.rs
@@ -35,9 +35,9 @@ pub enum Stability {
 impl Catalog {
     /// Returns the attribute name from an attribute ref if it exists
     /// in the catalog or None if it does not exist.
-    pub fn attribute_name(&self, attribute_ref: &AttributeRef) -> Option<String> {
+    pub fn attribute_name(&self, attribute_ref: &AttributeRef) -> Option<&str> {
         self.attributes
             .get(attribute_ref.0 as usize)
-            .map(|attr| attr.name.clone())
+            .map(|attr| attr.name.as_ref())
     }
 }
--- a/crates/weaver_resolved_schema/src/catalog.rs
+++ b/crates/weaver_resolved_schema/src/catalog.rs
@@ -36,6 +36,8 @@ impl Catalog {
     /// Returns the attribute name from an attribute ref if it exists
     /// in the catalog or None if it does not exist.
     pub fn attribute_name(&self, attribute_ref: &AttributeRef) -> Option<String> {
-        self.attributes.get(attribute_ref.0 as usize).map(|attr| attr.name.clone())
+        self.attributes
+            .get(attribute_ref.0 as usize)
+            .map(|attr| attr.name.clone())
     }
 }
--- a/crates/weaver_resolver/src/registry.rs
+++ b/crates/weaver_resolver/src/registry.rs
@@ -18,7 +18,7 @@ use crate::constraint::resolve_constraints;
 use crate::metrics::resolve_instrument;
 use crate::spans::resolve_span_kind;
 use crate::stability::resolve_stability;
-use crate::{Error, UnsatisfiedAnyOfConstraint};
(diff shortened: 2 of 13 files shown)


crates/weaver_resolved_schema/src/catalog.rs
```diff
@@ -31,3 +31,13 @@ pub enum Stability {
     /// A stable definition.
     Stable,
 }
+
+impl Catalog {
+    /// Returns the attribute name from an attribute ref if it exists
+    /// in the catalog or None if it does not exist.
+    pub fn attribute_name(&self, attribute_ref: &AttributeRef) -> Option<String> {
```
jsuereth, line 38: This is a bit of a rust design question.

I realize, as this is a command line tool, we're being a bit "free" with `clone` for now. However, when do you prefer `String` vs. `&str`?  This seems like an instance where just returning `&str` may be more ergonomic of an API.

crates/weaver_resolver/src/registry.rs
```diff
@@ -115,9 +133,97 @@ pub fn resolve_semconv_registry(
...
+    for group in registry.groups.iter() {
+        // Build a list of attribute names for the group.
+        let mut group_attr_names = HashSet::new();
+        for attr_ref in group.attributes.iter() {
+            let attr_name =
+                attr_name_index
+                    .get(attr_ref.0 as usize)
+                    .ok_or(Error::UnresolvedAttribute {
+                        attribute_ref: *attr_ref,
+                    })?;
```
jsuereth, line 173: `?` is sequential right now.  I don't think it's worth doing now, but we should look into attempting to find a way to detect errors in "parallel".  If we can, it's best to give users as many *valid* errors as we can at once.

crates/weaver_resolver/src/registry.rs
```diff
@@ -115,9 +133,97 @@ pub fn resolve_semconv_registry(
...
+    let mut any_of_constraints = vec![];
+    'outer: for constraint in constraints.iter() {
+        if constraint.any_of.is_empty() {
+            continue;
+        }
+        any_of_total += 1;
+
+        // Check if the group satisfies the `any_of` constraint.
+        any_of_constraints.push(constraint.any_of.clone());
+        for attr_name in constraint.any_of.iter() {
```
jsuereth, line 204: I think you can get rid of `'outer` if you do:

```
if constraint.any_of.iter().find(|name| !group_attr_names.contains(name)).is_some() {
  any_of_unsatisified += 1
}
```

If you need to know which isn't found, you could make a new error type to aggregate:

```
if let Some(attr) = constraint.any_of.iter().find(|name| !group_attr_names.contains(name)) {
  any_of_unsatisfied += 1;
  missing_attribute_errors.push(MissingAnyOfError::new(attr, constraint));
}
```

crates/weaver_resolved_schema/src/catalog.rs
lquerel: Good catch. In this context, the best practice is to return Option<&str> since the attribute name is owned by the catalog, and it's most likely that the catalog's lifetime will not be ephemeral on the caller's side. I will make this change. Additionally, I will check if there is a `Clippy` lint to enforce this best practice.

crates/weaver_resolver/src/registry.rs
lquerel: I agree, and I've already started to follow this approach in some other areas. However, it has not been implemented universally, particularly not here :-). It would probably be better to implement this sooner rather than later, so I will most likely work on it today.

crates/weaver_resolved_schema/src/catalog.rs
lquerel: I didn't find any Clippy lint to automatically check this.

crates/weaver_resolver/src/registry.rs
lquerel: Simplified the code to better handle this type of error

crates/weaver_resolver/src/registry.rs
lquerel: I improved the error handling to accumulate multiple errors.

crates/weaver_resolver/src/registry.rs
```diff
@@ -73,33 +79,44 @@ pub fn resolve_semconv_registry(
...
+                        provenance: group.provenance.clone(),
+                    });
+                }
+            }
         }
-        if !unresolved_refs.is_empty() {
-            return Err(Error::UnresolvedReferences {
-                refs: unresolved_refs,
-            });
+        if !errors.is_empty() {
```
jsuereth, line 118: Nit:  You can probably make a helper function that throws the first item or all errors.  Perhaps even something like "TryFrom".

```rust
if let Ok(error) = errors.try_into() {
  return Err(error);
}
```


And then somewhere:

```rust
impl TryFrom<Vec<Error>> for Error {
  type Error = ErrorConversionFailed;
  fn try_from(value: Vec<Error>) -> Result<Self, Self::Error> {
    if value.is_empty() {
        return Err(ErrorConversionFailed{});
    } else if value.len() == 1 {
       return Err(vec.into_iter().next().unwrap());
    } else {
      return Err(Error::CompoundError(errors));
    }
  }
}
```

Or some such.  COnsider this a non-blocking idea.

jsuereth (approved):
Awesome!  I actually expected we'd do the error merging thing as a follow-up PR, but this is better.

crates/weaver_resolver/src/registry.rs
lquerel: I did a slightly different version of this suggestion to simplify even further the caller site.

Caller site: 
```rust
handle_errors(errors)?;
```

Function definition.
```rust
/// Handles a list of errors and returns a compound error if the list is not
/// empty or () if the list is empty.
pub fn handle_errors(errors: Vec<Error>) -> Result<(), Error> {
    if errors.is_empty() {
        Ok(())
    } else {
        Err(Error::CompoundError(errors))
    }
}
```

PR #45, opened 2024-03-12
More flexible `registry generate` command

lquerel:
This PR integrates `MiniJinja` with `jaq` filters (a jq-compatible Rust implementation) to specify what to extract and process from the resolved registry, as well as how to apply the `Jinja2` templates.

All errors are collected and displayed at the end of the command execution. For each error, the command shows the location in the file, the reason for the error, and the content of the evaluation context. This information helps template authors understand the issues. Similar errors are grouped together.

Note 1: I had to update multiple files not related to this PR in order to make the last version of Clippy happy.

Note 2: Why use `jaq` filters instead of `JSONPath` or similar selectors? `jaq` filters are extremely powerful; they can be used to sort or group registry signals (e.g., by prefix) and much more. Such features are not supported by `JSONPath`.

Example of error returned by the command.
```
✖ undefined value (in group.md:3)
---------------------------------- group.md -----------------------------------
   1 | {%- set file_name = ctx.id | file_name -%}
   2 | {{- template.set_file_name("group/" ~ file_name ~ ".md") -}}
   3 > {{ rrr.rr }}
     i    ^^^^^^ undefined value
   4 | # Group `{{ ctx.id }}` ({{ ctx.type }})
   5 |
   6 | ## Brief
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
Referenced variables: {
    ctx: {
        "attributes": [
            {
                "brief": "A unique id to identify a session.",
                "examples": "00112233-4455-6677-8899-aabbccddeeff",
                "name": "session.id",
                "requirement_level": "opt_in",
                "stability": "experimental",
                "type": "string",
            },
            {
                "brief": "The previous `session.id` for this user, when known.",
                "examples": "00112233-4455-6677-8899-aabbccddeeff",
                "name": "session.previous_id",
                "requirement_level": "opt_in",
                "stability": "experimental",
                "type": "string",
            },
        ],
        "brief": "Session is defined as the period of time encompassing all activities performed by the application and the actions executed by the end user.\nConsequently, a Session is represented as a collection of Logs, Events, and Spans emitted by the Client Application throughout the Session's duration. Each Session is assigned a unique identifier, which is included as an attribute in the Logs, Events, and Spans generated during the Session's lifecycle.\nWhen a session reaches end of life, typically due to user inactivity or session timeout, a new session identifier will be assigned. The previous session identifier may be provided by the instrumentation so that telemetry backends can link the two sessions.\n",
        "events": [],
        "id": "session-id",
        "instrument": none,
        "lineage": {
            "provenance": "https://github.com/open-telemetry/semantic-conventions.git/model/session.yaml",
        },
        "metric_name": none,
        "name": none,
        "prefix": "session",
        "span_kind": none,
        "type": "attribute_group",
        "unit": none,
    },
    file_name: "SessionId",
    template: TemplateObject {
        file_name: Mutex {
            data: "group/SessionId.md",
            poisoned: false,
            ..
        },
    },
}
-------------------------------------------------------------------------------

Found 295 similar errors
```


crates/weaver_forge/templates/test/attribute_group.md
```diff
@@ -1,17 +1,17 @@
...
-{{ group.brief | trim }}
+{{ ctx.brief | trim }}
 
-prefix: {{ group.prefix }}
+prefix: {{ ctx.prefix }}
 
 ### Attributes
 
-{% for attribute in group.attributes %}
+{% for attribute in ctx.attributes %}
```
jsuereth, line 14: +1 for moving from group to ctx

src/registry/generate.rs
```diff
@@ -73,14 +75,20 @@ pub(crate) fn command(
...
-    engine
-        .generate_registry(
-            logger.clone(),
-            &schema.registries[0],
-            &schema.catalog,
-            args.output.as_path(),
-        )
-        .expect("Failed to generate registry assets");
+    let template_registry =
+        TemplateRegistry::try_from_resolved_registry(&schema.registries[0], &schema.catalog)
```
jsuereth, line 79: I'm not a huge fun of always indexing the first registry here.  Did you have a bug open to clean that up at some point or alter the representation?

I think you could even just pass the entire `&schema` directly to `try_from_resolved_registry` to hide that complexity until we clean it up.

jsuereth (approved):

src/registry/generate.rs
lquerel: I will open an issue to improve this as it's indeed far from optimal. It was like an intermediary/unfinished step, the final goal being at some point to support multiple registries from the same app telemetry schema (e.g. OTel official registry + 0 to n vendor or internal registries).

PR #47, opened 2024-03-13
Sub-command `registry stats` and improvements on ResolvedTelemetrySchema

lquerel:
This PR contains some API usage improvements regarding ResolvedTelemetrySchema and Registry. 

```rust
#[test]
    fn test_api_usage() -> Result<(), Box<dyn Error>> {
        let logger = TestLogger::new();
        let registry_id = "local";
        let registry_dir = "data/registry-test-7-spans/registry/*.yaml";

        // Load a semantic convention registry from a local directory.
        // Note: A method is also available to load a registry from a git
        // repository.
        let mut semconv_registry = SemConvRegistry::try_from_path(registry_id, registry_dir)?;

        // Resolve the semantic convention registry.
        let resolved_schema = SchemaResolver::resolve_semantic_convention_registry(
            &mut semconv_registry,
            logger.clone(),
        )?;

        // Get the resolved registry by its ID.
        let resolved_registry = resolved_schema.registry(registry_id).unwrap();

        // Get the catalog of the resolved telemetry schema.
        let catalog = resolved_schema.catalog();

        // Scan over all the metrics
        for metric in resolved_registry.groups(GroupType::Metric) {
            let _resolved_attributes = metric.attributes(catalog)?;
            // Do something with the resolved attributes.
        }

        // Scan over all the spans
        for span in resolved_registry.groups(GroupType::Span) {
            let _resolved_attributes = span.attributes(catalog)?;
            // Do something with the resolved attributes.
        }

        Ok(())
    }
```

Another example of using the ResolvedTelemetrySchema and the registry has been implemented in the form of a new sub-command displaying general statistics about a registry (see sub-command `weaver registry stats`).


crates/weaver_resolved_schema/src/catalog.rs
```diff
@@ -4,21 +4,37 @@
...
-    pub metrics: Vec<Metric>,
+}
+
+/// Statistics on a catalog.
+#[derive(Debug, Serialize)]
+pub struct Stats {
+    /// Total number of attributes.
+    pub attribute_count: usize,
+    /// Breakdown of attribute types.
+    pub attribute_type_breakdown: BTreeMap<String, usize>,
```
jsuereth, line 31: Is this based on benchmarking?  I've also antecdotally seen BTreeMap outpeform std HashMap.

crates/weaver_resolved_schema/src/error.rs
```diff
@@ -0,0 +1,49 @@
...
+    },
+
+    /// A generic container for multiple errors.
+    #[error("Errors:\n{0:#?}")]
+    CompoundError(Vec<Error>),
+}
+
+/// Handles a list of errors and returns a compound error if the list is not
+/// empty or () if the list is empty.
+pub fn handle_errors(errors: Vec<Error>) -> Result<(), Error> {
```
jsuereth, line 27: Maybe we shouldn't carry this everywhere. I wonder if we could abstract it in some way...

crates/weaver_resolver/src/registry.rs
```diff
@@ -680,4 +674,51 @@ mod tests {
 
         Ok(())
     }
+
+    #[test]
+    fn test_api_usage() -> Result<(), Box<dyn Error>> {
```
jsuereth, line 679: If you use `#[from]` annotation on errors (using ThisError), you can lift errors from other crates into a shared Error type and avoid Box/dyn here.

The alternative is just to implement `From<OtherError> for Error`.

I think we should definitely (eventually) do that for weaver.  Not hugely important for this test.

jsuereth (approved):

crates/weaver_resolved_schema/src/catalog.rs
lquerel: No, it's not performance-related. It's because I wanted a sorted breakdown.

crates/weaver_resolved_schema/src/error.rs
lquerel: Yes, I agree; I was thinking the same thing. I need to explore ways to unify that in a future PR.

Issue #52, opened 2024-03-17
Fix lineage computation


lquerel:

jsuereth:
This appears to be fixed with PR #53

lquerel:
Reopening this issue, because we still need to check/complete few things here. The calculation of the lineage is not yet fully tested.

PR #53, opened 2024-03-17
Fix resolution process

lquerel:
This PR fixes an issue in the resolution process when there are more than 1 level of `extends`.

Note: The lineage computation is still incomplete and will be fixed in the next PR (see https://github.com/open-telemetry/weaver/issues/52). This is not blocking.


crates/weaver_resolver/src/registry.rs
```diff
@@ -512,6 +528,192 @@ fn resolve_include_constraints(ureg: &mut UnresolvedRegistry) -> bool {
...
+                    sampling_relevant: parent_sampling_relevant,
+                    note: parent_note,
+                    stability: parent_stability,
+                    deprecated: parent_deprecated,
+                    ..
+                } => {
+                    // attr and attr_parent are both references.
+                    AttributeSpec::Ref {
+                        r#ref: r#ref.clone(),
+                        brief: if brief.is_some() {
```
jsuereth, line 608: For all these options, you can make a helper method to clean this up.

```rust
fn clone_first_some<T : Clone>(first: &Option<T>, second: &Option<T>): Option<T> {
  if first.is_some() {
      first.clone()
  } else {
     second.clone()
  }    
```


Then all this code becomes:

```rust
AttribtueSpec::Ref {
   ...
  brief: clone_first_some(brief, parent_brief),
  examples: clone_first_some(examples, parent_examples),
  ...
}
```

jsuereth (approved):
It's interesting how a small fix ripples through all the test code.  I think maybe I need to do a better job focusing on the markdown or we need to start pulling in more of the existing markdown as our tests....

IN any case, looks good! Thanks for fixing this!

crates/weaver_resolver/src/registry.rs
lquerel: Created the following GH issue https://github.com/open-telemetry/weaver/issues/54

lquerel:
> It's interesting how a small fix ripples through all the test code. I think maybe I need to do a better job focusing on the markdown or we need to start pulling in more of the existing markdown as our tests....
> 
> IN any case, looks good! Thanks for fixing this!

Yes looks like using the existing markdown files is our best option at this point.

PR #65, opened 2024-03-27
Add filtering markdown by tag to markdown snippet generation

jsuereth:
- Update model to deal with stability level now on enumeration values.
- Copy "parameter_tag" test from upstream build_tools to ensure complaince in weaver.
- Update various template fixes in latest build-tools release.
- Update sort-order of Attributes to be Requirement-Level then name (matching latest build-tools release).
- Add new mechanism to copy over build-tools markdown snippet tests


lquerel (approved):
LGTM

PR #67, opened 2024-03-28
Write in update-markdown command for registry

jsuereth:
After this PR, we can start trying out the update-markdown command against semconv registry.

- Create new helper method to deal with possible git repo location.
- Adds file-walking all markdown files
- For now, panics on failure to ensure bad exit from process.
- Adds dry_run argument for validation vs. actual update.


src/registry/update_markdown.rs
```diff
@@ -25,13 +26,53 @@ pub struct RegistryUpdateMarkdownArgs {
...
+    let registry = ResolvedSemconvRegistry::try_from_url(
+        args.registry.clone(),
+        args.registry_git_sub_dir.clone(),
+        cache,
+        log.clone(),
+    )
+    .unwrap_or_else(|e| {
+        panic!("Failed to resolve the semantic convention registry.\n{e}");
+    });
+    log.success("Registry resolved succesfully");
```
lquerel, line 50:
```suggestion
    log.success("Registry resolved successfully");
```

src/registry/update_markdown.rs
```diff
@@ -25,13 +26,53 @@ pub struct RegistryUpdateMarkdownArgs {
...
+    {
+        log.info(&format!("{}: ${}", operation, entry.path().display()));
+        if let Err(error) =
+            update_markdown(&entry.path().display().to_string(), &registry, args.dry_run)
+        {
+            has_error = true;
+            log.error(&format!("{error}"));
+        }
+    }
+    panic!("weaver registry update-markdown failed.")
```
lquerel, line 77: I don't understand this panic! at the end of the command.

lquerel (commented):
I don't understand how this command can end without a panic. I might miss something obvious, but I don't see it :-)

src/registry/update_markdown.rs
jsuereth: Fixed

lquerel (approved):

src/registry/update_markdown.rs
lquerel: Yes either panic! or process::exit(1). (see https://doc.rust-lang.org/std/process/fn.exit.html)

PR #73, opened 2024-03-28
Compute lineage metadata at the attribute field level

lquerel:
This PR refines the concepts of group and attribute lineage metadata. An attribute field may be defined locally, inherited from a parent group, or overridden locally. The lineage section, which is optional, describes the origin of a field. The absence of lineage information indicates that the field is defined locally.

These lineage metadata are useful for debugging, assisting semantic convention authors, and aiding application telemetry schema authors in determining the provenance of each field in the resolved registry or schema. Additionally, they can be integrated into documentation for data governance purposes.


crates/weaver_forge/src/lib.rs
```diff
@@ -497,6 +497,9 @@ mod tests {
                     expected_dir.as_ref().join(file),
                     observed_dir.as_ref().join(file)
                 );
+
+                eprintln!("Expected file content: {}", file1_content);
```
jsuereth, line 501: FYI - you can use the diff library I pulled for this purpose:

https://github.com/open-telemetry/weaver/blob/main/crates/weaver_semconv_gen/src/diff.rs#L12

We can move that somewhere more friendly if needed.

jsuereth:
I might need to see some kind of design document here to fully appreciate what you're going for and how to use this.


Could update the `README.md` file on one of the crates (either semconv or resolver) to describe the data model and how it's used? I'm having trouble seeing the forest for the trees in all the Rust changelogs.

It'd be nice if, as part of this, we document all the possible lineage paths we allow (and check the use cases are valid).

I will go back over the code, but I'm still trying to grok the output a bit.  A few more questions:

- One thing I worry about (in the core model today) is this "intermediate" ref where we `ref` solely by attribute name but have overrides in an attribute group.  Is this just a side-effect of your format or is this inherit in how resolution happens today?
- lineage has "source_file" today.  Should this be a relative path in a registry + the registry itself?  What does this look like when we "flatten" schema with resolved schema?
- For locally-overriden and inherited fields, do you want explicit or implicit references?  How much of that data could be inferred from the raw groups + refs, where lineage JUST tracks origin URLs/files/etc.

crates/weaver_forge/src/lib.rs
lquerel: Good point. Created an issue for a future PR. 
https://github.com/open-telemetry/weaver/issues/77

lquerel:
I've updated the `weaver_resolver` README.md to include a description of the lineage concept and addressed some of your questions in the documentation. However, a few questions remain unanswered as I may need further clarification; see below.

> * One thing I worry about (in the core model today) is this "intermediate" ref where we `ref` solely by attribute name but have overrides in an attribute group.  Is this just a side-effect of your format or is this inherit in how resolution happens today?

Are you concerned about the fact that the resolution process produces a description of the `intermediate.level` group with its local lineage, or is it something else? I don't think I understood the problem you're raising in this question.

> * For locally-overriden and inherited fields, do you want explicit or implicit references?  How much of that data could be inferred from the raw groups + refs, where lineage JUST tracks origin URLs/files/etc.

I described the process for locally-overriden and inherited fields in the README.md, let me know if that doesn't answer your initial question.

jsuereth:
> Are you concerned about the fact that the resolution process produces a description of the intermediate.level group with its local lineage, or is it something else? I don't think I understood the problem you're raising in this question.

My concern is about whether this inheritance is ambiguous in the underlying model.  When reading your YAML what crossed my mind was "how would I figure out the order if not for your lineage blocks", e.g. if lineage was *optional*.

crates/weaver_resolver/README.md
```diff
@@ -33,14 +35,80 @@ involves the following steps:
...
+a semantic convention registry. The lineage as such is not part of the syntax
+and structure of a semantic convention; rather, it's an extension produced by
+the `weaver` tool, intended for use in scenarios such as:
+- A semconv author wishes to verify the exact path followed by the resolution
+process in the case of a complex cascade of inheritance across multiple levels
+between groups.
+- A documentation process aims to add lineage information to the documentation
+of each attribute to allow readers to understand where the definitions of each
+field were defined.
+- The lineage information of a semantic convention could eventually feed into
```
jsuereth, line 61: So this an experimental feature (today).  I was curious if you wanted to include this in application-telemetry-schema.

This may require some discussion as I might want this to be a defacto part of "resolved telemetry schema" so provenance isn't an opt-in.  For now, this is fine, as the whole project is still experimental.

crates/weaver_resolver/README.md
```diff
@@ -33,14 +35,80 @@ involves the following steps:
...
+  "groups": [
+    {
+      // Standard semconv fields 
+      "id": "<group id>",
+      "type": "<group type>",
+      "brief": "<group brief>",
+      "attributes": [ /* ... */ ],
+      "lineage": {
+        // `source_file` is either:
+        // - a relative path in the case of a registry that has been explicitly
```
jsuereth, line 79: In the relative path case, I'm concerned about how this evolves from file->file->file in the application telemetry schema case.


A consumer of this resolution, do they preserve relative files or collapse them to some absolutely URL where they concsumed the original source?

For example:

- Semconv resolves their metrics and has lineage with relative paths.
- Application A creates their own schema using semconv, and resolves.
- When referring to lineage, do I see relative paths or urls or both?

This may be jumping the gun a bit, but if you plan to resolve this scenario using the same string field `source_file`, that's ok, I'm just concerned about how this "scales" from semconv -> application.

crates/weaver_resolver/README.md
```diff
@@ -33,14 +35,80 @@ involves the following steps:
...
+            // The group ID where the attribute fields were defined before a
+            // possible local redefinition. In the case of an inherited
+            // attribute, this field contains the ID of the group defined in
+            // the `extends` clause of the initial definition. In the case of
+            // a reference to a root attribute, this field contains the ID of
+            // the group where this attribute was defined globally.
+            "source_group": "<group id>",
+            // The field names inherited from the source group. This field is
+            // present only if the attribute is inherited.
+            "inherited_fields": [ "<field name>", /* ... */ ],
```
jsuereth, line 97: Should we assume all fields are inherited unless otherwise overridden?

crates/weaver_resolver/README.md
```diff
@@ -33,14 +35,80 @@ involves the following steps:
...
+            "locally_overridden_fields": [ "<field name>", /* ... */ ],
+          }
+        }
+      }
+    }
+  ]
+}
+```
+
+Important note: Fields that appear in the group but are not mentioned in the
```
jsuereth, line 108: This means the group is the *first* time they appear?  Is this why you split inherited from overridden?

If we have a simple "local always overrides" I think perhaps maybe just "locally_defined" is enough, as anything locally defined would override inherited and you can (easily?) recompute inherited fields by following the lineage train.

crates/weaver_cache/Cargo.toml
```diff
@@ -14,7 +14,7 @@ workspace = true
 [dependencies]
 tempdir = "0.3.7"
 dirs = "5.0.1"
-gix = { version = "0.61.1", default-features = false, features = [
+gix = { version = "0.61.0", default-features = false, features = [
```
jsuereth, line 17: We should do these bumps w/ dependabot PRs I think.

jsuereth (approved):
While I still have some overall questions, code looks good.

For an experimental feature of an experimental project, this is good, but I think we should be discussing some long term objectives for the YAML and requirements.  As stated before I think it's hard to have these discussions in code, so maybe we should write up requirements of lineage before tackling more in that area.  This demo is pretty strong right now for showing feasibility.

lquerel:
> > Are you concerned about the fact that the resolution process produces a description of the intermediate.level group with its local lineage, or is it something else? I don't think I understood the problem you're raising in this question.
> 
> My concern is about whether this inheritance is ambiguous in the underlying model. When reading your YAML what crossed my mind was "how would I figure out the order if not for your lineage blocks", e.g. if lineage was _optional_.

I see. As I originally imagined it, the purpose of the resolution process is to produce a self-contained description where all the complex elements to be resolved are replaced by their final interpretation. That's why, unless lineage is enabled, the resolved schema no longer retains traces of inheritance relationships, global references, or operations of local override. I feel that lineage might be useful for documenting a semconv registry, but it's less useful for generating an SDK client or for creating a dashboard. Let me know what you think.

crates/weaver_resolver/README.md
lquerel: Ok I will mark this feature as experimental in the documentation and I will add a note to precise that the lineage should be part of the application telemetry schema.

crates/weaver_resolver/README.md
lquerel: I agree, this part is not fully cooked. However, I believe the only viable option for a published resolved schema is to support only URLs, because a local path will always be problematic, even if it's an absolute path. This is definitely something we need to resolve. I will add a note in the document to continue this discussion later.

crates/weaver_resolver/README.md
lquerel: Yes. I will update the description.

crates/weaver_cache/Cargo.toml
lquerel: I agree

crates/weaver_resolver/README.md
lquerel: I replaced "Fields that appear ..." with "Attributes that appear ..." because it was ambiguous. I was talking about attributes that are totally defined locally in a group (i.e. attribute with an ID).

PR #74, opened 2024-03-28
Add docker image and publishing rules for weaver

jsuereth:
This makes a distribution via docker that could help try out existing weaver on semantic convention registry.  I'm not certain we'd consider this a "primary" way of consuming weaver, but for intermediate tasks and experimenting with attribute registry generation, this could help unblock us.

Posting as a draft for thoughts.


lquerel (approved):
LGTM

lquerel:
@jsuereth Thanks !

lmolkova (approved):

PR #75, opened 2024-04-02
Code coverage automation + CHANGELOG.md init

lquerel:
This PR integrates `cargo tarpaulin` to report code coverage on each PR. The current code coverage is 61% (library only), which is acceptable. The goal is to gradually bring the project to 75%. An integration with Codecov is implemented for adding a badge and making a dashboard available.

An additional set of `clippy` checks has also been added to enhance the quality of the project.

A start of a changelog has also been generated with `git cliff` (in unreleased mode). Full automation of the changelog and releases will come in a subsequent PR.


crates/weaver_semconv_gen/Cargo.toml
```diff
@@ -19,11 +19,11 @@ similar = "2.4.0"
 nom = "7.1.3"
 
 thiserror.workspace = true
-serde.workspace = true
-serde_yaml.workspace = true
-serde_json.workspace = true
-rayon.workspace = true
-walkdir.workspace = true
+#serde.workspace = true
```
jsuereth, line 22: Why are all of these commented out?

jsuereth (approved):
I think this PR is attempting to do too much in a single PR, making it very hard to review.

I wound up going in a reviewing each commit individually to get a better sense of what was going on, but in geneally I'd recommend:

1. Changelog automation could have been split into its own PR.
2. Codecov could have been its own PR
3. Clippy rules as their own PR or individual PRs per-enabled-rule would help.   There's still some things I don't understand why/what's happening.

My main concern with this PR is for us to be on the same page about expected environment and requirements. So far clippy has been failing inconsistently in many PRs for the same issue.  IF you'd like to start enforcing it, then it'd be nice if we could guarantee it gave consistent issues on all PRs (or, as you've done here, go update all fixes for enforced thigns).

I've been slowly adjusting required GHA checks on this project, so the other thing to be careful of is renaming GHAs as that will break the merge requirements.  You'll have to let me know if you can update those.

In any case, overall LGTM - but wanted to give some feedback to please split this up in smaller chunks in the future.

jsuereth:
You'll notice "check-external-types" and "rustfmt" aren't being completed on this PR.  You'll have to let me know which status checks to update as mandatory

.github/workflows/ci.yml
```diff
@@ -98,9 +102,11 @@ jobs:
       - name: Check documentation
         env:
           RUSTDOCFLAGS: -D warnings
-        run: cargo doc --workspace --all-features --no-deps --document-private-items
+        # [ToDo LQ] Re-enable --all-features once the issue is resolved in Tantivy (zstd-safe). This is an experimental feature anyway.
+        # run: cargo doc --workspace --all-features --no-deps --document-private-items
+        run: cargo doc --workspace --no-deps --document-private-items
   rustfmt:
-    name: rustfmt
+    name: Rustfmt
```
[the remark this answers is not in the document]
lquerel: Okay, I will revert these changes in the naming. I was not aware of the additional mandatory checks in CI. Could you make it visible to me in the project settings? Thanks

crates/weaver_semconv_gen/Cargo.toml
lquerel: Unless I am mistaken, these dependencies were not used in `weaver_semconv_gen`. I believe I saw this in the output of `cargo machete`.

jsuereth: It's odd, when I try to build/use this crate locally it fails unless they're included.  Likely my setup is messed up, but this is why they're in there.

jsuereth: You can remove the comments though, if we need them we'll add them back in fully.

.github/workflows/ci.yml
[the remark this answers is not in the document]
jsuereth: It's ok if you change these, we just need to agree on what the name should be, and ideally not have names that change often (like the MSRV rule now).

You should be able to see what's required under branch protection rules, but we don't (by default) give admin access to these to maintainers.

crates/weaver_semconv_gen/Cargo.toml
lquerel: Checked and Done

PR #76, opened 2024-04-02
Attribute registry location is now an argument to markdown update

jsuereth:
This is the next step in a "demo" for moving markdown generation from build-tools to weaver for semantic-conventions repository.

- Add new attribute registry url parameter
- Update legacy tests to look more similar to build-tools using this parameter.


jsuereth:
@lquerel My PRs keep seeing a clippy failure of:

```
error: field `type_mapping` is never read
  --> crates/weaver_forge/src/config.rs:59:9
   |
41 | pub struct TargetConfig {
   |            ------------ field in this struct
...
59 |     pub type_mapping: HashMap<String, String>,
   |         ^^^^^^^^^^^^
   |
   = note: `TargetConfig` has a derived impl for the trait `Debug`, but this is intentionally ignored during dead code analysis
   = note: `-D dead-code` implied by `-D warnings`
   = help: to override `-D warnings` add `#[allow(dead_code)]`
```


Even though I haven't touched this code at all.  Can I just add allow dead_code or do you have some other fix in place for this?

lquerel (approved):
LGTM

PR #81, opened 2024-04-04
Weaver + Policy Engine Integration - Proposal

lquerel:
I am considering adding a general-purpose policy engine to `Weaver` to support the declarative definition of policies that must be satisfied before publishing a new semantic convention registry. This same mechanism could also be used for future validation of application telemetry schemas.

This PR contains a description of this proposal, and I would like to get the community's feedback on this matter. I believe its implementation is relatively simple while providing an open and easily extensible mechanism.


jsuereth:
Highly supportive of the goals!  I need to look into the implementation proposal

docs/proposals/policy-engine.md
```diff
@@ -0,0 +1,49 @@
+# Policy Engine - Proposal
+
+## Goals
+
+The management of the evolution and maintainability of semantic conventions and
+application telemetry schemas relies on rules or policies that must be verified
+before publishing a new version. The goal of these policies is to ensure the
+coherence and quality of the conventions and schemas over the long term. Below
+are some examples of policies which are already verified or that could be
+verified:
```
pyohannes, line 10: I agree it would be great to have this automated and described independently of tooling (weaver or build-tools).

We already have a quite exhaustive textual description of policies [here](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/versioning-and-stability.md#semantic-conventions-stability).

docs/proposals/policy-engine.md
```diff
@@ -0,0 +1,49 @@
...
+The '[regorus](https://github.com/microsoft/regorus)' project by Microsoft could be used to implement this feature
+without having a dependency on the OPA toolchain.
+
+The policy verification could operate as follows:
+- Resolution of the new version of the semconv registry (or the app telemetry
+schema)
+- Resolution of the previous version of the semconv registry (or the app
+telemetry schema)
+- Application of policies on these two resolved schemas
+- Display of errors if the policies are not adhered to
```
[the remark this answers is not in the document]
lquerel: We already have some policies/rules encoded in the code of the Python build tool and Weaver (e.g., deprecated and stability logic). I would like these types of rules to be externalized and generalized in a declarative form so they can be applied during my check phase, which is used by all the subcommands of the Weaver registry command. Moving the execution of this phase outside of Weaver would only complicate integration from my perspective. Regorus is a library version of OPA, so it's very easy to implement during the file scanning phase of the registry. This would be applicable for both the OpenTelemetry registry and a private enterprise registry. For instance, we're going to use it at F5, and I think that in such scenarios, integration within Weaver is more appropriate.

docs/proposals/policy-engine.md
lquerel: I will add an example of rego rules to represent some of these policies.

lquerel:
This proposal has been implemented and merged in #103

PR #86, opened 2024-04-04
Extract diff utility and use it in weaver_forge tests

jsuereth:
- Extract minimal "diff" function for weaver into crate (hiding dependency on similar)
- Update semconv_markdown_gen to use the diff crate
- Update weaver_forge to use diff crate for test failures.


lquerel (approved):
Very cool. Thanks

PR #87, opened 2024-04-05
Weaver Policy Engine

lquerel:
This PR is based the proposal #81.

This initial implementation (based on the [regorus](https://github.com/microsoft/regorus) policy engine) allows for the detection of violations against best practices within a semantic convention registry (before the resolution phase) through one or several policy files (in rego format). A future PR will enable the application of policies after the registry's resolution to detect other types of violations.

A detailed description of the operating principle can be found in the [README.md](https://github.com/open-telemetry/weaver/blob/21bf966be509cba4527a37c72fdf2663f2a18283/crates/weaver_policy_engine/README.md) of the `weaver_policy_engine` crate.

The impact on performance from evaluating policies on all files in the registry is imperceptible.

Future developments:
- Allow the application of policies after the registry resolution phase.
- Enable specifying two versions of the same registry to apply policies comparing the schema evolution of groups, attributes, or any other signals.
- Currently, only the `weaver registry check` command has the capability to apply these policies. Support for other commands will be added in future PRs.

Note: Due to a conflict issue with serde_json, this PR references a fork of the `regorus` project. A description of the problem can be found [here](https://github.com/microsoft/regorus/issues/199). The described issue is not a major problem, and I believe it is temporary.


crates/weaver_resolver/src/lib.rs
```diff
@@ -419,10 +466,16 @@ impl SchemaResolver {
         resolver_config: ResolverConfig,
         cache: &Cache,
         log: impl Logger + Clone + Sync,
+        policy_engine: Option<&weaver_checker::Engine>,
```
jsuereth, line 469: This is honeslty my only contention with this whole PR here.

SHOULD the policy checking be part of the resolver or can this be a separate thing you need to piece together later as a consumer of these crates.

lquerel: I extracted the policy engine from the SchemaResolver and refactored a bit the command line accordingly.

src/error.rs
```diff
@@ -0,0 +1,77 @@
...
+    /// Call the error code and exit the process if the result is an error.
+    /// Otherwise, return the value.
+    ///
+    /// # Arguments
+    /// * `self` - The result to check
+    /// * `err_handler` - The error handler to call if the result is an error
+    ///
+    /// # Returns
+    /// The value if the result is Ok
+    fn exit_if_error<F: FnOnce(E)>(self, err_handler: F) -> T {
```
jsuereth, line 46: These are quite nice.  Thanks for adding them!

jsuereth (approved):
Wow, that was a lot more tearing things apart than I expected, but I think this separation will pay off later!

Thanks for adding this, excited to try it out/expand on it!

src/error.rs
lquerel: It's better than what we had before, but there are still a few things to do in a future PR.

Issue #49, opened 2024-03-14
Make `handle_errors` generic across crates if possible.


lquerel:
Will probable replaced by the more generic approach described in #124

jsuereth:
Closing this as resolved.  Part 1of fixing WeaverError created a generic handle_errors, but we agree we're moving a different direction.

PR #93, opened 2024-04-10
Streamline error handling in the resolver

pyohannes:
Relates to #49

* This PR pushes the creation of errors down into functions that resolve extend, attribute, and include references. 

  This removes some boilerplate code and also removes the need to iterate over groups another time just to collector error information.

* Some tests for invalid extend, attribute, and include references are added.


crates/weaver_resolver/src/registry.rs
```diff
@@ -329,22 +278,15 @@ fn group_from_spec(group: GroupSpecWithProvenance) -> UnresolvedGroup {
...
         let mut resolved_attr_count = 0;
 
         // Iterate over all groups and resolve the attributes.
         for unresolved_group in ureg.groups.iter_mut() {
             let mut resolved_attr = vec![];
 
-            if unresolved_group.group.extends.is_some() {
-                // If the group has an `extends` clause, we need to resolve the
-                // `extends` references first.
-                unresolved_attr_count += unresolved_group.attributes.len();
```
pyohannes, line 344: It seems to me that's not needed, as extends are supposed to be resolved before this function is called?

lquerel: I agree. The recent changes to the resolution steps have made this step unnecessary. Thanks for pointing it out.

crates/weaver_resolver/src/registry.rs
```diff
@@ -800,6 +764,94 @@ mod tests {
         }
     }
 
+    fn create_registry_from_string(registry_spec: &str) -> Result<Registry, crate::Error> {
```
lquerel, line 767: IMO, the right way to do this will be:
- Create a `SemConvRegistry` with the `new` constructor.
- Add a new method to `SemConvRegistry` to load a semantic convention group similar to `load_from_file` but for `&str`. This method could be named `load_from_string` and take a `&str` as a parameter.
- Remove the dependency to tempfile

crates/weaver_resolver/src/registry.rs
```diff
@@ -800,6 +764,94 @@ mod tests {
...
+            SemConvRegistry::try_from_path(registry_id, registry_file.path().to_str().unwrap())
+                .unwrap();
+
+        let mut attr_catalog = AttributeCatalog::default();
+
+        resolve_semconv_registry(&mut attr_catalog, "https://127.0.0.1", &sc_specs)
+    }
+
+    #[test]
+    #[allow(clippy::print_stdout)]
```
lquerel, line 782: Don't think this annotation is required

lquerel (commented):
Thank you for the simplifications made and the additional tests added. It definitely makes sense. I have made a few comments. I don't think we can close #49 yet though.

crates/weaver_resolver/src/registry.rs
pyohannes: Makes sense, I added a `load_from_str` method to the `SemConvRegistry`.

jsuereth (approved):
Nice change!!!!!

jsuereth:
One thing I'm a bit concerned about here is that errors NO LONGER aggregate across *all* resolution.  It's possible that failure in one step blocks accurate detection of further errors, but the original code was designed to fully aggregate all errors.

Using `?` with helper functions linear-izes the code and error detection.  I'm ok making that change BUT we may need to revisit if we think we can do better job in parallelizing error detection.

Specifically, it's better to report ALL errors at once, then block on first error.  UX from a tool that blocks on first error is pretty poor.

pyohannes:
> One thing I'm a bit concerned about here is that errors NO LONGER aggregate across all resolution. It's possible that failure in one step blocks accurate detection of further errors, but the original code was designed to fully aggregate all errors.

The behavior in this regard should be exactly the same as before the change. There were three invocations of `handle_errors(errors)?`, after each evaluation of extends, attribute references, and includes. If resolving an extends failed, attribute references and includes weren't resolved, but you got a complete list of unresolved extends. The behavior is the same with these proposed changes.

The tests actually detect whether the number of reported errors matches the number of unresolved extends, attribute references, or includes:

```rust
if let crate::Error::CompoundError(errors) = result.unwrap_err() {
  assert!(errors.len() == 2);
```

The tests pass both with and without the proposed changes.

jsuereth:
Thanks for confirming @pyohannes - I somehow missed the splits everywhere.  Just wanted to make sure you understood the original goal for attempting to be as parallel as possible.  This is a huge cleanup.  Will merge once updated.

PR #103, opened 2024-04-12
Improve the project documentation

lquerel:
This PR enhances Weaver's general documentation by explaining its platform design and long-term vision.


crates/weaver_forge/README.md
```diff
@@ -1,6 +1,217 @@
...
+#    application_mode: single
+#  - pattern: "**/span.md"
+#    filter: ".groups[] | select(.type == \"span\")"
+#    application_mode: each
+#  - pattern: "**/spans.md"
+#    filter: ".groups[] | select(.type == \"span\")"
+#    application_mode: single
+```
+
+Supported case converters:
```
jsuereth, line 177: This has me thinking about a few things.

- For semconv, `kebab-case` unfortunately splits on non-characters (e.g. numerics) so `k8s` becomes `k-8-s`.  I'm not sure why, but for that case specifically, we need something a bit more flexible.  At a minimum we should try to better document (or link to the case library in rust) here how these work.
- It *may* be better for us to expose *all* the case converters from the case libary as "raw" jinja filters rather than using the config file as is today.  I"m not sure what the value of `{{ foo | file_name }}` vs just `{{ foo | kebab-case ~ ".md" }}` (assuming kebab case worked).

crates/weaver_forge/README.md
```diff
@@ -1,6 +1,217 @@
...
+- lowercase
+- UPPERCASE
+- PascalCase
+- camelCase
+- snake_case
+- SCREAMING_SNAKE_CASE
+- kebab-case
+- SCREAMING-KEBAB-CASE
+
+## Custom Filters
```
jsuereth, line 187: Can you add:

- `split_ids`: Splits a string by '.' creating a list of nested ids.
- `flatten`:  Converts a List of Lists into a single list with all elements. e.g. `[[a,b],[c]] => [a,b,c]`


Also - I'm thinking we should drop "split_ids" for a more generic "split" filter that takes the pattern to split as argument....

docs/architecture.md
```diff
@@ -0,0 +1,56 @@
...
+purpose of the crate and how to use it. Furthermore, the name of each crate
+must be prefixed with `weaver_` to avoid any conflicts with existing crates on
+crates.io.
+
+The following is a list of crates in the workspace, along with a brief
+description and the current status of each crate:
+
+| Crate                                                             | Description                                                          | Status           |
+|-------------------------------------------------------------------|----------------------------------------------------------------------|------------------|
+| [weaver_semconv](crates/weaver_semconv/README.md)                 | Semantic Convention Registry Data Model                              | Alpha            |
```
jsuereth, line 40: This is missing `weaver_diff`

docs/usage.md
```diff
@@ -0,0 +1,131 @@
...
+Usage: weaver registry update-markdown [OPTIONS] <MARKDOWN_DIR>
+
+Arguments:
+  <MARKDOWN_DIR>  Path to the directory where the markdown files are located
+
+Options:
+  -r, --registry <REGISTRY>
+          Local path or Git URL of the semantic convention registry to check [default: https://github.com/open-telemetry/semantic-conventions.git]
+  -d, --registry-git-sub-dir <REGISTRY_GIT_SUB_DIR>
+          Optional path in the Git repository where the semantic convention registry is located [default: model]
```
jsuereth, line 108: We should update the docs to call out this is an unused parameter if the registry specified is NOT a git URL.

here - and everywhere else I think.

crates/weaver_forge/README.md
lquerel: This is fixed in #105 and #106

docs/architecture.md
lquerel: Fixed

docs/usage.md
lquerel: Fixed

PR #121, opened 2024-04-18
(fix) Fix resolution of attribute ids

jsuereth:
Attributes would ONLY have prefix applied when being extended in certain scenarios.   This lead to an issue where unresolved attribute ids were cloned *before* prefix would be applied.  This PR fixes that by introducing a new step in resolution.

- Adds new resolution step to resolve all attribute ids to "full" id before continuing any other resolution.
- Update all the (actually broken) generated templates that were using the wrong attributes for JVM metrics (stripping prefix).
- Add a  specific test case for the attribute id + extends case that was broken.
- Update weaver_resolver tests to output colored diff on failure

See https://github.com/open-telemetry/semantic-conventions/pull/917#discussion_r1566611724 for context

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4472,6 +4472,7 @@ dependencies = [
  "weaver_cache",
  "weaver_checker",
  "weaver_common",
+ "weaver_diff",
  "weaver_resolved_schema",
  "weaver_schema",
  "weaver_semconv",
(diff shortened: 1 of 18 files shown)


lquerel (approved):
Thanks for this fix. LGTM

PR #129, opened 2024-04-24
Part 1 of working on #[derive(WeaverError)]

jsuereth:
- Move all error-related raw methods into `weaver_common`
- Add `compound` method to `WeaverError` trait.
- Update *all* implementations of `WeaverError` to have consistently the same implementation.

This was just a quick pass at #49.  Should make it easier to refactor going forward.


lquerel:
Part 1 simplifies the current situation, so LGTM. 

However, before starting Part 2, I think we need to make a decision regarding issue #130  as there could potentially be an overlap. I'll let you review my proposal, and we can discuss it via Slack when you have a moment.

lquerel (approved):
See general comment.

jsuereth:
> Part 1 simplifies the current situation, so LGTM.
> 
> However, before starting Part 2, I think we need to make a decision regarding issue #130 as there could potentially be an overlap. I'll let you review my proposal, and we can discuss it via Slack when you have a moment.

Agreed - wasn't going to start part two just yet.

PR #136, opened 2024-04-26
[WIP] Code Generation with Weaver - Example of Code Generation for SemConv Attribute Registry and SemConv Metrics

lquerel:
IMPORTANT NOTES:
- THIS PR SHOWS HOW TO USE WEAVER TO GENERATE CODE FOR SEMANTIC CONVENTIONS.
- THE GENERATED RUST CODE IS JUST AN EXAMPLE, NOT AN OFFICIAL RUST SEMCONV VERSION.
----
# Guidance to review this PR
The purpose of this PR is to allow various maintainers of API/SDK clients to provide feedback on Weaver in the context of code generation. The generation of Rust code for SemConv is taken as an example.

File structure, under `crates/weaver_codegen_test/`:
- `semconv_registry/`: Contains the SemConv registry used in this example.
- `templates/registry/rust/`: Holds the templates for this PR.
- `templates/registry/rust/weaver.xml`: Defines the JQ expression for applying each template to the resolved registry.
- `generated_code/`: Location where the Rust SemConv code is generated.

The following command can be used to reproduce this code generation:

```bash
cargo run -- registry generate \
-t crates/weaver_codegen_test/templates/ \
-r crates/weaver_codegen_test/semconv_registry \
rust crates/weaver_codegen_test/generated_code/
``` 

# To the attention of API/SDK client maintainers,

Weaver uses a Jinja2-compatible engine to render the templates. The following custom filters are available:
- `lower_case`: Converts a string to lowercase.
- `upper_case`: Converts a string to UPPERCASE.
- `title_case`: Converts a string to TitleCase.
- `pascal_case`: Converts a string to PascalCase.
- `camel_case`: Converts a string to camelCase.
- `snake_case`: Converts a string to snake_case.
- `screaming_snake_case`: Converts a string to SCREAMING_SNAKE_CASE.
- `kebab_case`: Converts a string to kebab-case.
- `screaming_kebab_case`: Converts a string to SCREAMING-KEBAB-CASE.
- `acronym`: Replaces acronyms in the input string with the full name defined in the `acronyms` section of the `weaver.yaml` configuration file.
- `split_ids`: Splits a string by '.' creating a list of nested ids.
- `type_mapping`: Converts a semantic convention type to a target type (see weaver.yaml section `type_mapping`).
- `comment_with_prefix(prefix)`: Outputs a multiline comment with the given prefix.
- `flatten`: Converts a List of Lists into a single list with all elements.
e.g. \[\[a,b\],\[c\]\] => \[a,b,c\]
- `attribute_sort`: Sorts a list of `Attribute`s by requirement level, then name.
- `metric_namespace`: Converts registry.{namespace}.{other}.{components} to {namespace}.
- `attribute_registry_file`: Converts registry.{namespace}.{other}.{components} to attributes-registry/{namespace}.md (kebab-case namespace).
- `attribute_registry_title`: Converts registry.{namespace}.{other}.{components} to {Namespace} (title case the namespace).
- `attribute_registry_namespace`: Converts metric.{namespace}.{other}.{components} to {namespace}.
- `attribute_namespace`: Converts {namespace}.{attribute_id} to {namespace}.

> Note 1: Most of the standard Jinja2 filters are also supported.

> Note 2: Some filters present in the build-tools have not yet been implemented in Weaver. It would be interesting to get feedback from the maintainers to identify which are missing.

The following custom testers are available:
- `stable`: Tests if an `Attribute` is stable.
- `experimental`: Tests if an `Attribute` is experimental.
- `deprecated`: Tests if an `Attribute` is deprecated.

# To the attention of the maintainers of opentelemetry-rust
@jtescher, @TommyCpp , @cijothomas , @lalitb , @djc (and any other opentelemetry-rust maintainers/contributors), please review this PR and provide feedback on both the Weaver tool and the Rust code generated from a few Jinja2 templates that I assembled specifically for this exercise.

The generated code adheres to the recommendations defined in this [PR](https://github.com/open-telemetry/semantic-conventions/pull/953) by @lmolkova, along with a few other suggestions of my own:
- One Rust file per group namespace.
- Every SemConv experimental attribute is annotated with `#[cfg(feature = "semconv_experimental")]`
- Every SemConv deprecated attribute is annotated with `#[deprecated(note="{{ attribute.deprecated }}")]`.
- Every SemConv enumeration is translated into a non-exhaustive Rust enum. Each of these enums implement the `Display` trait and expose a method `as_str` returning the string value for each variant. A `_Custom` variant is also added to let the users define their own variant.
- Every attribute of the registry is declared as a typed AttributeKey constant. A method is exposed to generate a standard KeyValue from a properly typed given value. 
- For the metrics, this example generates an API similar to the one used by the Python SDK, as well as a type-safe API (more details below).

Below is an example showcasing the creation of metrics using the type-safe API generated by the templates. The goal is to guide our users through the smart completion offered by most IDEs (including documentation), and also to leverage Rust's type system and compiler to guide the user and prevent any omissions, errors, and misuse.

```rust
    // ==== A TYPE-SAFE HISTOGRAM API ====
    // Create a u64 http.server.request.duration metric (as defined in the OpenTelemetry HTTP
    // semantic conventions)
    let http_request_duration = HttpServerRequestDuration::<u64>::new(&meter);

    // Records a new data point and provide the required and some optional attributes
    http_request_duration.record(100, &HttpServerRequestDurationReqAttributes {
        http_request_method: HttpRequestMethod::Connect,
        url_scheme: "http".to_owned(),
    }, Some(&HttpServerRequestDurationOptAttributes {
        http_response_status_code: Some(200),
        ..Default::default()
    }));

    // ==== A TYPE-SAFE UP-DOWN-COUNTER API ====
    // Create a f64 http.server.request.duration metric (as defined in the OpenTelemetry HTTP
    // semantic conventions)
    let http_client_active_requests = HttpClientActiveRequests::<f64>::new(&meter);

    // Adds a new data point and provide the required attributes. Optional attributes are not
    // provided in this example.
    http_client_active_requests.add(10.0, &HttpClientActiveRequestsReqAttributes {
        server_address: "10.0.0.1".to_owned(),
        server_port: 8080,
    }, None);

    // ==== A TYPE-SAFE COUNTER API ====
    // Create a f64 system.cpu.time metric (as defined in the OpenTelemetry System semantic
    // conventions)
    let system_cpu_time = SystemCpuTime::<f64>::new(&meter);

    // Adds a new data point and provide some optional attributes.
    // Note: In the method signature, there is no required attribute.
    system_cpu_time.add(10.0, Some(&SystemCpuTimeOptAttributes {
        system_cpu_logical_number: Some(0),
        system_cpu_state: Some(SystemCpuState::Idle)
    }));
    // Adds a new data point with a custom CPU state.
    system_cpu_time.add(20.0, Some(&SystemCpuTimeOptAttributes {
        system_cpu_logical_number: Some(0),
        system_cpu_state: Some(SystemCpuState::_Custom("custom".to_owned()))
    }));

    // ==== A TYPE-SAFE GAUGE API ====
    // Create a i64 system.cpu.utilization metric (as defined in the OpenTelemetry System semantic
    // conventions)
    let system_cpu_utilization = SystemCpuUtilization::<i64>::new(&meter);

    // Adds a new data point with no optional attributes.
    system_cpu_utilization.record(-5, None);
    // Adds a new data point with some optional attributes.
    system_cpu_utilization.record(10, Some(&SystemCpuUtilizationOptAttributes {
        system_cpu_logical_number: Some(0),
        system_cpu_state: Some(SystemCpuState::Idle)
    }));
```

> Note 1: **Similarly to other Client SDKs, the individual SemConv attributes and metrics are accessible via the generated code without using the type-safe API (as seen in the Python or Java SDKs). We could offer a configuration mechanism allowing users to choose between permissive (but error-prone) APIs or more rigid, type-safe APIs.**

> Note 2: A similar approach could be used for Span, ...

**Discussion on performance**:
- The current experimentation generates code relying on the opentelemetry and opentelemetry-sdk crates. The entirety of the existing pipeline is reused. The type-safe API generated on top of the SDK is relatively light but adds a small overhead.
- It is possible in the long run to completely eliminate the numerous layers of abstraction by directly generating the code that will interface this type-safe API with the opentelemetry-otlp crate (or any other target exporter). With this approach, many optimizations are possible: Direct conversion of Rust structures representing required and optional attributes to corresponding protobuf messages. Aggregation relies on computing a hash combining all key+value attributes sorted by key; it is possible to completely eliminate this sorting as well as half of the overhead of the hash (hashing of values only). Many other optimizations are possible with this type of approach.

# Status
- [x] Implement the `Display` trait for every enum.
- [x] Expose the method `as_str` for every enum. 
- [x] Generate deprecated and experimental attributes.
- [x] Use the first segment in the attribute id instead of the prefix of groups.
- [x] Generate Metrics
- [x] Test code generate (see crate weaver_codegen_test).
- [x] More documentation on JQ syntax.

Work for future PR: Generate type-safe API for spans.


jsuereth (commented):
All of the code in weaver lgtm.

Regarding rust templates -  I think those either belong in an examples directory or a separate or for otel-rust.  At a minimum there should be a test on them if they live in weaver.

lquerel:
@jsuereth yes this PR is still a draft I will add some tests today and I will move the Rust templates into an examples directory.

crates/weaver_forge/codegen_examples/expected_codegen/http.rs
```diff
@@ -0,0 +1,139 @@
...
+    /// PATCH method.
+    Patch,
+    /// POST method.
+    Post,
+    /// PUT method.
+    Put,
+    /// TRACE method.
+    Trace,
+    /// Any HTTP method that the instrumentation has no prior knowledge of.
+    Other,
```
lmolkova, line 69: Those should have string values matching ones defined in the spec, i.e. `CONNECT`, `_OTHER` - is it possible in Rust?

Also, enums are notoriously difficult in other languages for their back-compat issues  (calling `Enum.NewValueAddedInVNext` but in runtime you deal with vOld of the enum vlass). I assume it's not an issue in Rust?

crates/weaver_forge/codegen_examples/mini_registry/deprecated/network.yaml
```diff
@@ -0,0 +1,121 @@
+groups:
+  - id: registry.network.deprecated
```
lmolkova, line 2: it seems deprecated attributes are not generated - is it intended?

crates/weaver_forge/codegen_examples/templates/registry/rust/semantic_attributes.rs.j2
```diff
@@ -0,0 +1,54 @@
...
+{%- if attribute.brief %}
+{{ attribute.brief | comment_with_prefix("/// ") }}
+{%- endif %}
+{%- if attribute.note %}
+///
+/// Notes:
+{{ attribute.note | comment_with_prefix("///   ") }}
+{%- endif %}
+{%- if attribute is experimental %}
+#[cfg(feature = "semconv_experimental")]
```
lmolkova, line 31: this is so cool!

crates/weaver_forge/codegen_examples/templates/registry/rust/weaver.yaml
```diff
@@ -0,0 +1,62 @@
...
+    filter: >
+      .groups
+      | map(select(.id | startswith("registry.")))
+      | map(select(.type == "attribute_group" and .prefix != null and .prefix != "")    
+        | { 
+          id, 
+          type, 
+          brief, 
+          prefix, 
+          attributes: (.attributes
```
lmolkova, line 28: it depends on the SIG decision, but I'd generate experimental deprecated attributes

crates/weaver_forge/codegen_examples/templates/registry/rust/weaver.yaml
```diff
@@ -0,0 +1,62 @@
...
+      | map(
+          {
+            id,
+            type,
+            brief,
+            prefix
+          }
+        )
+      | unique_by(.prefix)
+      | sort_by(.prefix)
```
lmolkova, line 40: existing tooling does not rely on the group prefix, it uses the first segment in the attribute id calculated during the resolution process and provided in the jinja env.

I don't think it's important right now but presence of prefix is not enforced and from the current schema perspective the following are equivalent:

```yaml
- id: group1
  prefix: foo
  attributes: 
  - id: bar
```

and 

```yaml
- id: group1
  attributes: 
  - id: foo.bar
```

crates/weaver_forge/codegen_examples/expected_codegen/http.rs
lquerel: Yes, there are ways to create this mapping, either with a macro or directly via the implementation of the Display trait. I will add that support.

To handle the evolution of enum definitions, I used the `#[non_exhaustive]` macro attribute, which indicates that a type or variant may have more fields or variants added in the future (more details [here](https://doc.rust-lang.org/reference/attributes/type_system.html)).

crates/weaver_forge/codegen_examples/mini_registry/deprecated/network.yaml
lquerel: Yes, I intentionally didn't generate attributes that are both `experimental` and `deprecated`. However, I can easily remove this filter from the JQ expression defined in the `weaver.yaml` config file. Please let me know if you prefer to generate these attributes.

crates/weaver_forge/codegen_examples/templates/registry/rust/weaver.yaml
```diff
@@ -0,0 +1,62 @@
...
+    # - groups with an id starting with the prefix `registry.`
+    # - groups of the type `attribute_group`.
+    # - groups with a well-defined prefix.
+    # - groups with a non-empty list of attributes that are neither deprecated nor experimental.
+    # - groups are deduplicated by prefix.
+    # - groups are sorted by prefix.
+    filter: >
+      .groups
+      | map(select(.id | startswith("registry.")))
+      | map(select(.type == "attribute_group" and .prefix != null and .prefix != "")
```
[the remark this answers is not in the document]
lquerel: No, I'd like to generate metrics, spans, etc. Is the Java SDK the best example for such signals derived from semconv?

crates/weaver_forge/codegen_examples/templates/registry/rust/weaver.yaml
lquerel: Ok I will change the configuration, no change in the template is required.

crates/weaver_forge/codegen_examples/templates/registry/rust/weaver.yaml
lquerel: Thanks for the advice, I will follow this approach.

crates/weaver_forge/codegen_examples/templates/registry/rust/weaver.yaml
lquerel: Changed the config.

crates/weaver_forge/codegen_examples/mini_registry/deprecated/network.yaml
lquerel: Changed the default configuration to include `experimental` and `deprecated` attributes.

crates/weaver_forge/codegen_examples/expected_codegen/http.rs
lquerel: The string values for the enums are now supported. See https://github.com/open-telemetry/weaver/pull/136/commits/9483d63349c12aaa53da99c3a6002f6eaab7aef9

crates/weaver_forge/codegen_examples/expected_codegen/lib.rs
```diff
@@ -0,0 +1,68 @@
...
+/// These attributes may be used to describe the client in a connection-based network interaction where there is one side that initiates the connection (the client is the side that initiates the connection). This covers all TCP network interactions since TCP is connection-based and one side initiates the connection (an exception is made for peer-to-peer communication over TCP where the "user-facing" surface of the protocol / API doesn't expose a clear notion of client and server). This also covers UDP network interactions where one side initiates the interaction, e.g. QUIC (HTTP/3) and DNS.
+pub mod client;
+/// This document defines the shared attributes used to report a single exception associated with a span or log.
+pub mod exception;
+/// This document defines semantic convention attributes in the HTTP namespace.
+pub mod http;
+/// These attributes may be used for any network related operation.
+pub mod net;
+
+/// A typed attribute key.
```
cijothomas, line 21: Not a problem introduced in this PR, but just wanted to share that OTel Rust already is suffering from performance issues in its abstractions (https://github.com/open-telemetry/opentelemetry-rust/issues/1642), so another abstraction here is going to make it worse :(

We are still working through the issue in OTel Rust group, not a blocker for this PR !

crates/weaver_forge/codegen_examples/templates/registry/rust/weaver.yaml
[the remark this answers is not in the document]
lquerel: Metrics are now generated.

crates/weaver_forge/codegen_examples/expected_codegen/lib.rs
lquerel: In theory, this generated code, particularly for the type-safe API, could lead to more optimized code than what might be achievable with a generic client SDK.

lquerel: @cijothomas I've added a discussion on performance and on how, in the long run, a code generation approach can lead to optimized versions of an SDK client for a given application.

crates/weaver_forge/src/debug.rs
```diff
@@ -2,14 +2,34 @@
...
-use crate::error::Error;
 use crate::error::Error::{CompoundError, TemplateEvaluationFailed};
 use indexmap::IndexMap;
+use std::error::Error;
 use weaver_common::Logger;
 
-/// Return a nice summary of the error.
+/// Return a nice summary of the error including the chain of causes.
+/// Only the last error in the chain is displayed with a full stack trace.
 pub(crate) fn error_summary(error: minijinja::Error) -> String {
```
jsuereth, line 12: If this does what I think it does, maybe it's time to move the Docker image from debug=>release?

I can start a tracker but it'd be nice to rely on release builds (~100MB vs. 400MB somehow).

crates/weaver_forge/codegen_examples/semconv_registry/http-common.yaml
```diff
@@ -0,0 +1,87 @@
+groups:
```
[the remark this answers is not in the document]
lquerel: Okay, I removed the duplicate registry and the generated code example from the weaver_forge crate.
The existing tests/codegen.rs already tests parts of the generated code and checks things like:
- Does the generated code compile?
- Does the generated code expose the public API we expected?

crates/weaver_forge/src/debug.rs
lquerel: I don't remember the exact problem you observed with the release version.

jsuereth: So there were two issues:

1. The error output was truncated to the point of being useless, whereas in debug build you got line numbers and pointers at problematic templates.
2. The `{{ debug(var) }}` trick for templates stopped rendering.

crates/weaver_forge/codegen_examples/semconv_registry/http-common.yaml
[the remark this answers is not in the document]
jsuereth: Awesome.

crates/weaver_codegen_test/build.rs
```diff
@@ -0,0 +1,219 @@
...
+                    .file_stem()
+                    .expect("Failed to extract the file name")
+                    .to_str()
+                    .expect("Failed to convert to string");
+                let parent_modules = relative_path.parent().map_or(vec![], |parent| {
+                    let parent = parent.display().to_string();
+                    if parent.is_empty() {
+                        vec![]
+                    } else {
+                        parent.split('/').map(|s| s.to_owned()).collect::<Vec<_>>()
```
jsuereth, line 107: This line may not be windows friendly.  I think you should be able to split using `Path` in some fashion.

jsuereth (approved):
Approved, pending a windows build fix

crates/weaver_forge/src/debug.rs
lquerel: I will check that

crates/weaver_codegen_test/build.rs
lquerel: Fixed. Path.components() is the platform independent solution.

lquerel: Well not completely...

PR #139, opened 2024-05-01
Update weaver to understand deprecated enum values.

jsuereth:
Besides the stability marker, we now have a deprecation note we allow.


lquerel (approved):
LGTM

PR #142, opened 2024-05-02
Remove protoc install from github CI actions.

jsuereth:
Causing build failures and I don't think we actually use it.


lquerel (approved):
LGTM, Copy/Paste issue :-)

jsuereth:
FYI - the macosx test failures is due to the *cache* failing. I have a fix in my other PR.

PR #144, opened 2024-05-03
(feat): Add attribute_sort filter to weaver_forge

jsuereth:
Add a filter that lets us perform attribute sorting the way we need to for semconv templates.

Also document other filters/tests added previously.

### DIFF
--- a/crates/weaver_forge/src/extensions/otel.rs
+++ b/crates/weaver_forge/src/extensions/otel.rs
@@ -139,8 +139,7 @@ pub(crate) fn attribute_sort(input: Value) -> Result<Value, minijinja::Error> {
                 .iter()
                 .sorted_by(|lhs, rhs| {
                     // TODO - Actually output error message here or find another way to do this sorting...
-                    compare_requirement_level(lhs, rhs)
-                        .expect("Need a better way to fail on compare")
+                    compare_requirement_level(lhs, rhs).unwrap_or(std::cmp::Ordering::Less)
                 })
                 .to_owned()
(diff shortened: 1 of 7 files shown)


crates/weaver_forge/src/extensions/otel.rs
```diff
@@ -83,6 +85,72 @@ pub(crate) fn attribute_namespace(input: &str) -> Result<String, minijinja::Erro
...
+/// Sorts a sequence of attributes by their requirement_level, then name.
+pub(crate) fn attribute_sort(input: Value) -> Result<Value, minijinja::Error> {
+    input
+        .as_seq()
+        .map(|values| {
+            let result: Vec<Value> = values
+                .iter()
+                .sorted_by(|lhs, rhs| {
+                    // TODO - Actually output error message here or find another way to do this sorting...
+                    compare_requirement_level(lhs, rhs).unwrap_or(std::cmp::Ordering::Less)
```
lquerel, line 142: One way to report errors from this closure is by creating a mutable `Vec` of errors at the top-level scope of the function to collect all errors. Alternatively, replacing the `Vec` with an `Option` can be used to collect only the last observed error.

lquerel (approved):
I made a suggestion you could integrate before to merge this PR. 
As, it's not blocking, I approved the PR.

crates/weaver_forge/src/extensions/otel.rs
jsuereth: done!

PR #152, opened 2024-05-09
(fix) Expand id parsing

### LINKED ISSUE #151: model with id on name don't generate markdown docs
**Describe the bug**
When a model has a number on its id, the update markdown command doesn't work. It doesn't return any error, it just doesn't update the file.

**To Reproduce**
Example that doesn't work:
yaml file
```
- id: metric.namewithnumber1
    type: metric
    metric_name: namewithnumber1
```

readme file
```
<!-- semconv metric.namewithnumber1(metric_table) -->
<!-- endsemconv -->
```

Example that works:
yaml file
```
- id: metric.namewithnumber
    type: metric
    metric_name: namewithnumber1
```

readme file
```
<!-- semconv metric.namewithnumber(metric_table) -->
<!-- endsemconv -->
```

**Expected behavior**
Doc to be generated correctly or if there is a problem with the id name it should show some error message.

jsuereth:
Current parsing is a bit inflexible, improve parsing to include more.

Fixes #151 

This is a fast update.  Longer term, we should have a very "loose" parsing grammar for recognizing tags where we can issue errors if the tags don't match our current parser, due, e.g. to poor tag syntax.

### DIFF
--- a/crates/weaver_semconv_gen/src/parser.rs
+++ b/crates/weaver_semconv_gen/src/parser.rs
@@ -72,7 +72,10 @@ fn parse_markdown_gen_parameters(input: &str) -> IResult<&str, Vec<MarkdownGenPa
 
 /// nom parser for semconv ids.
 fn parse_id(input: &str) -> IResult<&str, &str> {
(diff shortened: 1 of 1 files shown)


lquerel (approved):
LGTM

PR #153, opened 2024-05-09
Detect and Process Policy Files into SemConv Registry + Generic Diagnostic Reporting

lquerel:
# Main updates

This PR introduces several enhancements:

1. **Policy Enforcement**: It detects policy files (`.rego`) embedded in semantic convention registries and uses the policy engine to enforce all found policies. In this PR, only the `before_resolution` policy stage is supported. 

2. **Enhanced Error Reporting**: All `WeaverError` instances are now extended to support Miette diagnostics, enabling rich annotations and improved error reporting. This extension is utilized by the diagnostic message rendering process described below.

3. **Generic Diagnostic Message Output**: A new mechanism for outputting diagnostic messages in various formats is introduced, including ANSI-compatible text, JSON, and `gh_workflow_command`. Users can create new formats by adding a Jinja template, which will be used to render diagnostic messages in the new format.

4. **Centralized Error Processing and Exit Code Determination**: All errors and diagnostic messages are now processed and rendered in the main function, which also determines the exit code.

The ANSI template renders policy violations in a specific way, while all other diagnostic messages are rendered using the Miette "graphical" renderer.

Below are the newly supported parameters that can be used to configure the diagnostic rendering process:

```
      --diagnostic-format <DIAGNOSTIC_FORMAT>
          Format used to render diagnostic messages. Predefined formats are: ansi, json, gh_workflow_command

          [default: ansi]

      --diagnostic-template <DIAGNOSTIC_TEMPLATE>
          Path to the directory where the diagnostic templates are located

          [default: diagnostic_templates]
```

# Other minor updates/improvements
- Used `assert_cmd` to test the command line.
- Removed the diagnostic channel infrastructure as we have decided not to pursue this direction.
- Removed all the telemetry schema examples which are no longer supported by this version of weaver.

# Future PRs
- Implementation of the `after_resolution` policy stage.
- Nice annotations on all errors/diagnostics.


jsuereth:
A few thoughts:

1. Ideally you shouldn't have a different set of policies for output, the rendering is just a matter of where you're running.  E.g. you'd have something like `weaver registry check ... --output console` in your makefile for your dev workflow and `weaver registry check ... --output gh_workflow` for the github workflow.  I don't think you be altering the policies along with the template output.  Not sure you called this out in your flow.

2. There's a lot of things in `weaver.yaml` you may want to share between outputs (e.g. acronyms). For now we can keep it separate, but I think eventually we may want something shared.

3. I assume the `registry` directory is because of the `registry` command.  Throwing diagnostics in there now feels kinda weird.  We could "flatten" the templates/ directory, or expand to have something like:

```
templates/
   registry/
       check/
          {output_target}/
       generate/
       snippet/
```

With corresponding `weaver registry {check|generate|snippet}` commands.  I personally would prefer a more flat structure, as I found I was sharing a lot of templates between the generate/markdown-update (maybe we should call this snippet?) commands.

crates/weaver_semconv_gen/src/gen.rs
```diff
@@ -62,20 +64,28 @@ fn write_example_list<Out: Write, Element: std::fmt::Display>(
...
+            Ok(write!(out, "`{value}`").map_err(|e| Error::StdIoError(e.to_string()))?)
+        }
+        Examples::Int(value) => {
+            Ok(write!(out, "`{value}`").map_err(|e| Error::StdIoError(e.to_string()))?)
+        }
+        Examples::Double(value) => {
+            Ok(write!(out, "`{value}`").map_err(|e| Error::StdIoError(e.to_string()))?)
+        }
+        Examples::String(value) => {
+            Ok(write!(out, "`{value}`").map_err(|e| Error::StdIoError(e.to_string()))?)
```
jsuereth, line 87: Nit:

You could probably make a single instance of `From<StdIoError> for Error` and you'd be able to remove all of these `map_err` cases

crates/weaver_common/src/diagnostic.rs
```diff
@@ -0,0 +1,208 @@
...
+}
+
+/// A list of diagnostic messages
+#[derive(Debug, serde::Serialize)]
+#[serde(transparent)]
+pub struct DiagnosticMessages(Vec<DiagnosticMessage>);
+
+impl DiagnosticMessage {
+    /// Creates a new diagnostic message from an error
+    pub fn new<M: Error + Diagnostic + Serialize + Send + Sync + 'static>(error: M) -> Self {
```
jsuereth, line 57: Do you need Send + Sync here (and throughout this file)?

crates/weaver_checker/src/lib.rs
```diff
@@ -81,6 +113,28 @@ impl WeaverError<Error> for Error {
     }
 }
 
+/// A list of supported policy stages.
+pub enum PolicyStage {
+    /// Policies that are evaluated before resolution.
+    BeforeResolution,
+    /// Policies that are evaluated after resolution.
+    AfterResolution,
```
jsuereth, line 121: Question:  Are we planning to write a "AgainstPreviousVersion" or some such that allows us to compare two semconv resolutions against each other?

crates/weaver_checker/src/lib.rs
```diff
@@ -81,6 +113,28 @@ impl WeaverError<Error> for Error {
...
+}
+
+impl Display for PolicyStage {
+    /// Returns the name of the policy stage.
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        match self {
+            PolicyStage::BeforeResolution => {
```
jsuereth, line 128: Nit: can you document PolicyStage and its usage in the README for weaver_checker?

deny.toml
```diff
@@ -50,14 +50,9 @@ allow = [
     "CC0-1.0",
     "ISC",
     "OpenSSL",
-    "zlib-acknowledgement",
     "Zlib",
 ]
-exceptions = [
```
jsuereth, line 56: why are these removed?

diagnostic_templates/ansi/errors.txt.j2
```diff
@@ -0,0 +1,13 @@
+{{ "Diagnostic report" | ansi_bold | ansi_red }}:
```
jsuereth, line 1: Are these files now required to be installed with weaver in some directory, or are they bundled into the executable in some clever fashion I'm not seeing?

jsuereth (commented):
So far this looks really good. There's a LOT to this so it took a while for me to sort through it all.

Have  one major question around what installation of weaver looks like with these diagnostic templates, but otherwise just a lot of nits / cleanups.

lquerel:
@jsuereth My goal in an upcoming PR is to embed the default templates within the binary to avoid depending on an external directory. However, a user will still be able to create an external directory to (re)define templates as they wish. We could even imagine an init command to initialize this directory with the default content.

crates/weaver_semconv_gen/src/gen.rs
lquerel: I thought about that and forgot... Thanks for the reminder !

crates/weaver_checker/src/lib.rs
lquerel: Yes. Refining the policy engine part is my next focus since I think the documentation/code generation part is in a usable state for most use cases.

crates/weaver_checker/src/lib.rs
lquerel: I will

crates/weaver_common/src/diagnostic.rs
lquerel: I will double check now that I removed the DiagChannel approach

deny.toml
lquerel: Because these exceptions were necessary only because of Tantivy. But I have temporarily removed this crate while waiting to reintroduce Tantivy with the current version of the semconvs.

diagnostic_templates/ansi/errors.txt.j2
lquerel: I wanted to embed them in the binary. I was just planning to do it in a separate PR. But if you prefer, I can take the time to do it in this one.

crates/weaver_semconv_gen/src/gen.rs
lquerel: Done in [5b064f2](https://github.com/open-telemetry/weaver/pull/153/commits/5b064f2144029278c43b4e715408cf1b9c08453b)

crates/weaver_common/src/diagnostic.rs
lquerel: This bound is required by `miette` for the creation of a `Report`. I think the idea for `miette` is to allow these reports to be transferred between threads.

diagnostic_templates/ansi/errors.txt.j2
jsuereth: IN the principle that "we should always be releasable", I'd prefer one of three things:


1. Documentation requiring them to exist somewhere / bundling them in the docker image.
2. Embed them in this PR.
3. Keep the functionality hidden / test-only until release is figured out.


I realize this is a higher bar, and may lead to this PR being larger.  I think the principle of "always releasable" is worth preserving though.

lquerel: Created this GH issue for tracking https://github.com/open-telemetry/weaver/issues/167

crates/weaver_checker/src/lib.rs
lquerel: Done [d2343a3](https://github.com/open-telemetry/weaver/pull/153/commits/d2343a32635148eaab9c4a93e0c6a8c76eed4073)

crates/weaver_checker/src/lib.rs
lquerel: I created a new GH issue for tracking, see https://github.com/open-telemetry/weaver/issues/168

jsuereth (approved):
Ok to merge as long as we consider the template issue a blocker for release.

PR #163, opened 2024-05-16
Additional filters and tests

lquerel:
This PR adds/updates the following filters:
- `instantiated_type`: Filters a type to return the instantiated type. 
- `enum_type`: Filters a type to return the enum type or an error if the type is not an enum.
- `capitalize_first`: Capitalizes the first letter of a string.
- `map_text` introduces a second parameter to define the default value if the name of the text map or the input are not found in the `text_maps` section (optional parameter).

This PR adds the following tests:
- `enum`: Tests if an attribute has an enum type.
- `simple_type`: Tests if a type is a simple type (i.e.: string | string[] | int | int[] | double | double[] | boolean | boolean[]).
- `template_type`: Tests if a type is a template type (i.e.: template[]).
- `enum_type`: Tests if a type is an enum type.

Documentation has been updated.
Unit tests have been created.


crates/weaver_forge/src/extensions/otel.rs
```diff
@@ -693,4 +767,142 @@ mod tests {
...
+                .unwrap(),
+            "enum"
+        );
+        assert_eq!(
+            env.render_str(
+                "{{ {allow_custom_values: true, members: []} | instantiated_type }}",
+                &ctx
+            )
+            .unwrap(),
+            "enum"
```
jsuereth, line 839: We actually should do better here.  We want to know the *type of the enum*, i.e. are the values of the enum booleans, doubles, ints or strings.

lquerel: I can implement this, although it seems that the Python version didn't do it. Isn't there also a potential stability issue with the APIs? For example, an enum containing only integers will switch to a string representation if we add a string variant to it. This is not necessarily a big problem at the instrumentation level, but it can have implications downstream in the system.

lquerel: Resolved, see [548edf7](https://github.com/open-telemetry/weaver/pull/163/commits/548edf7d62912674e37b16d0c0672def45c83d8c)

crates/weaver_forge/src/extensions/otel.rs
```diff
@@ -230,23 +257,146 @@ pub(crate) fn is_deprecated(input: Value) -> bool {
...
+                        value
+                    )));
+                }
+            };
+            inferred_type = match inferred_type {
+                Some(current_inferred_type) => {
+                    if current_inferred_type != member_type {
+                        // If the inferred type is different from the member type, then the enum
+                        // type is "promoted" to a string.
+                        Some("string".to_owned())
```
jsuereth, line 312: I believe this is the right way to go... We may need to improve enum in the model so type is actually specified AND enforced on members.

jsuereth (approved):

Issue #167, opened 2024-05-20
Embed Default Templates into Binary to Improve Self-Containment and Flexibility


lquerel:
Currently, diagnostic templates are located in a versioned directory and are part of the Docker image. To make the binary more self-contained, a better approach would be to embed the default templates into the binary of the app. This can be achieved with a macro like include_dir! along with some kind of abstraction to enable the template engine to work with both local and embedded directories.

This default template directory will not prevent Weaver's users from (re)defining their own local templates. An init command could be created to initialize such a template directory.

lquerel:
Implemented in #171

PR #171, opened 2024-05-23
Support for loading templates from the file system or from an embedded representation in the app's binary.

lquerel:
This PR contains the following changes:
- Integration of default diagnostic templates into the application binary.
- A new command `weaver diagnostic init` to locally create diagnostic templates, allowing the user to override them or define new targets.
- A mechanism to detect the presence of local templates for rendering diagnostic messages.
- An infrastructure to test the execution of CLI commands, their outputs, and their exit codes.

Closing: #167


Dockerfile
```diff
@@ -10,7 +10,6 @@ COPY crates /build/crates
 COPY data /build/data
 COPY src /build/src
 COPY tests build/tests
-COPY diagnostic_templates /build/diagnostic_templates
```
lquerel, line 13: No longer needed as the default templates are now embedded into the app's binary.

crates/weaver_diff/src/lib.rs
```diff
@@ -30,3 +34,85 @@ pub fn diff_output(original: &str, updated: &str) -> String {
     result.push_str(RESET);
     result
 }
+
+/// Displays differences between two directories and returns whether they are identical.
+/// The function will print differences to stderr.
+#[allow(clippy::print_stderr)]
+pub fn diff_dir<P: AsRef<Path>>(expected_dir: P, observed_dir: P) -> std::io::Result<bool> {
```
lquerel, line 41: Moved from weaver_forge

crates/weaver_forge/src/lib.rs
```diff
@@ -119,8 +98,12 @@ impl Object for TemplateObject {
...
 /// registry and telemetry schema.
 pub struct TemplateEngine {
-    /// Template path
-    path: PathBuf,
+    /// File loader used by the engine.
+    file_loader: Box<dyn FileLoader + 'static>,
+
+    /// The file loader function with a 'static lifetime (required by MiniJinja).
+    loader_function:
+        Arc<dyn for<'a> Fn(&'a str) -> Result<Option<String>, Error> + Send + Sync + 'static>,
```
lquerel, line 106: Use Arc because templates are loaded in parallel.

src/registry/update_markdown.rs
```diff
@@ -113,3 +109,41 @@ pub(crate) fn command(
...
+    // use weaver_common::TestLogger;
+    //
+    // use crate::cli::{Cli, Commands};
+    // use crate::registry::{RegistryArgs, RegistryCommand, RegistryPath, RegistrySubCommand};
+    // use crate::registry::update_markdown::RegistryUpdateMarkdownArgs;
+    // use crate::run_command;
+
+    #[test]
+    fn test_registry_update_markdown() {
+        // ToDo
```
lquerel, line 124: See my question in the chat.

crates/weaver_diff/src/lib.rs
jsuereth: Looks good.

crates/weaver_forge/src/lib.rs
jsuereth: I get that, but the design feels awkward to me.

1. This is exposing A LOT of details publicly - Arc, Error, Send/Sync, etc.
2. I'd prefer keeping raw types e.g. (T) with a trait you can use.  

For example, instead of returning a loader-function, why not just expose the function itself and keep the loader in an Arc if you want?  Then the owner of the `Loader` is responsible for keeping it in an arc/safely and generates their own thread-safe closure.

```
pub trait FileLoader {
    /// Returns a textual representation of the root path of the loader.
    /// This representation is mostly used for debugging and logging purposes.
    fn root(&self) -> &Path;

    /// Returns a list of all files in the loader's root directory.
    fn all_files(&self) -> Vec<PathBuf>;

    /// Loads a file from a given name
    fn load_file<'a>(
        &self,
        &'a str
    ) -> Result<Option<String>, Error>;
}
```
(caveat you may need to sprinkle some `Send+Sync` in there)

Then, in weaver_forge you can have an `Arc<dyn FileLoader>` if you need it, or you can otherwise synchronize threading access to the trait as you want.

TL;DR;  If this is a public trait, I think its impl is way too tiightly coupled to how we're doing threading today with little flexibility.

crates/weaver_forge/src/file_loader.rs
```diff
@@ -0,0 +1,357 @@
...
+        let mut files = vec![];
+        collect_files(self.embedded_dir, &mut files);
+        files
+    }
+
+    /// Returns a function that loads a file from a given name
+    fn file_loader(
+        &self,
+    ) -> Arc<dyn for<'a> Fn(&'a str) -> Result<Option<String>, Error> + Send + Sync + 'static> {
+        if let Some(fs_loader) = &self.fs_loader {
```
jsuereth, line 110: Should the presence of a file system load completely OVERRIDE looking up a file, or should we *try* to look up using the fs_loader and on failure look up from the embeded target?

crates/weaver_forge/src/lib.rs
lquerel: I agree, the interface is too complex. I'm going to simplify this as much as possible.

crates/weaver_forge/src/file_loader.rs
lquerel: It seems simpler to have an all-or-nothing logic in this specific case. A more granular approach risks involving more complicated logic when, for example, we want to remove a template defined in the embedded version but that should not be in the overloaded version.

crates/weaver_forge/src/lib.rs
lquerel: Done, see [3d86ccd](https://github.com/open-telemetry/weaver/pull/171/commits/3d86ccde248d43eee009ddb3b5a379f1d4e8d536)[3d86ccd](https://github.com/open-telemetry/weaver/pull/171/commits/3d86ccde248d43eee009ddb3b5a379f1d4e8d536)

crates/weaver_forge/src/file_loader.rs
```diff
@@ -0,0 +1,343 @@
...
+use std::{fs, io};
+
+use minijinja::ErrorKind;
+use walkdir::WalkDir;
+
+use crate::error::Error;
+use crate::error::Error::TargetNotSupported;
+
+/// An abstraction for loading files from a file system, embedded directory, etc.
+pub trait FileLoader: Send + Sync {
```
jsuereth, line 15: Nit: Can you add the `Send + Sync` requirement only in the `weaver_forge` where you need the type to be send+sync?

crates/weaver_forge/src/file_loader.rs
jsuereth: AH - since you're exposing all_files, I think I agree.  I wasn't seeing the full design before.

crates/weaver_forge/src/lib.rs
```diff
@@ -119,8 +98,8 @@ impl Object for TemplateObject {
 /// Template engine for generating artifacts from a semantic convention
 /// registry and telemetry schema.
 pub struct TemplateEngine {
-    /// Template path
-    path: PathBuf,
+    /// File loader used by the engine.
+    file_loader: Arc<dyn FileLoader + 'static>,
```
jsuereth, line 102: nit: Here can you add the `+Send + Sync`?

jsuereth (approved):
Just a nit.

jsuereth:
Looks like the docker build also needs tobe updated for the rename of the templates.

crates/weaver_forge/src/lib.rs
lquerel: See [9b8e7a7](https://github.com/open-telemetry/weaver/pull/171/commits/9b8e7a7564c059510f63b361f8336b485f5ef12c)

crates/weaver_forge/src/file_loader.rs
lquerel: See [9b8e7a7](https://github.com/open-telemetry/weaver/pull/171/commits/9b8e7a7564c059510f63b361f8336b485f5ef12c)

PR #177, opened 2024-05-28
Added support for List of Array examples.

jerbly:
The current implementation does not support all the Example types as defined here: https://github.com/open-telemetry/build-tools/blob/main/semantic-conventions/syntax.md#examples-for-examples - specifically multiple example values for array types.

I have added these as `ListOf*` variants to the `Examples` enum.

The legacy markdown generation had to be updated to accommodate this, and to fix the presentation of array type examples. An example for a `string[]` attribute type should be displayed as an array: `["a", "b"]`. In the current implementation they are shown as multiple examples appropriate for `string` type: `a`; `b`. I'm not sure where the legacy markdown writer is actually used in practice though.

This error in representation of examples has also been carried through to the jinja templates in the semconv project and so currently all the published examples for array types in the documentation are incorrect. (I have a PR to fix this going up ASAP).

Not sure what the pattern is for the changelog, it appears to show all unreleased entries and yet 0.3.0 has been released?? (But the main cargo.toml has 0.1.0??)


crates/weaver_semconv/src/attribute.rs
```diff
@@ -410,6 +410,14 @@ pub enum Examples {
     Bools(Vec<bool>),
     /// A array of strings example.
     Strings(Vec<String>),
+    /// List of arrays of integers example.
+    ListOfInts(Vec<Vec<i64>>),
```
jsuereth, line 414: Could you add a test to ensure this is working?

crates/weaver_semconv_gen/src/gen.rs
```diff
@@ -58,10 +58,15 @@ fn enum_type_string(members: &[EnumEntriesSpec]) -> &'static str {
     }
 }
 
-fn write_example_list<Out: Write, Element: std::fmt::Display>(
+fn write_example_list<Out: Write, Element: std::fmt::Display + std::fmt::Debug>(
```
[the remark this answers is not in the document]
jerbly: Agreed. If this is going to be deleted then I'll leave it... I couldn't see where it was actually used.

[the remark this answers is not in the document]
jerbly: Agreed. If this is going to be deleted then I'll leave it... I couldn't see where it was actually used.

crates/weaver_semconv/src/attribute.rs
jerbly: I've added tests... feels like I'm just testing serde_yaml though ;)

crates/weaver_checker/README.md
```diff
@@ -105,10 +105,8 @@ deny[schema_evolution_violation("attr_removed", old_group.id, old_attr.id)] {
 }
 ```
 
-> [!NOTE]
```
lquerel, line 108: These [!NOTE] are valid GitHub Markdown Extensions, see the following blog post.

https://docs.github.com/fr/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts

lquerel (changes requested):
Overall, it LGTM, but the [!NOTE] tags need to be put back in place because they are interpreted by GitHub in a specific way. It is possible that `rustdoc` or `clippy` might not be happy, in which case you should use a #[allow(<something>)] to suppress the warnings.

lquerel:
@jerbly The issue with `check-external-types` will be fixed if you update your branch with the last version of the main branch.

lquerel (approved):
Thank you for the updates.

crates/weaver_semconv/src/attribute.rs
jsuereth: Thanks, and agreed. I had intended for those tests to be on full resolution ensuring these are preserved through the resolution process, but what you have is fine for now.

jsuereth (approved):

PR #179, opened 2024-05-29
Add distribution  (binaries + installers) publishing workflows.

### LINKED ISSUE #7: Automate GitHub releases for the project
Use `cargo-dist` or similar tools to automatically generate releases.
Ideally, each release should contain binaries for the three major operating systems (Linux, macOS, and Windows).

jsuereth:
- Update repository + author configuration for OpenTelemetry.
- Install / run `cargo dist init`
  - Initial set up for Linux/OSX/Windows
  - Add wix (MSI) installer
  - Add shell script installers.
  
We may need to do some finagling of github configuration to get these to work on tag/release.  I'll work on that during the next release of weaver.

Note: This does not touch the docker release functionality.  I may be updating/moving that next.

Fixes #7

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,271 @@
+# Copyright 2022-2024, axodotdev
+# SPDX-License-Identifier: MIT or Apache-2.0
+#
+# CI that:
+#
+# * checks for a Git Tag that looks like a release
+# * builds artifacts with cargo-dist (archives, installers, hashes)
+# * uploads those artifacts to temporary workflow zip
+# * on success, uploads the artifacts to a GitHub Release
+#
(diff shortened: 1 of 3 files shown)


lquerel (approved):
I'm not familiar with WiX, but otherwise LGTM.

jsuereth:
> I'm not familiar with WiX, but otherwise LGTM.

[I have a bit of experience with it, from a while ago](https://github.com/sbt/sbt-native-packager/blob/master/src/main/scala/com/typesafe/sbt/packager/windows/WixHelper.scala). If we have issues, I'm confident I can get us into a working state.  Windows is an... interesting packaging experience.  It has all the components of a great package manager but no one ever tied the knots to make it great.  Wix makes it bearable.

Issue #184, opened 2024-05-29
When resolving to stdout do not display status


MadVikingGod:
**Describe the bug**
When the output is stdout, the status lines (example below) will break any kind of stream processing, such as passing it to `jq`.

example:
```
✔ SemConv registry loaded (111 files)
✔ No policy found
✔ SemConv registry resolved
```

**To Reproduce**
1. Run `weaver registry resolve -f json | jq .`
2. watch it fail because of the status printed at the beginning

**Expected behavior**
The resolve output should be in the correct format, free of any status outputs.  

**Additional context**
These statuses are nice if we are outputting to a file or if they were on stderr.

lquerel:
`--quiet` is no longer required with #187 (not yet approved at this point of time).

MadVikingGod:
Is there any way to promote this option to A) show up in the `weaver registry resolve --help`, and B) Be usable from the end of the command?

For B) `weaver --quiet registry resolve` works, but `weaver registry resolve --quiet` does not.

lquerel:
Yes, it's definitely something to improve in the user experience. We use Clap for command-line parsing, and while there are some limitations, there are workarounds to ensure that both A and B are feasible.

lquerel:
@MadVikingGod with #187 merged `--quiet` is no longer required. 

However, I'd like to keep this issue open as I will make some changes to accept the `--quiet` argument in a more discoverable position.

lquerel:
Fixed in #196 (pending approval).

jerbly:
@lquerel - this is not fixed. We have a general problem with the separation of artifacts from diagnostics when using stdout. Most registry commands can now produce an artifact to stdout for downstream consumption e.g `resolve`, `diff`, `stats`, `json-schema`, `live-check`.

To cleanly separate artifacts from diagnostics we could send diagnostics to `stderr`. This means that `weaver registry resolve -f json | jq .` works AND shows you warnings from the diagnostics.

I believe this is a neat solution for all registry commands and would make the code and behaviour consistent.

The trade-off here is that for `check` the diagnostic output is also the "artifact". So perhaps for this special case, we could add `--diagnostic-stdout` to override the default and switch diagnostics to stdout. And perhaps `check` always switches to this mode. (Note, the current Semantic Conventions Makefile does not use the output, only the exit-code).

WDYT?

lquerel:
[...]

PR #187, opened 2024-05-30
Generate JSON Schema for both Resolved Telemetry Schema and Resolved Registry

lquerel:
This PR uses [schemars](https://github.com/GREsau/schemars) to automatically generate JSON schema from Serde-annotated Rust structs. 

**This PR will serve as a basis for future discussions and the generation of documentation on the topic of resolved registry and schema.**

Two types of resolved assets coexist:
- **Resolved Registry**: A resolved view of a semantic convention registry. This is basically a resolved/flattened version of the semantic convention files. The command `weaver registry resolve` can be used to generate this type of resolved asset. The command `weaver registry json-schema` generates the JSON Schema representation of a `ResolvedRegistry` data structure.
- **Resolved Telemetry Schema**: An extension of the existing Telemetry Schema containing a catalog of attributes, registries, etc., in addition to the original `versions` section. This type of schema is still in discussion.

This PR also improves the output of commands such as `weaver registry resolve`, which can now be directly piped to a `jq` command without having to use the `--quiet` parameter (see #189).

Note 1: `TemplateRegistry` and `TemplateGroup` have been renamed to `ResolvedRegistry` and `ResolvedGroup` because it's a more logical name. This renaming has no impact on the templates and the policies. I might move these definitions into a dedicated crate.

Note 2: A conversation has been initiated to talk about resolved registry and resolved telemetry schema, see https://cloud-native.slack.com/archives/C0697EXNTL3/p1717183349294419?thread_ts=1717104224.284179&cid=C0697EXNTL3


crates/weaver_common/src/lib.rs
```diff
@@ -47,6 +47,9 @@ pub trait Logger {
 
     /// Logs a message without icon.
     fn log(&self, message: &str);
+
+    /// Mute all the messages except for the warnings and errors.
+    fn mute(&self);
```
lquerel, line 52: Introduced this new function to fix #184 and to simplify testing.

src/diagnostic/init.rs
```diff
@@ -3,7 +3,7 @@
 //! Initializes a `diagnostic_templates` directory to define or override diagnostic output formats.
 
 use crate::diagnostic::{Error, DEFAULT_DIAGNOSTIC_TEMPLATES};
-use crate::DiagnosticArgs;
+use crate::{DiagnosticArgs, ExitDirectives};
```
lquerel, line 6: ExitDirectives is a way for commands to inform the main program on how it should behave once the command has been executed (e.g., suggested exit code, quiet mode).

src/format.rs
```diff
@@ -0,0 +1,25 @@
+// SPDX-License-Identifier: Apache-2.0
+
+//! Supported output formats for the resolved registry and telemetry schema
+
+use clap::ValueEnum;
+use serde::Serialize;
+
+/// Supported output formats for the resolved schema
+#[derive(Debug, Clone, ValueEnum)]
+pub(crate) enum Format {
```
lquerel, line 10: Moved `Format` in a more global location as this structure is now used from multiple sub-commands.

src/registry/resolve.rs
```diff
@@ -78,7 +65,10 @@ pub(crate) fn command(
     logger: impl Logger + Sync + Clone,
     cache: &Cache,
     args: &RegistryResolveArgs,
-) -> Result<(), DiagnosticMessages> {
+) -> Result<ExitDirectives, DiagnosticMessages> {
```
lquerel, line 68: Now this command can only display the resolved registry. See `weaver schema resolve` to display the resolved telemetry schema (this part is still experimental).

crates/weaver_common/src/lib.rs
```diff
@@ -55,6 +58,7 @@ pub trait Logger {
 pub struct ConsoleLogger {
     logger: Arc<Mutex<paris::Logger<'static>>>,
     debug_level: u8,
+    mute: Arc<Mutex<bool>>,
```
jsuereth, line 61: I'd just use an atomic here.

crates/weaver_common/src/lib.rs
```diff
@@ -64,14 +68,16 @@ impl ConsoleLogger {
...
+            mute: Arc::new(Mutex::new(false)),
         }
     }
 }
 
 impl Logger for ConsoleLogger {
     /// Logs an trace message (only with debug enabled).
     fn trace(&self, message: &str) {
-        if self.debug_level > 0 {
+        let mute = *self.mute.lock().expect("Failed to lock mute");
```
jsuereth, line 79: Why do we need to lock? Just check current value.

Also if you set this early, you may even be able to lazy-read in practice...

src/schema/mod.rs
```diff
@@ -0,0 +1,60 @@
...
+pub enum SchemaSubCommand {
+    /// Resolves a telemetry schema.
+    ///
+    /// Rego policies present in the registry or specified using -p or --policy will be automatically validated by the policy engine before the artifact generation phase.
+    ///
+    /// Note: The `-d` and `--registry-git-sub-dir` options are only used when the registry is a Git URL otherwise these options are ignored.
+    ///
+    /// The process exits with a code of 0 if the resolution is successful.
+    #[clap(verbatim_doc_comment)]
+    Resolve(SchemaResolveArgs),
```
jsuereth, line 35: I'm not sure I understand what this is meant to do.

jsuereth (commented):
Are we declaring the difference between schema + registry the difference between using optimisation on attributes vs. not?

At this point, I'm still a bit confused on terminology and I think maybe we should discuss that first.

Also have a minor nit about threading choices.

crates/weaver_common/src/lib.rs
lquerel: I agree, a mutex is overkill in this context. An atomic is indeed enough, I don't know why I didn't use it in the first place.

src/schema/mod.rs
lquerel: We need to discuss that in a more interactive way.

lquerel:
> At this point, I'm still a bit confused on terminology and I think maybe we should discuss that first.

Yes, I think the best way to discuss this is through chat or via Zoom.

crates/weaver_common/src/lib.rs
lquerel: Replaced the mutex by an atomic in [d14e711](https://github.com/open-telemetry/weaver/pull/187/commits/d14e711c0028d10c21b260cbf287db549b6dc8d4)

crates/weaver_common/src/lib.rs
lquerel: Replaced the mutex by an atomic in [d14e711](https://github.com/open-telemetry/weaver/pull/187/commits/d14e711c0028d10c21b260cbf287db549b6dc8d4)

src/schema/mod.rs
lquerel: Conversation initiated here https://cloud-native.slack.com/archives/C0697EXNTL3/p1717183349294419?thread_ts=1717104224.284179&cid=C0697EXNTL3

jsuereth (approved):

PR #188, opened 2024-05-30
Create a scaffold for a search app.

jsuereth:
This just sets up an initial architecture with all components in place to make it easier for someone to learn Ratatui/Crosstool with a resolved schema and implement searching features.

The goal of this is NOT completeness. It's meant to be "bare bones" enough someone can take this and make it whatever in the future, but also NOT a burden for us to evolve weaver going forward if it exists.

Hopefully someone wanting to learn Rust and/or Weaver sees this and decided to contribute/drive this forward.


lquerel:
This is the commit where I removed the search command before the release 0.1.0. That should help initiate this task.

https://github.com/open-telemetry/weaver/commit/96e2c134214c237cf3c70e57b90ed7dbd08763d1

src/registry/search.rs
```diff
@@ -17,7 +40,96 @@ pub struct RegistrySearchArgs {
...
+    let registry_path =
+        semconv_registry_path_from(&args.registry.registry, &args.registry.registry_git_sub_dir);
+
+    // Load the semantic convention registry into a local cache.
+    let semconv_specs = load_semconv_specs(&registry_path, cache, logger.clone())?;
+    let mut registry = SemConvRegistry::from_semconv_specs(registry_id, semconv_specs);
+    let schema = resolve_semconv_specs(&mut registry, logger.clone())?;
+
+    // TODO - We should have two modes:
+    // 1. An interactive UI
```
lquerel, line 132: For the interactive mode, see this commit https://github.com/open-telemetry/weaver/commit/96e2c134214c237cf3c70e57b90ed7dbd08763d1 path=src/search.

Cargo.toml
```diff
@@ -74,6 +74,9 @@ weaver_forge = { path = "crates/weaver_forge" }
 weaver_checker = { path = "crates/weaver_checker" }
 
 clap = { version = "4.5.4", features = ["derive"] }
+rayon = "1.10.0"
+ratatui = { version = "0.26.3", features=["serde"] }
+crossterm = { version = "0.27.0", features = ["serde"] }
```
lquerel, line 79: I used Tantivy for the search

jsuereth:
@lquerel This is ready for review.  See my comments to your initial comments.  The goal of this is NOT completeness.  It's meant to be "bare bones" enough someone can take this and make it whatever in the future, but also NOT a burden for us to evolve weaver going forward if it exists.

This is why I didn't opt for anything more than iterator-based searching and only support displaying Attributes for now.  We should still have room to alter data model without updating rendering code here, which I expect we'll need as we worked towards "application schema" stabilziation.

src/registry/search.rs
```diff
@@ -17,7 +44,358 @@ pub struct RegistrySearchArgs {
...
+        let title_block = Block::default()
+            .borders(Borders::TOP)
+            .style(Style::default().bg(Color::Black))
+            .border_style(Style::default().fg(Color::Gray))
+            .title_alignment(ratatui::layout::Alignment::Center)
+            .title_style(Style::default().fg(Color::Green))
+            .title("Weaver Search");
+        let title_contents = Line::from(vec![Span::styled(
+            format!(
+                "Loaded {0:?} registries w/ {1} attributes",
```
lquerel, line 111: Is the registry the entire set of groups starting with registry. or is each group starting with registry. considered an individual registry? If as I think it's the former then message should be something like "Loaded {0:?} groups w/ {1} attributes".

lquerel (approved):
See my comment regarding one of the messages displayed to the console.

Issue #193, opened 2024-06-03
Support command line parameters to add an additional layer of configurability in the documentation/code generator.


lquerel:
In #192, Liudmila expresses the need to make the generation of incubating assets optional while simultaneously reducing the level of redundant definitions in the templates. This issue is a proposal to solve those problems.

Let’s start by introducing two new arguments to the weaver registry generate command: `--param` and `--params`.

To specify one or several template parameters directly from the command line:
`weaver registry generate --param incubating=true --param excluded="[ios, jvm]" <target> <output-dir>`

To specify a YAML file containing a set of parameters:
`weaver registry generate --params <yaml-params-file> <target> <output-dir>`

The structure of this YAML file is as follows:
```yaml
params:
  incubating: true
  excluded: [ios, aspnetcore, signalr, android, dotnet, jvm]
```

> Note: the value of a param can be any valid YAML value.

In the `weaver.yaml`, we will also introduce a new optional section named `params`. The parameters provided via the command line will override the ones defined in the `weaver.yaml` params section. This way, we can provide default values for params in the `weaver.yaml` file.

A new global variable will be introduced in the evaluation context of the templates. We currently have two global variables, `ctx` and `template`. We will introduce a new one named `params`, which will expose the parameters defined in `weaver.yaml` and potentially overridden via the command line.

These parameters will be directly available in the templates, and conditional logic can be used to influence the generation of code or documentation.

```
...
{% if params.incubating %}
... generate incubating code ...
{% endif %}
...
```

The JQ expression has also access to the parameters via the $var notation. In the following example, the `$excluded` JQ global variable is defined in the params. 

```yaml
templates:
  - pattern: <glob-pattern>
    filter: >
      if $incubating then
        .groups
          | map(select(.type == "attribute_group"))
          | map(select(.id | startswith("registry")))
          | map({ id: .id, group_id: .id | split(".") | .[1], attributes: .attributes })
          | group_by(.group_id)
          | map({ id: .[0].group_id, attributes: [.[].attributes[]] | sort_by(.id), output: "_incubating/attributes/", stable_package_name: "opentelemetry.semconv.attributes" })
          | map(select(.id as $id | any($excluded[]; . == $id) | not))
          | map(select(.attributes | length > 0))
      else
        empty
      end
     application_mode: single | each
```

In the context of #192 and with this improvement in place, it will be easy to create a single target `python` to generate either stable artifacts or both `stable` and `incubating` artifacts based on the parameters provided in the command line.

[...]

PR #195, opened 2024-06-04
[WIP] Support command line parameters to add an additional layer of configurability in the documentation/code generator.

lquerel:
Closing: #193


crates/weaver_forge/src/filter.rs
```diff
@@ -17,8 +16,9 @@ pub struct Filter {
 impl Filter {
     /// Create a new filter from a string expression or return an error if the
     /// expression is invalid.
-    pub fn try_new(filter_expr: &str) -> Result<Self, Error> {
-        let vars = Vec::new();
+    /// The vars parameter is a list of variable names that can be used in the
+    /// filter expression.
+    pub fn try_new(filter_expr: &str, vars: Vec<String>) -> Result<Self, Error> {
```
jsuereth, line 21: Nit: Should we pass in a `Map<String,Value>` here for vars?  It feels wierd to have two lists that are dependent on each other in different places.

crates/weaver_forge/src/lib.rs
```diff
@@ -95,6 +98,37 @@ impl Object for TemplateObject {
...
+        for (key, value) in params {
+            _ = new_params.insert(key, Value::from_serialize(&value));
+        }
+        Self { params: new_params }
+    }
+}
+
+impl Display for ParamsObject {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        f.write_str(&format!("{:#?}", self.params))
```
jsuereth, line 120: Nit:  I'm not a fan of relying on debug formatting...  Can/should we create a formatter for HashMap and use it here and/or just a helper method?

crates/weaver_forge/src/filter.rs
lquerel: I agree, it's weird but it's the jaq API...

crates/weaver_forge/src/lib.rs
lquerel: This `Display` implementation is unused. I will remove it in the next commit.

jsuereth (approved):

Issue #181, opened 2024-05-29
Identify the first version of semconv that is interpretable by Weaver


lquerel:
Add this information to the documentation.

MadVikingGod:
TLDR: v1.19.0-v1.26.0 work.

I explored this a bit and found:
from [open-telemetry/semantic-conventions](https://github.com/open-telemetry/semantic-conventions/tree/main/model) these worked:

- [v1.21.0](https://github.com/open-telemetry/semantic-conventions/tree/v1.21.0/model)
- [v1.22.0](https://github.com/open-telemetry/semantic-conventions/tree/v1.22.0/model)
- [v1.23.0](https://github.com/open-telemetry/semantic-conventions/tree/v1.23.0/model)
- [v1.24.0](https://github.com/open-telemetry/semantic-conventions/tree/v1.24.0/model)
- [v1.25.0](https://github.com/open-telemetry/semantic-conventions/tree/v1.25.0/model)
- [v1.26.0](https://github.com/open-telemetry/semantic-conventions/tree/v1.26.0/model)

from [open-telemetry/opentelemetry-specification](https://github.com/open-telemetry/opentelemetry-specification/tree/v1.20.0/semantic_conventions) these work:

- [v1.19.0](https://github.com/open-telemetry/opentelemetry-specification/tree/v1.19.0/semantic_conventions)
- [v1.20.0](https://github.com/open-telemetry/opentelemetry-specification/tree/v1.20.0/semantic_conventions)

v1.18.0 failed with the following error:
```
/weaver registry resolve -r ../opentelemetry-specification/semantic_conventions -f yaml
⠇ Resolving registry `../opentelemetry-specification/semantic_conventions`Diagnostic report:

  × The semantic convention spec is invalid (path_or_url: "../opentelemetry-specification/semantic_conventions/
  │ trace/http.yaml"). groups[2].attributes: data did not match any variant of untagged enum AttributeSpec at line
  │ 135 column 7


Total execution time: 0.100291173s
```
I don't know if this is a simple error to bypass.

lquerel:
I believe the issue comes from `requirement_level: optional` used for one of the attributes. This is not a variant supported by Weaver. My experience with SemConv isn't extensive enough to know if this variant was removed or renamed, but that's probably what happened. If necessary, we could add this variant and issue a warning. @MadVikingGod and @jsuereth What do you think?

MadVikingGod:
[...]

Issue #185, opened 2024-05-29
[Call For Participation] List of policies we would like to implement with Weaver’s policy engine


lquerel:
This list aims to:
1) Keep track of the policies identified by the community.
2) Identify the sources of information needed to define the policies in Rego (e.g., semconv before resolution, resolved telemetry schema).

Identified policies:
- We only allow new attribute definitions inside the attribute registry; all other attributes must be references.
- All groups in the attribute registry MUST be of type attribute_group.
- The attribute registry is not allowed to contain references.
- An attribute can only belong to one group.
- Implement a check to prevent referencing the same attribute multiple times within one group (see https://github.com/open-telemetry/weaver/issues/175).
- Checking for naming collisions between namespaces and attribute names (see https://github.com/open-telemetry/semantic-conventions/issues/1068).

jsuereth:
- A template attribute forms a namespace, and we should not allow attributes below it.
  e.g. `http.request.headers.<key>` is implied by the template attribute `http.request.headers` (this may be done by semconv#1068)
- We do not allow the removal of an attribute once added to the registry.  It must exist SOMEWHERE in a group.
- We do not allow the removal of a metric name once added to the registry (the group can move around, be deprecated, but its "name" must be exist)

jsuereth:
@lquerel should we close this issue now?

PR #199, opened 2024-06-05
Add optional variant to requirement_level

MadVikingGod:
This enables weaver to parse semantic convention v1.18.0 

This will enable parsing of semantic conventions v1.13.0 - v1.18.0

I was looking to put a test in `crates/weaver_resolver/data`, but I don't know if that's appropriate. And, I don't fully grok the format of what I should put there to make sure that it is parsing appropriately.

I also don't know how to signify that this is a deprecated option and shouldn't be used in general cases.

Expands the supported versions found in https://github.com/open-telemetry/weaver/issues/181


lquerel (approved):
LGTM

lquerel:
> I also don't know how to signify that this is a deprecated option and shouldn't be used in general cases.

@MadVikingGod Ideally, I think we should generate a violation via the policy engine when the semconv registry is greater than or equal to v1.19.0. However, the version number is not in the YAML docs themselves but in the form of a tag. So, we'll need to find a way to expose it to the policies one way or another. We should create a GH issue for that.

jsuereth (approved):

lquerel:
Added the policy into this list -> https://github.com/open-telemetry/weaver/issues/185#issuecomment-2152719505

PR #202, opened 2024-06-10
Add display_name field

joaopgrassi:
Follow up from https://github.com/open-telemetry/semantic-conventions/pull/985#issuecomment-2092432528

Creating this as a draft, as I'm not sure what exactly has to be changed to get the new field working in Weaver.


lquerel (approved):
LGTM

PR #209, opened 2024-06-12
New entry in developer guide to describe the process of adding new fields in the semantic convention registry

lquerel:
No change in the code.

This PR is to support https://github.com/open-telemetry/weaver/pull/202 and any future similar additions to the semconv registry definition.


joaopgrassi (approved):
Thanks for adding this! LGTM

jsuereth (approved):

PR #214, opened 2024-06-19
Add support for post-resolution policies

lquerel:
With this PR, it is now possible to define policies to validate a resolved semconv registry (the previous version of Weaver was only supporting pre-resolution policies). 

As before, all the `*.rego` files are imported into the policy engine. The package names are used to identify groups of rules to apply either before or after the resolution process (`package before_resolution` vs `package after_resolution`). It is valid to define multiple policy files with the same package name. In this case, the rules defined in these files will be combined and evaluated together. This option can be used to logically split the rules per topic or domain of control.

Note: The definition of policies comparing two versions of the resolved semconv will be added in a future PR.


crates/weaver_common/src/diagnostic.rs
```diff
@@ -139,6 +151,43 @@ impl DiagnosticMessages {
     }
 }
 
+/// An extension trait for `Result` that captures the diagnostic messages
+pub trait ResultExt<T, E> {
+    /// Captures the diagnostic messages into the provided `DiagnosticMessages`
+    /// or returns the value if there are no diagnostic messages.
+    fn capture_diag_msgs_into(self, diags: &mut DiagnosticMessages) -> Option<T>;
```
jsuereth, line 158: I like these additions.  Will need to figure out if we have the right set of combinators here.

src/registry/check.rs
```diff
@@ -51,19 +63,63 @@ pub(crate) fn command(
...
+        )?)
+    } else {
+        None
+    };
+
+    if let Some(policy_engine) = policy_engine.as_ref() {
+        // Check the policies against the semantic convention specifications before resolution.
+        // All violations should be captured into an ongoing list of diagnostic messages which
+        // will be combined with the final result of future stages.
+        _ = check_policies(policy_engine, &semconv_specs, logger.clone())
```
jsuereth, line 81: Nit: There's something about ignoring results that bothers me in this PR.  Can you add to the comment above it WHY ignoring the result is ok here?

src/util.rs
```diff
@@ -17,7 +18,12 @@ use weaver_resolver::SchemaResolver;
 use weaver_semconv::registry::SemConvRegistry;
 use weaver_semconv::semconv::SemConvSpec;
 
-/// Convert a `RegistryPath` to a `weaver_semconv::path::RegistryPath`.
+/// Converts a `RegistryPath` to a `weaver_semconv::path::RegistryPath`.
+///
+/// # Arguments
+///
+/// * `registry` - A reference to the `RegistryPath`.
+/// * `path` - An optional string representing the path.
```
jsuereth, line 26: Nit: Lets use better language here:

```
* registry - A reference to a registry of telemetry schema.  This is considered identifying for that registry, e.g a git url or local file path.
* path: An optional string representing a sub-directory in the registry identifying path where model/yaml files are located.
```

jsuereth (approved):
Just minor nits and wording concerns.

jsuereth:
One other comment, code coverage is pretty poor here.  We need to figure out *negative* tests where we expect failures and diagnostic messages to print.

src/util.rs
lquerel: Fixed in https://github.com/open-telemetry/weaver/pull/214/commits/f28cd820ac47f984ea536111b549d655fe591bb2

src/registry/check.rs
lquerel: Fixed in https://github.com/open-telemetry/weaver/pull/214/commits/f28cd820ac47f984ea536111b549d655fe591bb2

lquerel:
> One other comment, code coverage is pretty poor here.  We need to figure out *negative* tests where we expect failures and diagnostic messages to print.

I improved the comments, but regarding the tests, it's not easy to add more from my iPad. The overall coverage is still greater than 70%, so I'm going to merge this PR. I will create a GitHub issue to address this in a future PR once I'm back.

PR #224, opened 2024-06-27
Add Jinja whitespace control

### LINKED ISSUE #220: Add whitespace control in Jinja templates
Minijinja allows controlling how whitespaces are rendered in templates. See https://docs.rs/minijinja/latest/minijinja/syntax/index.html#whitespace-control.

To expose this in Weaver, we need to:
- Add a 'whitespace_control' section in 'weaver.xml'. This section should be defined in 'crates/weaver_forge/src/config.rs', specifically in the 'TargetConfig' struct. Use the 'template_syntax' section as a model.
- In the same crate's 'lib.rs', initialize the engine in the 'template_engine' method. The configuration file representation is accessible there.

joaopgrassi:
Closes #220

I added the new properties to the existing struct `TemplateSyntax` but I'm not so convinced that's the best place. I can see how it makes sense as the "Jinja config things" but 🤔. I'm open to suggestions. Maybe a different struct to configure jinja environment "in general"? Or maybe rename `TemplateSyntax` to something else, so it not only contain syntax things but general Jinja settings?


joaopgrassi:
weird sometimes the test fails also on my machine for the line end of the file. I will look into it

lquerel (changes requested):
Instead of integrating it into 'TemplateSyntax', I propose creating a separate structure in parallel that we could name 'WhitespaceControl', containing the two fields you created. Otherwise your PR LGTM. Thanks!

crates/weaver_forge/src/config.rs
```diff
@@ -298,6 +301,49 @@ impl Default for TemplateSyntax {
...
 }
 
+/// Whitespace control configuration for the template engine.
+#[derive(Deserialize, Debug, Clone)]
+pub struct WhitespaceControl {
+    /// Configures the behavior of the first newline after a block.
+    /// See <https://docs.rs/minijinja/latest/minijinja/struct.Environment.html#method.set_trim_blocks>
+    #[serde(default = "default_trim_blocks")]
+    pub trim_blocks: bool,
+    /// Configures the behavior of the first newline after a block.
```
lquerel, line 311: This line is probably a leftover from a copy/paste.

crates/weaver_forge/src/config.rs
```diff
@@ -298,6 +301,49 @@ impl Default for TemplateSyntax {
     }
 }
 
+/// Whitespace control configuration for the template engine.
+#[derive(Deserialize, Debug, Clone)]
+pub struct WhitespaceControl {
+    /// Configures the behavior of the first newline after a block.
+    /// See <https://docs.rs/minijinja/latest/minijinja/struct.Environment.html#method.set_trim_blocks>
+    #[serde(default = "default_trim_blocks")]
```
lquerel, line 309: Since you have provided an explicit implementation of the Default trait, these individual declarations are redundant.

crates/weaver_forge/src/config.rs
```diff
@@ -298,6 +301,49 @@ impl Default for TemplateSyntax {
...
+    #[serde(default = "default_lstrip_blocks")]
+    pub lstrip_blocks: bool,
+    /// Configures whether trailing newline are preserved when rendering templates.
+    /// See <https://docs.rs/minijinja/latest/minijinja/struct.Environment.html#method.set_keep_trailing_newline>
+    #[serde(default = "default_keep_trailing_newline")]
+    pub keep_trailing_newline: bool,
+}
+
+/// Default trim_blocks behavior.
+fn default_trim_blocks() -> bool {
```
lquerel, line 323: See my previous comment on the Default trait. These functions can be removed.

lquerel (approved):
Just a few minor modifications to make, but otherwise, it seems ready to be merged.

crates/weaver_forge/src/config.rs
trisch-me, line 309: I thought if we want default to be false/0/empty string etc we don't need to define anything as it's default behaviour

jsuereth (approved):
I echo @lquerel's comments on default trait.  Once that's cleaned up we can merge.

joaopgrassi:
If I got it right, I don't need the defaults annotations and functions, and I also removed the constructor. Sorry still very noob in Rust :D. Hope it is alright now.

trisch-me (approved):

joaopgrassi:
Not sure why the build is stuck? 🤔. I don't have permission to retry it.

trisch-me:
it's not stuck, I think it's because your test has failed

PR #231, opened 2024-07-06
Support a more flexible configuration overriding mechanism.

lquerel:
## Rationale

**Hierarchical Configuration Needs**: There is a requirement to support configurations that are shared across multiple language variants, teams, etc. Introducing hierarchical definitions for the Weaver configuration will allow splitting complex configurations into reusable components.

## Description

This PR includes the following changes:

- By default, the Weaver configuration is built from the following `weaver.yaml` files:
  - The configuration located in `$HOME/.weaver/weaver.yaml` (if it exists).
  - All `weaver.yaml` files from the root to the directory `templates/registry/<target>` or specified by the combination of the `--templates` parameter and the positional target parameter. This approach mirrors the `cargo` command's handling of `Cargo.toml` files.
- **New `--config` Parameter**: This optional parameter allows specifying one or more `weaver.yaml` files to load before generating templates. Multiple configuration files can be listed, and their order is crucial as overrides are applied sequentially (i.e., file n overrides file n-1, etc.).
- The `type_mapping` filter, which had been deprecated for several releases, has been removed.

More details have been introduced in the documentation of the project.


crates/weaver_forge/src/config.rs
```diff
@@ -387,40 +307,464 @@
...
+            },
+            whitespace_control: Default::default(),
+            params: None,
+            templates: None,
+            acronyms: None,
         }
     }
+}
+
+impl WeaverConfig {
```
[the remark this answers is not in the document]
lquerel: Fixed

crates/weaver_forge/src/config.rs
```diff
@@ -387,40 +307,464 @@
...
+    /// Builds the Weaver configuration from a path pointing to a configuration file.
+    /// The returned configuration is the result of the resolution of all the configurations found
+    /// in:
+    /// - the $HOME/.weaver/weaver.yaml file,
+    /// - the <path>/weaver.yaml and its <parent folders>weaver.yaml.
+    ///
+    /// The resolution process is done in the order of the enumeration above.
+    /// When a configuration is not found, it is skipped.
+    pub fn from_path<P: AsRef<Path>>(path: P) -> Result<Self, Error> {
```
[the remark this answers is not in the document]
lquerel: Outdated

crates/weaver_forge/src/config.rs
```diff
@@ -387,40 +307,464 @@
...
+
+        configs.reverse();
+        Self::from_paths(configs)
+    }
+
+    /// Builds the Weaver configuration from a collection of paths pointing to configuration files.
+    /// If a configuration file is not found, it is skipped.
+    /// The order of the paths is important as the first configuration in the slice is loaded, then
+    /// if present, the second configuration overrides the first one, and so on.
+    pub fn from_paths<P: AsRef<Path>>(paths: Vec<P>) -> Result<Self, Error> {
```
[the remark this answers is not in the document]
lquerel: Outdated

crates/weaver_forge/src/config.rs
```diff
@@ -387,40 +307,464 @@
...
+        let mut configs : Vec<FileContent> = Vec::new();
+
+        for path in paths {
+            if path.as_ref().exists() {
+                configs.push(FileContent::try_from_path(path.as_ref().clone())
```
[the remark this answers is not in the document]
lquerel: Outdated

crates/weaver_forge/src/config.rs
```diff
@@ -306,13 +211,30 @@ impl Default for TemplateSyntax {
...
     /// See <https://docs.rs/minijinja/latest/minijinja/struct.Environment.html#method.set_keep_trailing_newline>
-    pub keep_trailing_newline: bool,
+    pub keep_trailing_newline: Option<bool>,
+}
+
+impl WhitespaceControl {
+    /// Override the current `WhitespaceControl` with the `WhitespaceControl` passed as argument.
+    /// The merge is done in place. The `WhitespaceControl` passed as argument will be consumed and
+    /// used to override the current `WhitespaceControl`.
+    pub fn override_with(&mut self, other: WhitespaceControl) {
```
jsuereth, line 227: It's interesting you chose this manner of override. I consider this right to left override.

E.g.

```rust
let mut generic_config = load_generic();
let specific_config = load_specific();
generic_config.override_with(specific_config);
```

I haven't seen the directory walking yet, but I assume this is easier to understand / "fold" the config in this manner?

crates/weaver_forge/src/config.rs
```diff
@@ -387,40 +309,446 @@ impl CaseConvention {
...
-            serde_yaml::from_str(&weaver_file).map_err(|e| InvalidConfigFile {
-                config_file: WEAVER_YAML.into(),
-                error: e.to_string(),
-            })
-        } else {
-            Ok(TargetConfig::default())
+impl Default for WeaverConfig {
+    fn default() -> Self {
+        Self {
+            file_name: Default::default(),
```
jsuereth, line 315: Nit: Seeing this again, in the future, we can take file, function, arg, struct, field naming conventions config and put them in a (flattened) nested struct.

crates/weaver_forge/src/config.rs
```diff
@@ -387,40 +309,446 @@ impl CaseConvention {
...
+        }
+    }
+}
+
+impl WeaverConfig {
+    /// Collects all the configurations from the path passed in parameter. The configuration files
+    /// are collected in the order of the enumeration below:
+    /// - the $HOME/.weaver/weaver.yaml file,
+    /// - the <path>/weaver.yaml and its <parent folders>weaver.yaml.
+    pub fn collect_from_path<P: AsRef<Path>>(path: P) -> Vec<FileContent> {
```
jsuereth, line 343: Should we be exposing this "raw", or is this just pub for testing?

I think you should only be exposing a `pub fn load_config_from_path<P: AsRef<Path>>(path: P) -> Result<WeaverConfig, Error>`

The two functions as separate components seems like exposing implementation details.

crates/weaver_forge/src/file_loader.rs
```diff
@@ -21,7 +21,28 @@ pub trait FileLoader {
     fn all_files(&self) -> Vec<PathBuf>;
 
     /// Returns the content of a file from a given name.
-    fn load_file(&self, file: &str) -> Result<Option<String>, Error>;
+    fn load_file(&self, file: &str) -> Result<Option<FileContent>, Error>;
+}
+
+/// A struct that represents the content of a file.
+#[derive(Debug)]
+pub struct FileContent {
```
jsuereth, line 29: I'm not sure we want to preserve the full string contents of files we find before we parse the weaver config.  We should leave room for lazy loading, or Load + drop of the contents.

This is acceptable to fix post-this PR, but I don't think the struct as-is should be public and used across crates.

crates/weaver_forge/src/lib.rs
```diff
@@ -430,16 +440,36 @@ impl TemplateEngine {
...
+                ),
+                Cow::Owned(template_syntax.block_end.unwrap_or_else(|| "%}".to_owned())),
             )
             .variable_delimiters(
-                Cow::Owned(template_syntax.variable_start),
-                Cow::Owned(template_syntax.variable_end),
+                Cow::Owned(
+                    template_syntax
+                        .variable_start
+                        .unwrap_or_else(|| "{{".to_owned()),
```
jsuereth, line 454: Nit - We seem to have lost a simple-to-find place for defaults.  Can we at least make these constants in the header of this file?

jsuereth (commented):
Overall, I'm onboard with the mechanism for loading config, including explicit overrides and multiple layers.  I anticipate the need for that going forward, given our scope and mission.

LOTS of nits on crate structure and public functions.   Some are about leaving room for us to optimise in the future.  Let me know what you think we should address here and we can open feature request issues for the rest.

crates/weaver_forge/src/config.rs
lquerel: I feel like your example is identical to what is done in this PR. The `other` parameter in the `override_with` is actually a more specific configuration, just like in your example. Maybe there is something ambiguous somewhere in my comments that led you to think otherwise, or perhaps I misinterpreted your example.

crates/weaver_forge/src/config.rs
lquerel: It's an option indeed. More generally, I was wondering if this approach would be used by the authors of code generation, or if these authors would directly use conversion filters that are semantically agnostic to the target language. That is, using filters like `snake_case` or `screaming_name_case` directly, instead of a more semantic version of these filters such as `function_name` or `constant_name` (where the mapping is defined by these entries in the config). I initially thought it was a good idea; I'm just not sure if it will be used. What do you think?

crates/weaver_forge/src/config.rs
lquerel: I agree, this is a strange design. I'm working on improving the interface.

lquerel: Fixed in [add9388](https://github.com/open-telemetry/weaver/pull/231/commits/add9388de8f167e777ac9a07fc6010ec3d7850b0)

crates/weaver_forge/src/file_loader.rs
lquerel: Fixed in [582516b](https://github.com/open-telemetry/weaver/pull/231/commits/582516bf868d62e5c58c315eae33f283f00d46ee). I removed FileContent from the public API.

crates/weaver_forge/src/lib.rs
lquerel: Fixed in [78d4f40](https://github.com/open-telemetry/weaver/pull/231/commits/78d4f40d3ab653a85cc557ffa41876aff0866c73)

crates/weaver_forge/src/config.rs
jsuereth: Personal preference (and what I've seen so far) is to directly use the filters like `snake_case`.

I think we could back out this feature set and add it back in if we get requests for it.  Better to maintain less things with higher composability.

jsuereth (approved):

PR #246, opened 2024-07-17
Define SemConv predefined JQ functions

lquerel:
This PR adds a set of predefined JQ functions to simplify the processing and generation of semconv documentation/code.

**Process Registry Attributes**

The following JQ filter extracts the registry attributes from the resolved registry and returns a list of registry attributes grouped by namespace and sorted by attribute names.

```yaml
templates:
  - pattern: attributes.j2
    filter: semconv_grouped_attributes
    application_mode: each
```

The output of the JQ filter has the following structure:

```json5
[
  {
    "root_namespace": "user_agent",
    "attributes": [
      {
        "brief": "Value of the HTTP User-Agent",
        "examples": [ ... ],
        "name": "user_agent.original",
        "root_namespace": "user_agent",
        "requirement_level": "recommended",
        "stability": "stable",
        "type": "string",
        // ... other fields
      }, 
      // ... other attributes in the same root namespace
    ]
  },
  // ... other root namespaces
]
```

The `semconv_grouped_attributes` function also supports options to exclude specified root namespaces, specific stability levels, and deprecated entities. The following syntax is supported:

```yaml
templates:
  - pattern: attributes.j2
    filter: >
      semconv_grouped_attributes({
        "exclude_root_namespace": ["url", "network"], 
        "exclude_stability": ["experimental"],
        "exclude_deprecated": true,
      })
    application_mode: each
```

The structure of the output of `semconv_grouped_attributes` with these options is exactly the same as without the options. The JSON object passed as a parameter describes a series of options that can easily be extended if needed. Each of these options is optional. It's also possible to use the special parameter `$params` as follows:

```yaml
params:
  exclude_root_namespace: ["url", "network"] 
  exclude_stability: ["experimental"]
  exclude_deprecated: true

templates:
  - pattern: attributes.j2
    filter: semconv_grouped_attributes($params)
    application_mode: each
```

Technically, the `semconv_grouped_attributes` function is a combination of two semconv JQ functions:

```jq
def semconv_grouped_attributes($options):
    semconv_attributes($options)
    | semconv_group_attributes_by_root_namespace;

def semconv_grouped_attributes: semconv_grouped_attributes({});
```

The `semconv_attributes` function extracts the registry attributes and applies the given options. The `semconv_group_attributes_by_root_namespace` function groups the attributes by root namespace. It's possible to combine these two functions with your own JQ filters if needed.
 
**Process Metrics**

The following JQ filter extracts the metrics from the resolved registry, sorted by group root namespace and sorted by metric names.

```yaml
templates:
  - pattern: metrics.j2
    filter: semconv_grouped_metrics
    application_mode: each
```

The output of the JQ filter has the following structure:

```json5
[
  {
    "root_namespace": "jvm",
    "metrics": [
      {
        "attributes": [ ... ],
        "brief": "Recent CPU utilization for the process as reported by the JVM.",
        "id": "metric.jvm.cpu.recent_utilization",
        "instrument": "gauge",
        "metric_name": "jvm.cpu.recent_utilization",
        "root_namespace": "jvm",
        "note": "The value range is [0.0,1.0]. ...",
        "stability": "stable",
        "type": "metric",
        "unit": "1",
        // ... other fields
      },
      // ... other metrics in the same root namespace
    ]
  },
  // ... other root namespaces
]
```

The same options are supported by `semconv_grouped_metrics`, as shown in the following example:

```yaml
params:
  exclude_root_namespace: ["url", "network"] 
  exclude_stability: ["experimental"]
  exclude_deprecated: true

templates:
  - pattern: metrics.j2
    filter: semconv_grouped_metrics($params)
    application_mode: each
```

All the `semconv_grouped_<...>` functions are the composition of two functions: `semconv_<...>` and `semconv_group_<...>_by_root_namespace`.

**Other improvements**

This PR contains an update of the Weaver Forge README.md.
This PR updates the Rust example to leverage both the whitespace control settings and the new JQ filters.


crates/weaver_forge/src/lib.rs
```diff
@@ -258,9 +287,31 @@
...
         let mut errors = Vec::new();
 
+        let params = if let Some(mut params) = self.target_config.params.clone() {
+            match serde_yaml::to_value(&self.target_config.params) {
+                Ok(value) => {
+                    let prev = params.insert("params".to_owned(), value);
+                    if prev.is_some() {
+                        errors.push(Error::DuplicateParamKey {
+                            key: "params".to_owned(),
+                            error: "The parameter `params` is a reserved parameter name".to_string()
```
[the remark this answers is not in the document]
lquerel: Fixed

defaults/jq/semconv.jq
```diff
@@ -0,0 +1,102 @@
+def semconv_group_attributes_by_namespace:
+    group_by(.namespace)
+    | map({ namespace: .[0].namespace, attributes: . | sort_by(.name) });
+
+#####################
+# Attribute functions
+#####################
+def semconv_attributes($options):
```
jsuereth, line 8: Nit: Can you document the purpose of these functions in JQ, including options supported?

defaults/jq/semconv.jq
```diff
@@ -0,0 +1,102 @@
...
+      else
+        .
+      end
+    | map(. + {namespace: (if .name | index(".") then .name | split(".")[0] else "other" end)})
+    | if ($options | has("exclude_namespace")) then
+        map(select(.namespace as $st | $options.exclude_namespace | index($st) | not))
+      else
+        .
+      end
+    | sort_by(.namespace, .name);
```
jsuereth, line 28: My main complain here is around `.namespace` here and elsewhere.

We are doing this today in codegen and semconv.  However, should namespace be the FIRST prefix before `.` or ALL BUT LAST before `.`.   How do we want to treat namespaces overall?

In the hopes of making things better, I think we go with this. It should be easy to clean up and modify going forward (as this is the cleanest JQ I've seen), but I *do* think we need to make hard decisions in semconv tooling about namespacing and what it means.

jsuereth (approved):
Overall LGTM.  Just called out some of my concerns going forward on formalizing semconv decisions over here in the weaver codebase.

defaults/jq/semconv.jq
lmolkova: we currently care about the root namespace in the codegen, but I agree that we'll probably need more flexibility in the future. 

For now I'd only propose to make it explicit in the naming (if `.namespace` here is the root one, then let's call it `root_namespace` or similar)

lmolkova:
Could we postpone adding `semconv_grouped_resources`, `semconv_grouped_scopes`, `semconv_grouped_spans`, `semconv_grouped_events` until we have a use-case?

crates/weaver_codegen_test/templates/registry/alt_weaver.yaml
```diff
@@ -2,4 +2,4 @@
 params:
   attributes: true
   metrics: false    # With this alternate configuration, the metrics are disabled
-  registry_prefix: "registry."
+  exclude_deprecated: true
```
lmolkova, line 5: the default guideline is to generate deprecated attributes. I believe only golang doesn't generate them today.

So I'm not sure we need it. I have no objection having it, but if do, it should be `false` by default.

defaults/jq/semconv.jq
```diff
@@ -0,0 +1,102 @@
+def semconv_group_attributes_by_namespace:
+    group_by(.namespace)
+    | map({ namespace: .[0].namespace, attributes: . | sort_by(.name) });
+
+#####################
+# Attribute functions
+#####################
+def semconv_attributes($options):
+    .groups
+    | map(select(.type == "attribute_group" and (.id | startswith("registry."))))
```
lmolkova, line 10: not blocking this PR, but could help our future selves: 

we should have a more formal way to identify attribute definition than a group name prefix.

defaults/jq/semconv.jq
```diff
@@ -0,0 +1,102 @@
...
+        map(select(.stability as $st | $options.exclude_stability | index($st) | not))
+      else
+        .
+      end
+    | if ($options | has("exclude_deprecated") and $options.exclude_deprecated == true) then
+        map(select(.id | endswith(".deprecated") | not))
+      else
+        .
+      end
+    | map(. + {namespace: (if .id | index(".") then .id | split(".") | .[1] else "other" end)})
```
lmolkova, line 56: the group id is not a namespace, at least not today. I don't think we have a way to group span, resources, or scopes (also there are no scopes in the schema anymore). 

Events have a name, but I don't think we want anyone to codegen things for span events. And new log-based events are not there yet. 

So I suggest removing support for resource, scopes, spans, and events in another comment and we can revisit it once we start doing codegen for them.

lmolkova (approved):
Left some minor comments, looks great otherwise!

crates/weaver_codegen_test/templates/registry/alt_weaver.yaml
lquerel: Yes it's `false` by default.

defaults/jq/semconv.jq
lquerel: I agree, but that probably implies a change in the semconv schema, or are you thinking about something else?

defaults/jq/semconv.jq
lquerel: For now, I don’t have a problem with that. @jsuereth, are you okay with that as well?

defaults/jq/semconv.jq
lquerel: I will

defaults/jq/semconv.jq
lquerel: I can rename `namespace` to `root_namespace` as I believe it aligns well with @lmolkova's definition. @jsuereth, do you agree with that?

lquerel:
@lmolkova, @jsuereth, once I get confirmation for the few questions I have (see previous comment), I will make the changes, bump the weaver version, and create a new release. This release will be a good starting point for the codegen authors as these new JQ filters improve the UX. The step-by-step guide should also ease the process of the codegen migration.

defaults/jq/semconv.jq
jsuereth: Absolutely. I don't want to commit to supporting concepts we haven't evolved enough in semconv (that's related to my earlier comments).

Let's keep this focused and expand as we know more.

defaults/jq/semconv.jq
jsuereth: I think we're talking a schema change, and I'm supportive of this.  We should enrich our model to both *simplify* and enhance it to support the concepts we need.

defaults/jq/semconv.jq
jsuereth: I think `root_namespace` is a fair compromise between making weaver usable and not creating concepts that are shifting sands and will break over time.

+1 to `root_namespace`

defaults/jq/semconv.jq
lquerel: Done

defaults/jq/semconv.jq
lquerel: Done

defaults/jq/semconv.jq
lquerel: Done

defaults/jq/semconv.jq
lmolkova: Created https://github.com/open-telemetry/semantic-conventions/issues/1269 to track it, resolving the comment here

PR #257, opened 2024-07-23
Infrastructure to support backward-compatibility check

lquerel:
## Description

This PR implements the infrastructure to support backward-compatibility testing and, more generally, policies applied to multi-version registries.

The following command checks the current registry (main branch in the official repo) with a baseline registry (here, the archive v1.26.0) against a package of rules (policies) defined in the file `compatibility_check.rego`:

```sh
weaver registry check \
--registry https://github.com/open-telemetry/semantic-conventions.git[model] \
--baseline-registry https://github.com/open-telemetry/semantic-conventions/archive/refs/tags/v1.26.0.zip[model] \
--policy compatibility_check.rego
```

The following screenshot is the output of this command if we artificially remove the file `registry/deprecated/container.yaml` from the main branch.

Screenshot 2024-07-26 at 4 56 39 PM

## Changes

- [X] Extend the registry path format to support local and remote archives in addition to local folders and git repositories. The format will also support the definition of sub-folders. Motivation: with this PR, we now have two types of registries. Without a registry path format, we would need to add many parameters to support all combinations (folder, git-url, git-archive, folder, tag).
  - `<local-path>` (e.g. $HOME/otel-registry)
  - `<git-url>` (e.g. https://github.com/open-telemetry/semantic-conventions.git)
  - `<git-url>[<sub-folder>]` (e.g. https://github.com/open-telemetry/semantic-conventions.git[model])
  - `<git-archive-url>` (e.g. https://github.com/open-telemetry/semantic-conventions/archive/refs/tags/v1.26.0.zip)
  - `<git-archive-url>[<sub-folder>]` (e.g. https://github.com/open-telemetry/semantic-conventions/archive/refs/tags/v1.26.0.zip[model])
  - Future extension: `<git-url>@<tag>`
  - Future extension: `<git-url>@<tag>[<sub-folder>]`
- [X] Unify `RegistryPath` (merge the two existing implementations defined in the `weaver` and `weaver_semconv` crates).
- [x] Document CLI usage for the new registry path format.
- [x] Add `--baseline-registry` in addition to `--registry` (both will support the same syntax for the registry path). The `--registry-git-sub-dir` will be marked as deprecated.
- [x] Implement support for local and remove git archives (supported formats: .tar.gz, .zip).
- [x] Replace Weaver `Cache` with the more general concept of `RegistryRepo` supporting all the variants of `RegistryPath`.
- [X] Integrate `--baseline-registry` with the policy engine. In particular, we must decide how to expose the two registries in Rego.
- [X] Create standard SemConv Rego rules and function to ease the creation of semconv policies.
- [X] Create examples of policies to detect backward-compatibility issues.


crates/weaver_cache/src/lib.rs
```diff
@@ -152,65 +185,438 @@ impl Cache {
...
+            let path = entry.path().map_err(|e| InvalidRegistryArchive {
+                archive: archive_filename.to_owned(),
+                error: e.to_string(),
+            })?;
+
+            if let Some(valid_entry_path) = Self::path_to_unpack(&path, sub_folder, target_path) {
+                Self::create_parent_dirs(&valid_entry_path, archive_filename)?;
+                // Unpack returns an Unpacked type containing the file descriptor to the
+                // unpacked file. The file descriptor is ignored as we don't have any use for it.
+                _ = entry
```
jsuereth, line 315: Nit for rust - Maybe we should open a FR - that `#[must_use]` for `Result` where you `?` away a `()` doesn't cause a warning, and then you can `map_error(..)` and something like `.ignore_result()?`

crates/weaver_cache/src/lib.rs
```diff
@@ -152,65 +185,438 @@ impl Cache {
...
+        drop(repo);
+        // The temporary directory should be deleted automatically.
+        assert!(!repo_path.exists());
+    }
+
+    #[test]
+    fn test_semconv_registry_git_repo() {
+        let registry_path = RegistryPath::GitRepo {
+            // This git repo is expected to be available.
+            url: "https://github.com/open-telemetry/semantic-conventions.git".to_owned(),
```
jsuereth, line 586: One thing i'm thinking over time - we should move away from relying on remote APIs like this and instead stand up a local server that hosts the file, for stability purposes.

Not sure how hard that is now, but we can at least open a bug to track this.  

It's happened in the past that we moved/renamed repositories, don't want this test to fail then.

crates/weaver_cache/src/registry_path.rs
```diff
@@ -0,0 +1,273 @@
...
+use regex::Regex;
+
+use crate::Error;
+
+/// Regex to parse a registry path supporting the following formats:
+/// - source
+/// - source@tag
+/// - source\[sub_folder\]
+/// - source@tag\[sub_folder\]
+static REGISTRY_REGEX: Lazy<Regex> = Lazy::new(|| {
```
jsuereth, line 18: Just stabilized in latest rust... nice usage.

crates/weaver_cache/src/registry_path.rs
```diff
@@ -0,0 +1,273 @@
...
+        /// URL of the remote archive
+        url: String,
+        /// Sub-folder within the archive containing the semantic convention registry
+        sub_folder: Option<String>,
+    },
+    /// Git repository containing a semantic convention registry.
+    GitRepo {
+        /// URL of the Git repository
+        url: String,
+        /// Tag of the Git repository (NOT YET SUPPORTED)
```
jsuereth, line 50: This should be tag, branch or commit

crates/weaver_cache/src/registry_path.rs
```diff
@@ -0,0 +1,273 @@
...
+                } else {
+                    write!(f, "{}", url)
+                }
+            }
+            RegistryPath::GitRepo {
+                url,
+                tag,
+                sub_folder,
+            } => {
+                let mut registry_path = url.clone();
```
jsuereth, line 132: (extra) Nit: I find this more readable with a match, so you can see the format permutations and no `.clone()` call.

```rust
match (tag, sub_folder) {
  (Some(tag), Some(folder)) => write!(f, "{}@{}[{}]", url, tag, folder),
  (Some(tag), None) => write!(f, "{}@{}", url, tag),
  (None, Some(folder)) => write!(f, "{}[{}]", url, folder),
  (None, None) => write!(f, "{}", url),
}
```

src/util.rs
```diff
@@ -89,13 +67,18 @@ pub(crate) fn init_policy_engine(
         engine.enable_coverage();
     }
 
+    // Add the standard semconv policies
+    // Note: `add_policy` the package name, we ignore it here as we don't need it
+    _ = engine
+        .add_policy("defaults/rego/semconv.rego", SEMCONV_REGO)
```
jsuereth, line 73: Should this only be added for a particular phase of the engine?  Aren't these symbols meaningless in `before_resolution` or are they meant to be applied there too?

src/util.rs
```diff
@@ -119,9 +102,19 @@ pub(crate) fn check_policy_stage<T: Serialize>(
     policy_stage: PolicyStage,
     policy_file: &str,
     input: &T,
+    data: &[T],
```
jsuereth, line 105: Nit: If you want data to actually be generic, you need to use a different type.  Right now input *AND* the array of data must be the same type.

At a minimum, you should have:

```
fn check_policy_stage<T: Serialize, U:Serialize>(
    policy_engine: &mut Engine,
    policy_stage: PolicyStage,
    policy_file: &str,
    input: &T,
    data: &[U],
```

However, if you want data to *not* be monomorphic, you'd need to use an HList (likely k-list) or tuple-type structure to preserve the original types.

I don't think this is what you want at all.  For now, at least, I'd use a different generic for data. In the future, you may want to use a different type to encapsulate serialization more generically.

jsuereth (approved):
Overall LGTM.  A couple nits - Predominately I think next time the Git-Url / path reference code should be a separate PR form the policy engine changing code.

THink you can close a few open issues for this one :)

lquerel:
> Overall LGTM. A couple nits - Predominately I think next time the Git-Url / path reference code should be a separate PR form the policy engine changing code.
> 
> THink you can close a few open issues for this one :)

Yes, creating two PRs was part of the initial intentions, but it got lost in the heat of the moment.

crates/weaver_cache/src/lib.rs
lquerel: I'm not 100% sure, but I think a templated `#[must_use]` function returning a `Result<T, Error>` with `T=()` doesn't emit a warning.

I like the idea of the `.ignore_result()?` if we can specify a parameter `&str` representing the justification.

crates/weaver_cache/src/lib.rs
lquerel: Created https://github.com/open-telemetry/weaver/issues/267

crates/weaver_cache/src/registry_path.rs
lquerel: Renamed `refspec` and updated the comment.

crates/weaver_cache/src/registry_path.rs
lquerel: Integrated. I like it.

src/util.rs
lquerel: IMO, some of the rules/functions definitions could be used in any phases.

src/util.rs
lquerel: Nice catch. I made the changes.

PR #270, opened 2024-07-30
Follow build tools case conversion rules 

### LINKED ISSUE #260: What are the exact rules for Pascal Case conversion?
Ideally, what should Pascal Case do for the following cases?
- k8s => K8s or K8S?
- service2endpoint => Service2endpoint or Service2Endpoint?
- my2ndVersion => My2ndVersion or My2NdVersion?
- version3releaseNote => Version3releaseNote or Version3ReleaseNote?

The first alternatives in each example follow the same rule, i.e., a digit is not a separator, and thus the following letter will not be capitalized.

The second alternatives in each example follow a different rule, which is that digits are separators, and thus the following letter will be capitalized.

The problem is that it seems to me that in Java, C#, etc., the rules are quite fluctuating, but I could be wrong. For example, I think that Service2Endpoint will be preferred as a class name, whereas K8s using different rules will be preferred.

I don't yet know how to best address this. Perhaps one should use one rule or the other depending on whether a word is an acronym or not, i.e., the rules of the first alternative for acronyms and the rules of the second alternative for the rest. It's quite complex, in the end, just for a Pascal Case filter. Any feedback is welcome on this delicate subject. In the meantime, I will open an issue for tracking.

### LINKED ISSUE #265: `(screaming_)snake_case` should not introduce new `_`
`screaming_snake_case` jinja filter introduces `_` between letters and numbers.

E.g. `inet6` results in `INET_6`, `tls.client.hash.md5` in `TLS_CLIENT_HASH_MD_5`.

Build-tools had a different behavior which arguably is nicer and can be expressed as `{{attr_name | upper | replace('.', '_')}}`, i.e. produces `INET6` and `TLS_CLIENT_HASH_MD5` constants.

lquerel:
Weaver considers numbers as word splitters, whereas build-tools doesn't. This results in invalid case conversion for acronyms such as k8s, s3, etc.

Closes #260, closes #265

### DIFF
--- a/crates/weaver_forge/src/config.rs
+++ b/crates/weaver_forge/src/config.rs
@@ -225,6 +225,7 @@ impl CaseConvention {
         static LOWER_CASE: OnceLock<Converter> = OnceLock::new();
         static TITLE_CASE: OnceLock<Converter> = OnceLock::new();
         static KEBAB_CASE: OnceLock<Converter> = OnceLock::new();
+        static SNAKE_CASE: OnceLock<Converter> = OnceLock::new();
 
         let text = text.replace('.', "_");
         match self {
@@ -263,9 +264,9 @@ impl CaseConvention {
             CaseConvention::PascalCase => text.to_case(Case::Pascal),
             CaseConvention::CamelCase => text.to_case(Case::Camel),
             CaseConvention::SnakeCase => {
-                // Convert to title case but do not consider digits
+                // Convert to snake case but do not consider digits
                 // as boundaries.
-                let conv = TITLE_CASE.get_or_init(|| {
+                let conv = SNAKE_CASE.get_or_init(|| {
                     Converter::new()
                         .add_boundary(Underscore)
                         .remove_boundary(DigitLower)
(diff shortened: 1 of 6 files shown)


crates/weaver_forge/src/config.rs
```diff
@@ -260,9 +264,38 @@ impl CaseConvention {
...
             }
-            CaseConvention::PascalCase => text.to_case(Case::Pascal),
+            CaseConvention::PascalCase => {
+                // Convert to pascal case but do not consider digits
+                // as boundaries.
+                let conv = PASCAL_CASE.get_or_init(|| {
+                    Converter::new()
+                        .add_boundary(LowerUpper)
+                        .add_boundary(Acronym)
+                        .remove_boundary(DigitLower)
```
lmolkova, line 274: I absolutely trust you that it works, but it'd be nice to add a test with a digit in the text

lquerel:
@lmolkova there are multiple tests with digit at the end of the file `crates/weaver_forge/src/extensions/case.rs`

joaopgrassi (approved):
LGTM!

jsuereth (approved):

PR #274, opened 2024-07-31
chore(jq): Report an error when the compilation fail

### LINKED ISSUE #264: JQ errors are suppressed
JQ errors are ignored during registry generation / markdown update.

**To Reproduce**

1. Define a `weaver.yaml` like 

  ```yaml
    - pattern: readme.md.j2
      filter: 'error("foo")' # or do something that'd result in JQ error
     ...
  ```
  (e.g. change one in semconv  https://github.com/open-telemetry/semantic-conventions/blob/a44ff768718b6e82f2dea42189ba14712c795e6b/templates/registry/markdown/weaver.yaml)

2. Run generation 

**Expected behavior**

- registry generation fails. 
- JQ error is propagated in the output
- generation stops

**Actual behavior**

- original groups are passed to jinja as if there was no JQ filter
- generation continues and fails later in jinja templates (if JQ modified the structure) or results in incorrect data (if structure is the same, but filters didn't run),

lquerel:
The interface of the method `ParseCtx.compile` in JAQ is error prone and it's a bit too easy to forget to check the errors. This PR fix this issue.

Closes #264


crates/weaver_forge/src/filter.rs
```diff
@@ -43,10 +43,22 @@ impl Filter {
             filter: filter_expr.to_owned(),
             error: "No parsed expression".to_owned(),
         })?;
+        let filter = ctx.compile(parsed_expr);
+        if !ctx.errs.is_empty() {
```
lquerel, line 47: Strangely, `ParseCtx.compile(...)` doesn't return a `Result` and forces us to check the `errs` field, as was done in the past... I opened an issue in the `jaq` project to improve that in the future.

lmolkova (approved):

jsuereth (approved):

PR #276, opened 2024-07-31
(fix) Remove legacy way of writing templates for semconv

### LINKED ISSUE #178: [cleanup] Remove `write!` based markdown generation in `semconv_gen`.
Now that we have a stable release semconv is on, we can remove the `writeln!` statements of `semconv_gen`, since going forward all markdown will be generated via jinja template.

jsuereth:
Fixes #178 - Removes the `write!` path of generating semconv snippets now that semantic conventions has moved onto the tempalte-engine based solution.


crates/weaver_semconv_gen/src/lib.rs
```diff
@@ -440,55 +390,4 @@ mod tests {
         ));
         Ok(())
     }
-
-    #[test]
-    fn test_http_semconv() -> Result<(), Error> {
```
lquerel, line 445: Do we still have unit tests for this crate?

lquerel (approved):
Overall LGTM. Just curious to know if we have tests covering these changes.

crates/weaver_semconv_gen/src/lib.rs
```diff
@@ -293,63 +269,43 @@ impl SnippetGenerator {
...
-        &self,
-        args: GenerateMarkdownArgs,
-        attribute_registry_base_url: Option<&str>,
-    ) -> Result<String, Error> {
-        let mut ctx = GenerateMarkdownContext::default();
-        let mut result = String::new();
-        if args.is_metric_table() {
-            let view = MetricView::try_new(args.id.as_str(), &self.lookup)?;
-            view.generate_markdown(&mut result, &mut ctx)?;
+        // TODO - define context.
```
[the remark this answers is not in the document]
jsuereth: I think we need to sort through what we expose in context.  Right now it's bare minimum for compatibility, not a long term "what should this be for good codegen"

crates/weaver_semconv_gen/src/lib.rs
jsuereth: Parser has a ton of unit tests, and there's a minimal template-engine test which was the only template engine test we were using.

I don't think it's worth re-testing whether we resolve semconv or how JSON is formed, so it just makes sure that a markdown file is placed in markdown snippets.

Before we had to test all the details of the printlns.

jsuereth:
So TL;DR; on tests - I just removed tests that covered deleted code.  We only had one test for the template engine, all other tests (and complex logic) is in the parser file.  So we have no real change in test coverage from the thing being used today.

PR #277, opened 2024-07-31
[PoC]of attribute ref renaming

MadVikingGod:
This is a proof of concept on the viability of using renaming of attributes to duplicate the functionality of attribute embedding in https://github.com/open-telemetry/semantic-conventions/issues/1264

This adds another field to the ref attributes that allows you to rename the attribute.  This lets you import multiple copies of the same reference into an attribute group. as demonstrated in the test.

This is a Proof of Concept please don't merge.


crates/weaver_resolver/src/attribute.rs
```diff
@@ -82,18 +82,22 @@ impl AttributeCatalog {
...
                 deprecated,
                 prefix,
+                rename,
             } => {
                 let name;
                 let root_attr = self.root_attributes.get(r#ref);
                 if let Some(root_attr) = root_attr {
                     let mut attr_lineage = AttributeLineage::new(&root_attr.group_id);
 
-                    if *prefix {
```
trisch-me, line 91: fyi: prefix means that attribute will be embedded i.e. having parent prefix in his name

jsuereth:
@MadVikingGod or @trisch-me -  Are we entertaining this PoC anymore or should we close this PR as stale at this point?

trisch-me:
As I mentioned during one of the meetings we already have working `old` way of embedding the attributes in weaver including renaming.

As we are moving slowly forward with embed I propose to implement everything there.

Issue #269, opened 2024-07-29
Consider adding jinja filter that returns enum member value 


lmolkova:
The following code would be common across languages

```jinja
{%- macro print_value(type, value) -%}
{%- if type == "string" -%}"{{value}}"{%-else-%}{{value}}{%-endif-%}
{%- endmacro -%}
```

If enum attribute has int values, it returns value without quotes, otherwise it returns value in quotes, e.g.
```python
class RpcGrpcStatusCodeValues(Enum):
    OK = 0
```

and 

```python
class RpcConnectRpcErrorCodeValues(Enum):
    CANCELLED = "cancelled"
```

It'd be nice to have a convenience method in weaver for it.

lquerel:
@lmolkova, before implementing it, I’d like to clarify a few questions:
- What does `print_value` return for a value that is not a string, number, or boolean? An empty string?
- String values are double-quoted, but do we automatically escape special characters like `"`, `\`, new line, etc., into their common escaping representation `\"`, `\\`, `\n`, etc.? This is important as this filter will be used in the context of code generation. I think most of the languages I’m aware of use this type of escaping, but we may need to support an optional parameter to specify the encoding for some exotic languages. Could be implemented in a second phase if needed.
- Is `print_value` the best name for this filter? Other ideas `fmt_val`, `smart_quote`, `codegen_quote`, `codegen_val`, `smart_fmt`, ... 

Adding @jsuereth for more feedback.

lmolkova:
Printing value is necessary for enum attributes. They can only have string, number, or boolean types - they are compile-time constants. 
Escaping is a great point - we probably need it since value can contain any unicode chars.

I'd prefer to have a very limited scope for this method (and very high convenience - it's a pure sugar). E.g. `member | print_member_value`.

If you'd rather have a generic format-anything, I have no objections, but also I think it's not critical and can wait for more feedback from people who work on the codegen before we add it.

lquerel:
Implemented in #278

PR #278, opened 2024-07-31
`print_member_value` filter

lquerel:
This PR implements #269


crates/weaver_forge/src/extensions/otel.rs
```diff
@@ -324,6 +325,19 @@ pub(crate) fn instantiated_type(attr_type: &Value) -> Result<String, minijinja::
     )))
 }
 
+/// Converts an enum member value into:
+/// - A quoted string if the input is a string.
+/// - A non-quoted string if the input is a number or a boolean.
+/// - An empty string otherwise.
+pub(crate) fn print_member_value(input: &Value) -> String {
+    match input.kind() {
+        ValueKind::String => format!("\"{}\"", input),
```
jsuereth, line 334: Do we need to "escape" quotes in the string when doing this?  Likely...

jsuereth (approved):
One, somewhat major, nit/fix we'll need to add.

crates/weaver_forge/src/extensions/otel.rs
lquerel: Oh yes. It was planned and I forgot...

lquerel: Fixed

PR #284, opened 2024-08-01
Let policy flag accept directories.

### LINKED ISSUE #259: Facilitate the import of multiple rego files in `weaver registry check`
Currently, the `--policy` parameter allows specifying a single rego file. It would be useful to be able to import all rego files present in a directory. I see two possible approaches:
* Extend the existing `--policy` parameter to accept a directory and modify the internal logic of weaver to import, in this case, all files with the `.rego` extension.
* Add a new `--policies` parameter to specify a directory and automatically import all `**/*.rego` files.

I don’t have a strong preference for either approach. Feel free to share your opinion in the comments of this issue.

MadVikingGod:
fixes #259 

This change allows --policy flags to accept a directory and include any *.rego file found.  It does not recurse into sub directories.

### DIFF
--- a/crates/weaver_checker/data/multi-policies/otel_policies.rego
+++ b/crates/weaver_checker/data/multi-policies/otel_policies.rego
@@ -1,0 +1,79 @@
+package before_resolution
+
+# Conventions for OTel:
+# - `data` holds the current released semconv, which is known to be valid.
+# - `input` holds the new candidate semconv version, whose validity is unknown.
+#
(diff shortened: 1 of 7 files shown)


lquerel (approved):
Overall, LGTM. Thank you for this contribution. I made a few suggestions. Before merging, could you also update the comment on lines 35-36 in src/registry/check.rs with the following text: ‘List policy files or directories containing policy files to verify against the semantic convention registry.’ There are other sub-commands like generate, resolve, etc., where you should also make this change. Thanks.

Issue #288, opened 2024-08-02
Add support for a `params` section per template


lquerel:
It’s not uncommon to have the same template applied with a different JQ filter and a different name for the generated file. Having the ability to specify a set of parameters per template specification in the weaver.yaml configuration file would be helpful. Today, it’s possible to simulate that by adding a projection in the JQ filter to include this information, but it’s not obvious and not necessarily easy to understand.

```yaml
params:
  excluded_attributes: ["messaging.client_id"]

templates:
  - pattern: "attributes.ts"
    filter: semconv_attributes({"exclude_stability": ["experimental"]})
    params:
      file_name: "stable_attributes.ts"
    application_mode: single
  - pattern: "attributes.ts"
    filter: semconv_attributes({"exclude_stability": ["stable"]})
    params:
      file_name: "experimental_attributes.ts"
    application_mode: single
```

lmolkova:
It'd be nice to support `file_name` with a jinja template like `file_name: "{{ctx.output}}/{{ctx.root_namespace}}_attributes.ts"` or `file_name: "{{ctx.stability}}_attributes.ts"`.

Ideal stolen from @jsuereth approach to file name in build-tools 

https://github.com/open-telemetry/build-tools/blob/751485681aadcf9f47b51a9b5a9f6d91a9fc7ec8/semantic-conventions/README.md?plain=1#L341

lquerel:
Implemented in #294

PR #294, opened 2024-08-06
Add template-level parameters and `file_name` per template config

lquerel:
This PR adds two new optional fields in the `templates` section of a Weaver configuration file:
- `params`: Defines a map of key-value pairs which will be merged with the ones defined at the top level of the `weaver.yaml` file. CLI parameters can also be used to override these parameters.
- `file_name`: Defines the relative file path of the output generated from the template. This field is a Jinja expression which will be evaluated with the same context used in the template itself. The parameters mentioned above are also accessible.

An example of usage:
```yaml
params:
  exclude_deprecated: true
  
templates:
  - pattern: attribute.py
    params:
      exclude_stability: ["stable"]
    filter: "semconv_grouped_attributes($params)"
    application_mode: each
    file_name: "incubating/{{ctx.root_namespace | snake_case}}.py"
  - pattern: attribute.py
    params:
      exclude_stability: ["experimental"]
    filter: "semconv_grouped_attributes($params)"
    application_mode: each
    file_name: "{{ctx.root_namespace | snake_case}}.py"
```

Note: `params` are now merged in an additive way. For example, if a parameter is defined in the `params` section of a template, then the final `params` section for this template will include both the `params` inherited from the file-level configuration and the `params` defined in the template configuration. If the override is a parameter set to null, then this parameter is removed from the final `params` section. This modification should not have a significant impact on the existing codegen, as I don't believe any codegen was relying on this specific behavior.

Closes [open-telemetry/weaver#288](https://github.com/open-telemetry/weaver/issues/288).

### DIFF
--- a/crates/weaver_forge/src/lib.rs
+++ b/crates/weaver_forge/src/lib.rs
@@ -207,9 +207,9 @@ impl TemplateEngine {
 
                 // Add CLI-level params to the template params. If a param is already defined
                 // in the template params, the CLI-level param will override it.
-                // Note: The result of the insert method is ignored because we don't care about
-                // the previous value of the param.
                 for (name, value) in params.params.iter() {
+                    // The result of the insert method is ignored because we don't care about
+                    // the previous value of the param.
                     _ = template_params.insert(name.clone(), value.clone());
                 }
 
--- a/crates/weaver_forge/src/lib.rs
+++ b/crates/weaver_forge/src/lib.rs
@@ -910,7 +910,7 @@ mod tests {
         // for test coverage purposes.
         let mut templates = engine.target_config.templates.unwrap_or_default();
         templates.push(TemplateConfig {
-            pattern: Glob::new("converter.md").unwrap(),
+            template: Glob::new("converter.md").unwrap(),
             filter: ".".to_owned(),
             application_mode: ApplicationMode::Single,
             params: None,
--- a/crates/weaver_forge/src/lib.rs
+++ b/crates/weaver_forge/src/lib.rs
@@ -701,6 +701,7 @@ mod tests {
 
     fn prepare_test(
         target: &str,
+        cli_params: Params,
     ) -> (
         TestLogger,
         TemplateEngine,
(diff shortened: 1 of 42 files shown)


crates/weaver_semconv_gen/Cargo.toml
```diff
@@ -18,7 +18,6 @@ weaver_resolved_schema = { path = "../weaver_resolved_schema" }
 weaver_semconv = { path = "../weaver_semconv" }
 nom = "7.1.3"
 
-itertools.workspace = true
```
jsuereth, line 21: Think I missed this when removing the `writeln!` code, good catch.

crates/weaver_forge/src/lib.rs
```diff
@@ -192,13 +198,33 @@ impl TemplateEngine {
...
+        if let Some(templates) = config.templates.as_mut() {
+            for template in templates {
+                let template_params = template.params.get_or_insert_with(BTreeMap::new);
+
+                // Add CLI-level params to the template params. If a param is already defined
+                // in the template params, the CLI-level param will override it.
+                // Note: The result of the insert method is ignored because we don't care about
+                // the previous value of the param.
+                for (name, value) in params.params.iter() {
+                    _ = template_params.insert(name.clone(), value.clone());
```
jsuereth, line 213: Nit: can you add the comment about ignoring the previous value here?

jsuereth (commented):
Overall looks good BUT I find the weaver config confusing.

I think we should deprecated `pattern` for `template`.

Specifically:

```
templates:
  - pattern: test.md
    filter: .
    application_mode: single
    params:
      template_name: test1
      test_1: true
    file_name: "{{params.template_name | snake_case}}.json"
```

Looks confusing, since `test.md` is the *template* you're using, not a pattern at all.

Instead, let's do:

```
templates:
  - template: test.md
    filter: .
    application_mode: single
    params:
      template_name: test1
      test_1: true
    file_name: "{{params.template_name | snake_case}}.json"
```

I think we can do this in a non-breaking way by deprecating the "pattern" attribute and adding "template" where we pull from pattern if it exists, otherwise use template.

lquerel:
> I think we can do this in a non-breaking way by deprecating the "pattern" attribute and adding "template" where we pull from pattern if it exists, otherwise use template.

Can be done with a serde alias. I will also update the docs.

lquerel:
@jsuereth I renamed `templates.pattern` to `templates.template` and added `pattern` as a serde alias of `template` for backward-compatibility. I also updated documentations and examples.

crates/weaver_forge/src/lib.rs
lquerel: Done

Issue #263, opened 2024-07-26
Change what prefix means in the schema


lmolkova:
Depends on https://github.com/open-telemetry/semantic-conventions/issues/1292 and https://github.com/open-telemetry/semantic-conventions/pull/1293

Once we migrate semaconv tooling to  weaver, we can reconsider usage prefix:

- prefix could be a property that's used to validate that all attributes defined in scope of group have this prefix
  - it could be required to embed things and might be optional in other cases
- prefix is removed and only used when embedding things.

lmolkova:
blocked on embedding https://github.com/open-telemetry/semantic-conventions/issues/1264

jsuereth:
Let's remove prefix form weaver in `--future` mode.

PR #297, opened 2024-08-07
feat: Add support for event body fields

MSNev:
Adds support for the event body fields reusing the existing `event` type.
- Supports existing `event` (Span) definition
- Supports the log based `event` definition as an extension of the existing span definition
- Adds `Body` (optional) definition which supports a collection of `fields` or a single `string` value
  - Body has a `type` of string or map
  - An optional brief (required for string type)
  - An optional note
  - An optional stability
  - An optional examples (required for string type)


MSNev:
The PR supersedes #236 as discussed in tools sigs this morning.

crates/weaver_resolver/data/registry-test-4-events/registry/browser-event.yaml
```diff
@@ -0,0 +1,54 @@
+groups:
```
MSNev, line 1: This is testing that the resolved schema is handling ALL of the existing Semconv defined events as well as `body` and `fields` definition.

Hence, all of the files added to the `registry` folder

defaults/jq/semconv.jq
```diff
@@ -104,3 +108,51 @@ def semconv_grouped_metrics($options): semconv_metrics($options) | semconv_group
...
+#################
+# Event Functions
+#################
+
+# Extracts and processes semantic convention events based on provided options.
+# $options is an object that can contain:
+# - exclude_stability: a list of stability statuses to exclude.
+# - exclude_deprecated: a boolean to exclude deprecated events.
+# - exclude_event_namespace: a list of event domain to exclude.
+def semconv_events($options):
```
jsuereth, line 121: I'd suggest we hold off on providing this function in weaver for now.  We can directly encode this in semconv repo for markdown generation while we work on the data model to clean up issues around events and make hard-than-this-PR decisions around, e.g. namespacing and group-level deprecation.

crates/weaver_semconv/src/group.rs
```diff
@@ -114,11 +118,30 @@ impl GroupSpec {
...
+        if self.r#type == GroupType::Event {
+            if self.body.is_some() && self.name.is_none() {
+                // Must have a name which is assigned to event.name for log based events
+                errors.push(Error::InvalidGroup {
+                    path_or_url: path_or_url.to_owned(),
+                    group_id: self.id.clone(),
+                    error: "This group contains an event type with a body definition but the name is not set.".to_owned(),
+                });
+            }
+            if self.name.is_none() && self.prefix.is_empty() {
```
jsuereth, line 130: cc @lmolkova - For concerns on prefix.

MSNev: FYI - This is really here to "handle" all of the existing semconv events, as there are several existing `events` that don't have a name and instead use the `prefix` as the name.

defaults/jq/semconv.jq
```diff
@@ -104,3 +108,51 @@ def semconv_grouped_metrics($options): semconv_metrics($options) | semconv_group
...
+    .groups
+    | map(select(.type == "event"))
+    | if ($options | has("exclude_stability")) then
+        map(select(.stability as $st | $options.exclude_stability | index($st) | not))
+      else
+        .
+      end
+    | if ($options | has("exclude_deprecated") and $options.exclude_deprecated == true) then
+        map(select(.id | endswith(".deprecated") | not))
```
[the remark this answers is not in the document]
MSNev: Thats exactly what this is doing?? -- marking the `event` (the whole group) as deprecated?

defaults/jq/semconv.jq
MSNev: Let me move this to the required "test" `weaver.yaml` definitions then. Will be tomorrow (more likely Friday) before I get to this, so any additional comments by others (to roll into a single update) would be appreciated.

crates/weaver_semconv/src/group.rs
lmolkova: they are gone now and nothing is allowed to add a prefix as of last version https://github.com/open-telemetry/semantic-conventions/pull/1293 :)

But it's a limitation on the semconv side and tooling is not aware of it (yet). We might remove prefix from the schema as well - https://github.com/open-telemetry/weaver/issues/263.

There is also a chance that weaver is used with older versions of semconv.

TL:DR: we should probably to support prefix for the time being, but it's unlikely it'll ever be used on events.

defaults/jq/semconv.jq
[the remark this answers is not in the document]
lmolkova: it should not use group id - there is a `deprecated`  property on the event group which should be used instead.


I.e.
```yaml
  - id: az.resource.log
    type: event
    deprecated: "foobar"
    name: az.resource.log
    ...
```

and then the event-specific JQ filters should use the explicit property instead of unreliable group id.

crates/weaver_semconv/src/group.rs
```diff
@@ -86,6 +87,9 @@ pub struct GroupSpec {
     pub name: Option<String>,
     /// The readable name for attribute groups used when generating registry tables.
     pub display_name: Option<String>,
+    /// The event body definition
+    /// Note: only valid if type is event
+    pub body: Option<BodySpec>,
```
lmolkova, line 92: Do you have any thoughts on how to prevent somebody from adding body on the existing 'span events' definitions beyond code review?

Also, assuming we have smart codegen for events, what should it do? Generate span event? log event? Doesn't matter? Not possible until we deprecate span-events?

defaults/jq/semconv.jq
```diff
@@ -104,3 +108,51 @@ def semconv_grouped_metrics($options): semconv_metrics($options) | semconv_group
...
+      else
+        .
+      end
+    | map(. + {
+        event_namespace: (if .id | index(".") then (.id | split(".") | .[0:-1] | join(".")) else "other" end)
```
lmolkova, line 135: we should never use group id in JQ - please use full event name.

crates/weaver_resolver/src/attribute.rs
```diff
@@ -23,12 +24,55 @@ pub struct AttributeCatalog {
...
+/// The Attribute with its group ID.
+pub struct AttributeWithGroupId {
+    /// The attribute.
     pub attribute: attribute::Attribute,
+    /// The group ID.
     pub group_id: String,
 }
 
+/// A reference to an attribute with its original attribute and group ID.
+pub struct AttributeRefWithGroup {
```
lquerel, line 36: Do we really need this if a body field is, in fact, an attribute once resolved? My feeling is that we are adding too much complexity here. Probably the best course of action for now is to finalize the discussion on the aliased attributes and then see if we really need all this code in the AttributeCatalog.

crates/weaver_semconv/src/group.rs
lquerel: Looks like a good candidate for the policy engine.

lquerel (changes requested):
This PR represents a lot of work and is overall really good. However, my main concern is the lack of alignment between the concept of a body field and the concept of embedded/aliased attributes. Depending on how this discussion ends, the implications for this PR could be more or less significant.

crates/weaver_resolver/data/registry-test-4-events/registry/browser-event.yaml
```diff
@@ -0,0 +1,54 @@
...
+          SHOULD be left unset.
+      - id: language
+        type: string
+        brief: 'Preferred language of the user using the browser'
+        note: >
+          This value is intended to be taken from the Navigator API
+          `navigator.language`.
+        examples: ["en", "en-US", "fr", "fr-FR"]
+    body:
+      fields:
```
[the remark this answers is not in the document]
MSNev: At some point yes. https://github.com/open-telemetry/semantic-conventions/issues/755

crates/weaver_resolver/data/registry-test-4-events/registry/browser-ref-event.yaml
```diff
@@ -0,0 +1,52 @@
...
+    name: browser.pageview
+    type: event
+    brief: >
+      This event represents an occurrence of a page view.
+    attributes:
+      - ref: session.id
+      - ref: browser.platform
+    body:
+      fields:
+        - ref: http.url
```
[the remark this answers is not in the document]
MSNev: These are NOT attributes in the same sense as the registry attributes. The recommendation is that field names should be as short as possible, they are also unique to the specifically named event (ie. they are not shared with ANYTHING else).

crates/weaver_resolver/data/registry-test-4-events/registry/browser-event.yaml
```diff
@@ -0,0 +1,54 @@
+groups:
+  - id: registry.browser
+    name: browser.test.event
+    prefix: browser
```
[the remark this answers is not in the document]
MSNev: These were copied either directly from SemConv or previously open PR's, perhaps I've missed something but the `prefix` is (or was) a standard mechanism for defining a common "group" name which is applied to the defined `attributes`.

For events the `prefix` is EXPLICITLY NOT used for the body fields (only any (optionally) defined attributes.

crates/weaver_semconv/src/group.rs
MSNev: The intent is that an `event` is an `event`, we would not want to "stop" someone adding `body field` to any `event` definition.

The `body` is optional, so as long as the new definition is defined as optional then (today) the same `event` definition can be used for both log and span events, on the proviso that it's currently not possible to pass the body fields as a `span` event.

One possible way of forward dealing with this would be something like https://github.com/open-telemetry/opentelemetry-specification/issues/3406#issuecomment-1523629641 where the `log.body` attribute would be defined as a JSON encoded string.

crates/weaver_forge/src/registry.rs
```diff
@@ -134,6 +138,12 @@ impl ResolvedGroup {
             })
             .collect();
         let lineage = group.lineage.clone();
+        let body = if group_type == GroupType::Event {
```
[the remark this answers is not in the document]
MSNev: This is really just explicitly identifying that `body` is only for events (belts and braces) to ensure that (if) somehow, someone in the future sees this (and/or bypasses calling the validate function -- eg by just directly calling this function) it's explicitly stopping the population of the body, which if expected (by them) this should give them some additional pause to figure out why.

crates/weaver_forge/src/registry.rs
```diff
@@ -196,6 +207,12 @@ impl ResolvedRegistry {
                     })
                     .collect();
                 let lineage = group.lineage.clone();
+                let body = if group_type == GroupType::Event {
```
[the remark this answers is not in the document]
MSNev: Same as : https://github.com/open-telemetry/weaver/pull/297#discussion_r1709933735

crates/weaver_resolved_schema/src/body.rs
```diff
@@ -0,0 +1,90 @@
...
+    // Not yet defined in the spec or implemented in the resolver
+    // The body value when there are no fields
+    // #[serde(skip_serializing_if = "Option::is_none")]
+    // pub value: Option<Value>
+}
+
+/// An attribute definition.
+#[derive(Serialize, Deserialize, Debug, Clone, Eq, PartialEq, Hash, JsonSchema)]
+#[serde(deny_unknown_fields)]
+pub struct BodyField {
```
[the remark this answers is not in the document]
MSNev: Yes, as a `BodyField` is NOT an `Attribute`, it is similar in that there are lots of "common" properties, but a field is missing several of them which will NEVER be used or required. So I defined as an explicit type to ensure that this never occurs or someone does go hey `this` has `xxx` how can I use that (when it will never and have or need it)

crates/weaver_resolver/src/attribute.rs
MSNev: Body fields will NEVER be present in the Attribute Catalog as they are NOT attributes.

Different event definitions MAY define the same field "name" but they MAY have semantically different meanings (ie. a Body Field named XXX is explicitly NOT directly comparable with any other Body Field named XXXX UNLESS the event.name is the same.

See https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/events.md#general-event-semantics and explicitly the section `Recommendations on using attributes vs. body fields`

crates/weaver_resolved_schema/src/body.rs
```diff
@@ -0,0 +1,90 @@
...
+    /// Attribute name.
+    pub name: String,
+    /// A reference to an attribute definition, used to populate the relevant
+    /// fields of the body field, unless they are overridden by the body field.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub r#attr: Option<AttributeRef>,
```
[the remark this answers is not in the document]
MSNev: Good point, I originally added as a just in case. But you are correct this is not required and should not be exposed on the resolved schema -- will remove.

crates/weaver_resolved_schema/src/body.rs
```diff
@@ -0,0 +1,90 @@
...
+    /// A alias to use for the field for the referenced attribute.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub alias: Option<String>,
```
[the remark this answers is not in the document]
MSNev: Another good point, will remove as the resolved schema doesn't need to know "how" the name of the field was derived -- will remove.

crates/weaver_resolved_schema/src/body.rs
```diff
@@ -0,0 +1,90 @@
...
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+use weaver_semconv::attribute::{AttributeType, Examples, RequirementLevel};
+use weaver_semconv::body::{BodyFieldSpec, BodySpec};
+use weaver_semconv::stability::Stability;
+
+/// An attribute definition.
+#[derive(Serialize, Deserialize, Debug, Clone, Eq, PartialEq, Hash, JsonSchema)]
+#[serde(deny_unknown_fields)]
+pub struct Body {
```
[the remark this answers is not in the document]
MSNev: Yeah, lots of copy/paste issues here, will address.

crates/weaver_resolver/data/registry-test-4-events/registry/browser-ref-event.yaml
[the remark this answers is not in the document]
MSNev: I'll bring this up in the Event Sig today, but this both predates any of the embed discussions and is a different scenario as a body field is not an attribute.

This referencing is mostly to avoid repeating and re-defining the same thing over and over again, for example things like `http.url` may be present in an event body as multiple fields but would need to have different names (alias) within the body. `url`, `referrer`, etc (the names will depend on the event definition).

defaults/jq/semconv.jq
MSNev: This was copied from the existing (or previous) jq query on this, as far as I understand the `id` here is the resolved `id` (derived from the name and prefix, not the `id` from the registry entry.
But I'm going to be moving this to be isolated to weaver for testing anyway.

crates/weaver_resolver/data/registry-test-4-events/registry/browser-ref-event.yaml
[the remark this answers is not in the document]
MSNev: Outcome from Events SIG: While we have a strong `desire` to have the ability to "ref" an existing attribute to avoid duplication, we don't have a strong `need` at this point, so we all agreed that I'm ok to remove this concept from weaver for this initial version.

We strongly "suspect" that once event yaml starts to get defined there will be "questions" raised about why all the duplication -- which is what we (and me) where trying to avoid.

crates/weaver_resolver/data/registry-test-4-events/registry/browser-ref-event.yaml
```diff
@@ -0,0 +1,52 @@
...
+          brief: >
+            This attribute represents the URL of the page that was viewed.
+          examples: ['https://en.wikipedia.org/wiki/Main_Page', 'https://en.wikipedia.org/wiki/Main_Page#foo']
+        - id: type
+          brief: >
+            This attribute represents the state the application has transitioned into at the occurrence of the event.
+          note: >
+            This identifies the type of page view. For example, if the page view represents a user 
+          type:
+            allow_custom_values: false
```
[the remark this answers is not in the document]
MSNev: This test file is being deleted from the next iteration due to the removal of `ref` support

defaults/jq/semconv.jq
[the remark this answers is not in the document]
MSNev: All changes in this file have been reverted

defaults/jq/semconv.jq
MSNev: All changes in this file have been reverted

crates/weaver_resolver/data/registry-test-4-events/registry/browser-event.yaml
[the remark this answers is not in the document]
MSNev: Was about to delete this and noticed that this is being explicitly used and tested for the defined attributes in this test event definition.

crates/weaver_forge/src/registry.rs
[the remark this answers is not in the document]
MSNev: Removed

crates/weaver_forge/src/registry.rs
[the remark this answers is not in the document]
MSNev: Removed

crates/weaver_resolved_schema/src/body.rs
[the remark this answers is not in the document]
MSNev: Removed as the `ref` / `alias` option is now also removed

crates/weaver_resolved_schema/src/body.rs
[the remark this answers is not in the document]
MSNev: removing as the `ref` / `alias` option is not also removed

crates/weaver_resolver/src/attribute.rs
MSNev: Removing as the no longer needed with the removal of `ref` / `alias` option

crates/weaver_semconv/src/group.rs
MSNev: > we should probably support prefix for the time being, but it's unlikely it'll ever be used on new events.

Agree, as if I remove this then this would also force that the released version of weaver (with this change) will only work with the versions of semconv that has them removed.

defaults/jq/semconv.jq
MSNev: All changes to the semconv.jq have been reverted.

crates/weaver_semconv/src/body.rs
```diff
@@ -0,0 +1,243 @@
...
+            id: "id".to_owned(),
+            r#type: AttributeType::PrimitiveOrArray(PrimitiveOrArrayTypeSpec::Int),
+            brief: Some("brief".to_owned()),
+            examples: Some(Examples::Int(42)),
+            requirement_level: RequirementLevel::Basic(BasicRequirementLevelSpec::Required),
+            note: "note".to_owned(),
+            stability: Some(Stability::Stable),
+            deprecated: Some("deprecated".to_owned()),
+        };
+        assert_eq!(field.id, "id");
```
MSNev, line 229: Except for the `is_required` these other asserts feel a little redundant now that the body field is a simple struct...

MSNev:
The tests are failing with this unrelated error
semconv::tests::test_semconv_spec_from_url' panicked at crates/weaver_semconv/src/semconv.rs:305:9:

Investigating locally, instead of being the RegistryNotFound its getting
Error: InvalidSemConvSpec { path_or_url: "http://unknown.com/unknown-semconv.yaml", line: Some(1), column: Some(1), error: "invalid type: string \"<!DOCTYPE html><html><head><script>window.onload=function(){window.location.href=\\\"/lander\\\"}</script></head></html>\", expected struct SemConvSpec" }
thread 'semconv::tests::test_semconv_spec_from_url' panicked at crates\weaver_semconv\src\semconv.rs:308:9:

crates/weaver_resolved_schema/src/body.rs
lmolkova, line 1: Please also update https://github.com/open-telemetry/weaver/blob/main/schemas/semconv.schema.json and syntax.md file next to it

MSNev: Updated

crates/weaver_resolved_schema/src/body.rs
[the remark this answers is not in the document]
jsuereth: IIUC though - you DON'T need an unresolved body type yet because body cannot be unresolved right now.  

Also - are we concerned about sharing the same types (`AttributeType`) between these?  Could we evolve without this being a breaking change?

crates/weaver_resolver/src/body.rs
```diff
@@ -0,0 +1,50 @@
...
+            // Add as a placeholder for now of where to resolve the value.
+            Err(Error::NotImplemented {
+                message: "Value type for body is not currently implemented.".to_owned(),
+            })
+        }
+    }
+}
+
+/// Resolve a `BodyField`` specification into a resolved `BodyField``.
+pub fn resolve_body_field_spec(field: &BodyFieldSpec) -> Result<BodyField, Error> {
```
jsuereth, line 39: This doesn't appear to be doing anything useful.  Can we nix this?

crates/weaver_resolver/src/body.rs
```diff
@@ -0,0 +1,50 @@
...
+use weaver_resolved_schema::{
+    body::{Body, BodyField},
+    error::{handle_errors, Error},
+};
+use weaver_semconv::body::{BodyFieldSpec, BodySpec};
+
+/// Resolve a `Body` specification into a resolved `Body`.
+pub fn resolve_body_spec(body: &BodySpec) -> Result<Option<Body>, Error> {
+    match body {
+        BodySpec::Fields { fields } => {
```
jsuereth, line 14: If fields aren't being resolved, do we need to resolve the whole spec as well?

crates/weaver_resolved_schema/src/error.rs
```diff
@@ -20,6 +22,13 @@ pub enum Error {
     /// A generic container for multiple errors.
     #[error("Errors:\n{0:#?}")]
     CompoundError(Vec<Error>),
+
+    /// A generic error identifying a feature that has not yet been implemented.
+    #[error("Not Implemented: {message}")]
+    NotImplemented {
```
jsuereth, line 28: is this  still used?

crates/weaver_semconv/src/body.rs
```diff
@@ -0,0 +1,244 @@
...
+    pub stability: Option<Stability>,
+    /// Specifies if the body field is deprecated. The string
+    /// provided as <description> MUST specify why it's deprecated and/or what
+    /// to use instead. See also stability.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub deprecated: Option<String>,
+}
+
+impl BodyFieldSpec {
+    /// Returns true if the body field is required.
```
jsuereth, line 89: Should this also return conditionally required or not?

jsuereth:
Nit: The PR should be marked as `feat` not `chore`

crates/weaver_resolved_schema/src/attribute.rs
```diff
@@ -91,7 +91,7 @@ pub struct UnresolvedAttribute {
 
 /// An internal reference to an attribute in the catalog.
 #[derive(
-    Serialize, Deserialize, Debug, Copy, Clone, Eq, PartialEq, PartialOrd, Ord, JsonSchema,
+    Serialize, Deserialize, Debug, Copy, Clone, Eq, PartialEq, Hash, PartialOrd, Ord, JsonSchema,
```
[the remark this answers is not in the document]
MSNev: At the time of adding this was required as the compilation failed because this was not tagged as implementing this, this is (probably) a left over of the now removed definition to return an Attribute Group with Ref. Will validate and remove if no longer required.

crates/weaver_resolver/src/body.rs
MSNev: Yeah, this was originally here to support the now defunct "Ref" Body field

crates/weaver_resolved_schema/src/error.rs
MSNev: Will check

MSNev: Based on keeping the `BodySpec::Value { value: _ }` then yes, as there are some event producers (and SDKs) that have been talking about just sending a string.

This is also part of where / how the body definition (might) be extended as per the [discussion here](https://github.com/open-telemetry/semantic-conventions/issues/755)

crates/weaver_semconv/src/body.rs
MSNev: This is really the `BodyFieldSpec` version of the to the `AttributeSpec` `is_required` (which is not currently checking for any conditionally required)/ So my simple answer is no. But it's also currently not directly used as we pulled out the markdown examples. So to keep them in-sync, my simple answer to this is No.

crates/weaver_resolved_schema/src/body.rs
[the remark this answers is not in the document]
MSNev: Leaving for now, until we see the PR's from ted and we can discuss further.

crates/weaver_resolved_schema/src/attribute.rs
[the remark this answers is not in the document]
MSNev: Removed as no longer required

crates/weaver_resolver/src/body.rs
MSNev: I've kept this function as it contains the `Value` definition, and it keeps the location for any potential resolution of any possible type (maybe) separate (for now) -- may be a candidate for removal based on on-going discussions and just moving this inline.

crates/weaver_resolver/src/body.rs
MSNev: Removed and inlined

lquerel (commented):
My main concern is the introduction of backward-incompatible changes in the span events. I believe there is a way to modify the event definition to maintain backward compatibility (see my comments on the event definition). Otherwise, I'm ok with this PR.

crates/weaver_resolved_schema/src/signal.rs
```diff
@@ -53,8 +54,6 @@ pub struct Event {
     /// References to attributes defined in the catalog.
     #[serde(skip_serializing_if = "Vec::is_empty")]
     attributes: Vec<AttributeRef>,
-    /// The domain of the event.
-    domain: String,
```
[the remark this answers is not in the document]
MSNev: > By removing this field, I believe we can no longer load and parse previous versions of the registry.

This was never a "top" level value and no "log" based events had ever defined in yaml or supported in semconv (or build tools). And the previous event "domain" was defined as an "attribute" associated with the event only and not a specific. As far as I can tell this is also never handled for anything else today either -- I'll go and recheck build-tools, but I don't recall seeing that there.

> Another similar issue is the renaming of the event_name field to name. While using name consistently across the signals is preferable, we should implement a serde alias for the name field to allow event_name for backward compatibility.
Nice catch -- I'll add that back

> attributes was also an optional field in the span events. Since attributes are no longer part of this generic event, it introduces another source of incompatibility with previous versions of the registries.
??? Optional Attributes are part of the standard event and I believe are still present (as (apart from the `event_name` to `name`) I just renamed the existing `SpanEvent` to `Event`...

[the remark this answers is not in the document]
MSNev: > `domain` removal

Just checked build-tools and prior version of weaver, the `domain` is only specified on the `Event` (log based event) which is not supported in any version of semconv yet, so I think this is fine.

> `attributes` removal

I just checked the `Event` and `SpanEvent are identical except for the `event_name`, so (I think) this is fine also.

schemas/semconv-syntax.md
```diff
@@ -102,7 +105,12 @@ specificfields ::= spanfields
                |   metricfields
 
 spanfields ::= [events] [span_kind]
-eventfields ::= [name]
+
+eventfields ::= [name] [body]
```
lmolkova, line 109: I believe name is required ([EBNF grammar](https://en.wikipedia.org/wiki/Extended_Backus%E2%80%93Naur_form#:~:text=In%20computer%20science%2C%20extended%20Backus,as%20a%20computer%20programming%20language.)), so it should be

```suggestion
eventfields ::= name [body]
```

crates/weaver_resolved_schema/src/signal.rs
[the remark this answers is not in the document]
jsuereth: It's not about whether it's a fine, it's about whether using weaver on previous versions of semconv crashes or continues working as it used to.

I think you need to keep those things and deprecate them / ensure they get removed from semconv before you can remove them from weaver.

crates/weaver_semconv/src/body.rs
```diff
@@ -0,0 +1,244 @@
...
+#[serde(rename_all = "snake_case")]
+pub enum BodySpec {
+    /// The collection of body fields associated with a body definition
+    Fields {
+        /// The collection of body fields associated with a body definition
+        #[serde(skip_serializing_if = "Vec::is_empty")]
+        fields: Vec<BodyFieldSpec>,
+    },
+    /// The body field value.
+    Value {
```
jsuereth, line 27: What's the difference between value and fields?

schemas/semconv-syntax.md
```diff
@@ -102,7 +105,12 @@ specificfields ::= spanfields
...
+
+body ::= body_fields
+
+body_fields ::= id type brief [examples] stability [deprecated] [requirement_level] [note]
```
lmolkova, line 113: not blocking:

I know we don't require `requirement_level` for attributes, but if I remember correctly, we wanted to start requiring it when referencing attribute.

I wonder if we can require it for events in this PR and take attributes separately.

If it's not possible in the actual implementation (because it would mean changing things for attributes), can we at least do it on the schema level?

@lquerel @jsuereth wdyt?  maybe we can discuss it on Wed tooling call.

lmolkova (approved):
Schema changes LGTM, thanks!

crates/weaver_resolved_schema/src/signal.rs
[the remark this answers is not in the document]
lmolkova: I don't  see any usages of domain in semconv and it was never supported by the build-tools, so I think it should be fine to remove right away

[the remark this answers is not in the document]
lquerel: Please disregard my previous comment on `attributes`; they are indeed included at the group level. At this point, I believe we can simply add an alias for the `name` field to support the previous `event_name`, and then add a test to verify this.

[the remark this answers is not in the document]
MSNev: > It's not about whether it's a fine, it's about whether using weaver on previous versions of semconv crashes or continues working as it used to.
> 
> I think you need to keep those things and deprecate them / ensure they get removed from semconv before you can remove them from weaver.

Thats what I was trying to describe, `domain` has never been defined / used in any previous versions of sem-conv (at least as far as I can tell) and it's not been defined in the old build tools. It's just that weaver defined a `Event` definition that has never been used and I'm now reusing (and making active) that stuct.

crates/weaver_semconv/src/body.rs
MSNev: https://github.com/open-telemetry/semantic-conventions/issues/755

crates/weaver_resolved_schema/src/signal.rs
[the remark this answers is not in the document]
MSNev: I've added several of these as topics for wednesday's meeting, as I added the #serde rename for the `event_name` and it didn't seem to do what I expected (ie. all of the json tests continued to pass), and as we don't have any "markdown" or "codegen" tests for "events" I suspect that the existing generations that (might) reference `event_name` won't work with this change alone.

schemas/semconv-syntax.md
MSNev: Agree, I've applied this change locally to my PR but will wait to push any changes until after the `event_name` discussion (just to limit the number of changes / pushes)

schemas/semconv-syntax.md
MSNev: Agree, as nothing exists today, if we want to always enforce then this sounds like the time to make that change. Although, technically we are not "referencing" anything else today. But I agree that starting out being tighter and loosing later is easier.

crates/weaver_semconv/src/body.rs
jsuereth: Commented my thoughts here.  Seeing this in action - I'm not a fan.

schemas/semconv-syntax.md
```diff
@@ -159,6 +167,77 @@ The following is only valid if `type` is `span` (the default):
...
+#### Event semantic convention example
+  
+  ```yaml
+  - id: event.some_event
+    name: the.event.name
+    type: event
+    brief: "Describes the event."
+    stability: experimental
+    attributes:                                        # Optional
+      - id: registry.attribute.id
```
lmolkova, line 217: nit: we don't allow attribute definitions outside of the registry

```suggestion
      - ref: registry.attribute.id
```

schemas/semconv-syntax.md
lquerel: This is outside the context of this PR, but personally, I am not convinced that forcing the override of the `requirement_level` field for each `ref` on an attribute is the right approach, and this for several reasons:

- It is possible to have multiple levels of inheritance at the level of `attribute_groups`. For example, the first level of inheritance converts certain requirement levels from `opt_in` to `required`. In the following levels, we might have several branches where the `brief` / `note` of the attributes referenced in the parent level are customized based on the context for which these `attribute_group`s are intended. Saying that we force the definition of the `requirement_level` at each `ref` would imply, in this scenario, redefining the requirement level in each branch with exactly the same information, which seems suboptimal to me.
  
- More generally, I find that the `requirement_level` of an attribute is not ideally decided at the OTel registry level. It is more of a decision/directive of the end-user who, based on their own policies, might consider that a certain attribute should be required or even eliminated. This, in my view, does not advocate for making the specification of the `requirement_level` at each `ref` necessarily evident, especially for intermediate levels. In the context of multi-registry support, I would like to propose something to address this. More details on that soon.

cc @lmolkova @jsuereth

crates/weaver_resolved_schema/src/signal.rs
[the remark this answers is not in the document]
lquerel: @MSNev, this is because you should use the annotation `#[serde(alias = "event_name")]` instead of the `rename` annotation. The `rename` annotation instructs Serde to use only the name specified in the `rename` statement, instead of using both the field name and the name specified in the `rename` annotation. The `alias` annotation does exactly what you need.

schemas/semconv-syntax.md
lmolkova: I agree it's tricky for attributes, but here we're talking about event fields. 
We don't have a registry of them, they are not inherited or referenceable, so these problems should not(?) affect them

lmolkova: BTW do we support event body inheritance? I.e. if I define

```yaml
- id: event.a
  name: event.a
  type: event
  ...
  attributes:  ...
  body:
  
- id: event.b
  name: event.b
  extend: event.a
```

what's going to happen? 

I'd prohibit it (one event extending another) for the time being. Someone always can use attribute_group to avoid copy-paste

crates/weaver_resolved_schema/src/signal.rs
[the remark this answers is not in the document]
jsuereth: FYI - here's where event domain was removed: https://github.com/open-telemetry/semantic-conventions/pull/473

crates/weaver_resolved_schema/src/body.rs
```diff
@@ -0,0 +1,99 @@
...
+    /// Sequence of example values for the body or single example
+    /// value. They are required only for string types. Example values
+    /// must be of the same type of the body. If only a single example is
+    /// provided, it can directly be reported without encapsulating it
+    /// into a sequence/dictionary.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub examples: Option<Examples>,
+    /// Identifies the definition of the "fields" of the body when the body type is "map".
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub fields: Option<Vec<BodyField>>,
```
jsuereth, line 40: I think I'd remove the fields from here and encode them in the `BodyType` struct, flattening that into this struct for disjoint mapped union.

crates/weaver_semconv/src/body.rs
```diff
@@ -0,0 +1,361 @@
...
+        }
+    }
+}
+
+/// Identifies the different types of body (specification).
+#[derive(Serialize, Deserialize, Debug, Clone, Eq, PartialEq, Hash, JsonSchema)]
+#[serde(rename_all = "snake_case")]
+pub enum BodyType {
+    /// A map body type.
+    Map,
```
jsuereth, line 89: I think it'd be better if this enum actually had values.  E.g.  Map would have nested `TreeMap<String, BodyType>` or `TreeMap<String,BodyField>` -> where `BodyField` has `requirement_level`, `notes`, etc.

I don't think my point about not having two ways of representing compositionality made it through.   What I want to avoid here is having TWO enums you need to walk through to understand the type system in semconv.   Right now, `Map` doesn't really cover the fields in the map, meaning I need to walk through two strutures.

I don't have a ton of free time today, but I can put together an example/suggested hierarchy here, but I envision something like:


```rust
trait BodySpec {
  // Whatever serde flag denotes tagged-disjoin union
  r#type: BodyType
}
enum BodyType {
  String(CommonAttributes),
  Map {
     fields: TreeMap<String, BodyType>,
     // whatever serde attribute denotes flattening.
     attributes: CommonAttributes,
  }
}
struct CommonAttributes {
  requirement_level: RequirementLevel,
  notes: Option<String>,
  ...
}
```

MSNev: OK, I think I'm following your suggestion and fundamentally it ends up representing the same structured output (without the existing duplication), not sure I'll get to (and complete) today before tomorrows meeting but I think I like this direction (I guess I'm still getting my head wrapped around the rust structured enum definition / usage)

MSNev: Created draft PR with some simple examples (tests) to discuss the definition, I'm going to start to merge the changes from that PR into this one (and any changes we make to that one)
https://github.com/open-telemetry/weaver/pull/365

MSNev:
Please also see comments and discussion topics in #365

MSNev:
AnyValue PR and this PR now have the same (common) set of changes.

crates/weaver_semconv/src/group.rs
```diff
@@ -220,6 +256,62 @@ impl GroupSpec {
...
+fn validate_any_value_examples(
+    errors: &mut Vec<Error>,
+    any_value: Option<&AnyValueSpec>,
+    group_id: &str,
+    path_or_url: &str,
+) -> WResult<(), Error> {
+    if let Some(value) = any_value {
+        let common_examples = &value.common().examples;
+        if let Some(examples) = common_examples {
+            match examples.validate_any_value(value, group_id, path_or_url) {
```
jsuereth, line 268: @lquerel don't we have a helper to simplify this code?

crates/weaver_resolver/data/registry-test-4-events/registry/trace-gen-ai.yaml
jsuereth, line 1: I mentioned this before, but I'm not a huge fan of copy-pasting in a ton of existing semconv YAML.

We did this initially in weaver, and I don't think it's been easily maintainable nor does it help us understand code-coverage of whether or not we have the right feature set. 

I think as a one-off or test of "can we resolve existing semconv" it's ok, but for testing resolution and scenarios of resolution -> It's hard for me to understand what each file is testing and what I should expect on the other side vs. clearly/more granular scenarios.

Consider this non-blocking, but also be prepared if one of us comes in and tries to refactor / slice these tests into smaller, more maintainable chunks (I've been wanting to do that for weaver for some time, just haven't found the time).

jsuereth (approved):
This looks good to me now.

I'm still unhappy with the direction our tests have been going in and I'd much prefer finer-grained, feature or regresison focused tests, but we can clean that up later.

crates/weaver_resolver/data/registry-test-4-events/expected-registry.json
```diff
@@ -1,15 +1,160 @@
...
+        1,
+        2,
+        3
+      ],
+      "name": "browser.test.event_with_fields",
+      "lineage": {
+        "source_file": "data/registry-test-4-events/registry/browser-event.yaml"
+      },
+      "body": {
+        "name": "browser.test.event_with_body.fields",
```
MSNev, line 20: Linking for the comment in the AnyValue PR (https://github.com/open-telemetry/weaver/pull/365#discussion_r1767768808) these "name" entries become `id`

```json
      "body": {
        "id": "browser.test.event_with_body_details.fields",
        "type": "map",
        "brief": "A map of fields that are not referenced in the attributes",
        "note": "This map is not referenced in the attributes",
        "stability": "experimental",
        "examples": [
          "{ \"some.field\": \"some value\" }"
        ],
        "requirement_level": "required",
        "fields": [
          {
            "id": "some.field",
            "type": "string",
```

crates/weaver_resolver/data/registry-test-4-events/expected-registry.json
```diff
@@ -45,29 +190,354 @@
...
         "source_file": "data/registry-test-4-events/registry/mobile-events.yaml"
+      },
+      "body": {
+        "name": "device.app.lifecycle.fields",
+        "type": "map",
+        "requirement_level": "required",
+        "fields": [
+          {
+            "name": "ios.state",
+            "type": "enum {active, inactive, background, foreground, terminate}",
```
MSNev, line 210: Also for AnyValue comment https://github.com/open-telemetry/weaver/pull/365#discussion_r1767768808

This becomes (`id` instead of `name` and the `enum` for the type instead of `enum {active, inactive, background, foreground, terminate}`

```json
      "body": {
        "id": "device.app.lifecycle.fields",
        "type": "map",
        "requirement_level": "required",
        "fields": [
          {
            "id": "ios.state",
            "type": "enum",
            "members": [
              {
                "id": "active",
```

crates/weaver_semconv/src/group.rs
lquerel: I'm not aware of any, but this function could be simplified as follows.

```rust
fn validate_any_value_examples(
    errors: &mut Vec<Error>,
    any_value: Option<&AnyValueSpec>,
    group_id: &str,
    path_or_url: &str,
) -> WResult<(), Error> {
    if let Some(value) = any_value {
        if let Some(examples) = &value.common().examples {
            match examples.validate_any_value(value, group_id, path_or_url) {
                WResult::Ok(_) => {}
                WResult::OkWithNFEs(_, errs) => errors.extend(errs),
                WResult::FatalErr(err) => return WResult::FatalErr(err),
            }
        } else {
            match value {
                AnyValueSpec::String { .. } | AnyValueSpec::Strings { .. } => {
                    errors.push(Error::InvalidAnyValueExampleError {
                        path_or_url: path_or_url.to_owned(),
                        group_id: group_id.to_owned(),
                        value_id: value.id(),
                        error: format!("This value is a {} but it does not contain any examples.",
                            if let AnyValueSpec::String { .. } = value { "string" } else { "string array" }),
                    });
                }
                _ => {}
            }
        }

        if let AnyValueSpec::Map { fields, .. } = value {
            for field in fields {
                if let WResult::FatalErr(err) =
                    validate_any_value_examples(errors, Some(field), group_id, path_or_url)
                {
                    return WResult::FatalErr(err);
                }
            }
        }
    }

    WResult::Ok(())
}
```

crates/weaver_resolver/data/registry-test-4-events/registry/trace-gen-ai.yaml
lquerel: I'm not saying the tests are perfect, but initially, the test scenario `registry-test-4-events` contained only 3 files, all focused on event resolution. Now, we have 16 files. I'm confident we can significantly reduce the number and size of these files to focus solely on what we want to check in this specific scenario. 
@MSNev It’s not blocking, but it would be great to have this simplification before the merge, if possible.

lquerel (commented):
@MSNev, @jsuereth  I haven’t finished the review, but I will do it in a few hours after my meetings. In the meantime, I’m posting some of my preliminary comments. This is definitely heading in the right direction!

crates/weaver_resolver/data/registry-test-4-events/observed-events.json
```diff
@@ -0,0 +1,219 @@
+[
```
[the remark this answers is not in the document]
MSNev: Yeah, we shouldn't

crates/weaver_resolver/data/registry-test-4-events/registry/trace-gen-ai.yaml
MSNev: Yeah, I added the additional files to ensure that for all current (and coming -- the gen-ai ones) event definitions that we could support them as expected. I'll review these again later today as I need to removed that observed file and we need to come to consensus if we want to remove the AnyValue in preference to just using AnyValueSpec as per the comment links in the AnyValue PR.

lquerel: Great. Regarding `AnyValue` vs `AnyValueSpec`, I recommend going with `AnyValueSpec` only for now to minimize duplication. We can always change it later if the need arises, without breaking the external interface with JQ and Jinja2.

crates/weaver_forge/templates/test/event.md
```diff
@@ -1,10 +1,10 @@
-{%- set file_name = ctx.root_namespace | snake_case -%}
+{%- set file_name = ctx.event_namespace | snake_case -%}
 {{- template.set_file_name("event/" ~ file_name ~ ".md") -}}
 
-## Events Namespace `{{ ctx.root_namespace }}`
+## Events Namespace `{{ ctx.event_namespace }}`
 
 {% for event in ctx.events %}
-## Event `{{ event.name }}`
+## Event `{{ event.id }}`
```
lquerel, line 7: Just a question: Why using `id` and not `name` here?
Not blocking.

crates/weaver_forge/templates/semconv_jq_fn/weaver.yaml
```diff
@@ -20,3 +20,12 @@ templates:
...
       semconv_grouped_metrics({"exclude_stability": ["experimental"]})
     application_mode: single
+  - pattern: semconv_events.json
+    filter: >
+      .groups
+      | map(select(.type == "event"))
+      | map(. + {
+          event_namespace: (if .id | index(".") then (.id | split(".") | .[0:-1] | join(".")) else "other" end)
+        })
+      | sort_by(.event_namespace, .id)
```
lquerel, line 30: Just a question: Why use `event_namespace` instead of `root_namespace`, as with the other signals? I just want to avoid the same kind of issues we had in the past with `name` and `metric_name`, which made it harder to generalize without providing significantly higher value for the UX. 

Not blocking.

crates/weaver_resolver/src/lib.rs
```diff
@@ -145,6 +146,17 @@ pub enum Error {
     /// A container for multiple errors.
     #[error("{:?}", format_errors(.0))]
     CompoundError(#[related] Vec<Error>),
+
+    /// An error indicating that the Body references are not resolved.
+    #[error("The following body reference is not resolved.\nBody reference: {group_id}\nProvenance: {provenance}\nError: {error}")]
+    UnresolvedBody {
```
lquerel, line 152: Not sure this error is required anymore.

lquerel (approved):
Once the merge with the other PR (AnyValueSpec) is done, and the few minor corrections mentioned in my comments are made, I see no problem with creating a release with this PR. Please update the `CHANGELOG.md` to announce the support of events in Weaver (section [Next]).

However, as a precaution, I think it would be a good idea to check the backward compatibility with the older registries. The previous version was able to check all versions from 1.21.0 to the latest one. You can test each version with the following command line.

```
cargo run -- registry check -r https://github.com/open-telemetry/semantic-conventions/archive/refs/tags/v1.21.0.zip[model]
```

Thanks for all your hard work on this!

crates/weaver_forge/templates/semconv_jq_fn/weaver.yaml
jsuereth: +1 to this from me.  We should avoid diverging here, although IIUC this is just testing code right?

We can clean this up later

crates/weaver_forge/templates/test/event.md
MSNev: Actually, the top level "name" will still exist on the `event` type, its just the `AnyValues` that will now only have the `id` (and no `name`)

crates/weaver_forge/templates/semconv_jq_fn/weaver.yaml
MSNev: Yes, it is just testing code. And for events the entire "event_namespace" is important and not just the "root_namespace"...

crates/weaver_resolver/src/lib.rs
MSNev: Yep, it no longer seems to be referenced -- removing and testing locally

crates/weaver_semconv/src/attribute.rs
```diff
@@ -490,6 +491,73 @@ impl Examples {
             ),
         }
     }
+
+    /// Validation logic for the any_value.
+    pub(crate) fn validate_any_value(
```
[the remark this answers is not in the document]
MSNev: I kept it here with the other "Examples" validate so that there was only one location for handling "Examples"...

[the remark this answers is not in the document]
MSNev: Local backward compat check past with the expected warnings about examples SHOULD be of type string[]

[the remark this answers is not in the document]
lquerel: Ok. So we are ready to go?

PR #300, opened 2024-08-08
[WIP] feat(resolver): Validate examples

lquerel:


crates/weaver_resolver/src/lib.rs
```diff
@@ -253,6 +255,7 @@ impl SchemaResolver {
     fn load_semconv_from_local_path(
         local_path: PathBuf,
         registry_path_repr: &str,
+        strict_mode: bool,
```
jsuereth, line 258: Why are we introducing strict mode?

lquerel: Because all the existing versions of SemConv have issues in their examples.

lquerel: Changed strict mode to future mode

crates/weaver_semconv/src/attribute.rs
```diff
@@ -429,6 +430,59 @@ pub enum Examples {
...
+                // enum types are open so it's not possible to validate the examples
+                Ok(())
+            }
+            // Only if future mode is disabled, we allow to have examples following
+            // the conventions used in semconv 1.27.0 and earlier.
+            (Examples::Ints(_), PrimitiveOrArray(PrimitiveOrArrayTypeSpec::Ints))
+            | (Examples::Doubles(_), PrimitiveOrArray(PrimitiveOrArrayTypeSpec::Doubles))
+            | (Examples::Bools(_), PrimitiveOrArray(PrimitiveOrArrayTypeSpec::Booleans))
+            | (Examples::Strings(_), PrimitiveOrArray(PrimitiveOrArrayTypeSpec::Strings))
+                if !future_mode =>
```
jsuereth, line 468: Another alternative, instead of wiring future mode down into every single module, is to find a way to put it on diagnostics and elevate warnings into errors.

Sadly - We don't really have a result type that allows this today, and we'd need to spend more time on the ole diagnostic framework setup.

For now this is fine, we can expand our capabilities if `--future` winds up needing more central features and controls.

src/registry/check.rs
```diff
@@ -46,6 +46,11 @@ pub struct RegistryCheckArgs {
     #[arg(long, default_value = "false")]
     pub display_policy_coverage: bool,
 
+    /// Enable the most recent validation rules for the semconv registry. It is recommended
+    /// to enable this flag when checking a new registry.
```
jsuereth, line 50: Maybe add a note that the `semantic_conventions` main branch should always enable this flag?

jsuereth (approved):

crates/weaver_semconv/src/attribute.rs
lquerel: I believe it’s actually possible with the existing diagnostic framework. I’ll try to implement it before the merge.

src/registry/check.rs
lquerel: Comment updated.

crates/weaver_semconv/src/attribute.rs
lquerel: I have completely restructured the implementation of this flag. Now, the code generates Rust warnings for items that were acceptable in the previous version of the semantic conventions but are considered errors when the `--future` flag is enabled. This change globally converts warnings into errors. The CLI’s error code has also been updated to reflect these changes.
Please review the modifications.

jsuereth (approved):

PR #307, opened 2024-08-09
Move semconv schema definition from build tools

lmolkova:
Fix https://github.com/open-telemetry/semantic-conventions/issues/916

(copy-paste from https://github.com/open-telemetry/build-tools/blob/15a0e831e0b0818ed666378adc85cba1fa66beef/semantic-conventions/semconv.schema.json)


.github/CODEOWNERS
```diff
@@ -0,0 +1,21 @@
+#####################################################
```
lquerel, line 1: Do know if that can be a source of conflict but there is already a CODEOWNERS file at the root of this project.

lquerel (approved):
I left a comment regarding CODEOWNERS.

.github/CODEOWNERS
lmolkova: oh, thanks! I didn't  realize there is one (and didn't know it could be in root). Fixing

lquerel:
@lmolkova LGTM but I'd like to get @jsuereth feedback on this one to double the CODEOWNERS changes.

jsuereth (approved):

jsuereth:
Reagarding: https://github.com/open-telemetry/community/blob/85f75d75789938ddfb42a9e3f251b85461cb210a/config.yaml.  We're moving to an git-ops like repository setup - They must have missed weaver in that, so we need to add it there.

PR #310, opened 2024-08-12
(chore) Bump ratatui, textarea and crossterm version in lock-step.

jsuereth:
Fixes the dependabot issues where they needed to be bumped together.

Also removes deprecated calls.


lquerel (approved):
LGTM

PR #318, opened 2024-08-15
Stop using root user in docker image.

jsuereth:
Also adds dockerignore to dramatically reduce context fired at docker for building and increase re-use for docker development purposes.


lquerel (approved):
Cool

PR #322, opened 2024-08-15
Buildx setup needed for build attestations, it seems.

jsuereth:


lquerel (approved):
Approved but there is a warning in the Dockerfile raised by the GH action

PR #336, opened 2024-08-21
Bump cargo-dist version and add github attestations.

jsuereth:
This updates us to the latest `cargo-dist` version (which restructures github actions a bit) and then also adds github attestations (so you can `gh attest` weaver).

Context: https://github.com/open-telemetry/sig-security/issues/47


Cargo.toml
```diff
@@ -206,17 +206,22 @@ string_to_string = "warn"
...
 ci = "github"
 # The installers to generate for each app
 installers = ["shell", "powershell", "msi"]
 # Target platforms to build apps for (Rust target-triple syntax)
 targets = ["aarch64-apple-darwin", "x86_64-apple-darwin", "x86_64-unknown-linux-gnu", "x86_64-pc-windows-msvc"]
-# Publish jobs to run in CI
+# Which actions to run on pull requests
 pr-run-mode = "plan"
 # Whether to install an updater program
 install-updater = false
```
lquerel, line 219: Should we revert the change we did this week on this property?

lquerel (approved):
LGTM. Just one question regarding `install-updater`.

Cargo.toml
jsuereth: Unclear - I'm fine not adding that back in until we've tested it more.

PR #337, opened 2024-08-21
Refactor handling of non-fatal errors, including warnings.

### LINKED ISSUE #334: `--future` is also applied to resolving `--baseline_registry`
**Describe the bug**
Today, the `--future` keyword adds model enforcements that are meant to protect future versions of semconv YAML files from using deprecated features.  However, this flag is used when specifying a `baseline_registry` for comparing current version to previous.

**To Reproduce**
1. Set `--baseline_registry={any semconv version before main}`, Set `--future`

**Expected behavior**
Baseline registry resolution should allow non-future features and warnings against it should be suppressed when evaluating against it as a "baseline".

**Additional context**
See: https://github.com/open-telemetry/semantic-conventions/actions/runs/10479733554/job/29025956727?pr=1359

lquerel:
This PR introduces a new result type named `WResult`, which distinguishes between non-fatal errors (NFEs) and fatal errors. NFEs do not prevent subsequent operations from completing successfully. For example, if a semconv file is invalid, a non-fatal error is generated, but processing of the other files continues.

**This design will reduce the risk of promoting non-fatal errors to fatal errors, as occurred in the previous design.**

With this PR:
- warnings detected in the baseline registry are now ignored,
- non-fatal errors will not interrupt any command before it completes.  

Closes #334


jsuereth (commented):
Ironically, this looks a lot like my initial proposal :)

This looks like a good approach.

crates/weaver_common/src/result.rs
```diff
@@ -0,0 +1,180 @@
...
+            WResult::OkWithNFEs(result, nfes) => f(result, Some(nfes)),
+            WResult::FatalErr(_) => {}
+        }
+
+        self
+    }
+
+    /// Converts a [`WResult`] into a standard [`Result`], potentially
+    /// aggregating non-fatal errors into a single error.
+    pub fn into_result(self) -> Result<T, E> {
```
jsuereth, line 104: NIt: I actually think ignoring non_fatal_errors could be the default here.  I'm not certain, but I'd suggest `into_result_*` for all methods instead of an `into_result` that isn't explicit:

- `into_result_with_non_fatal` -> `Result<(T,Vec<E>), E>`
- `into_result_ignoring_non_fatal`  -> `Result<T,E>`
- `into_result_failing_non_fatal` -> `Result<T,E>`

Or some such.

crates/weaver_common/src/result.rs
```diff
@@ -0,0 +1,180 @@
...
+            } else {
+                WResult::OkWithNFEs(result, errors)
+            }
+        } else {
+            self
+        }
+    }
+
+    /// Return a [`WResult`] without the warnings.
+    pub fn ignore_warnings(self) -> WResult<T, E> {
```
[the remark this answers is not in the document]
lquerel: Among the non_fatal_errors, we have two kinds of severity: warning and error (cf. miette severity annotation). The idea here is to filter out non-fatal errors with severity=warning. We could make the method name more explicit by using `ignore_severity_warning`.

crates/weaver_common/src/result.rs
lquerel: I agree. It’s wordy, but at least it avoids ambiguities in use.

crates/weaver_semconv/src/registry.rs
```diff
@@ -64,20 +65,37 @@ impl SemConvRegistry {
...
-            error: e.to_string(),
-        })? {
-            let path_buf = sc_entry.map_err(|e| Error::InvalidRegistryPathPattern {
-                path_pattern: path_pattern.to_owned(),
-                error: e.to_string(),
-            })?;
-            let semconv_spec = SemConvSpecWithProvenance::from_file(path_buf.as_path())?;
-            registry.add_semconv_spec(semconv_spec);
+    pub fn try_from_path_pattern(registry_id: &str, path_pattern: &str) -> WResult<Self, Error> {
+        fn create_registry_or_fatal(
```
lquerel, line 69: This pattern simplifies the management of fatal errors using a simple `?` operator. This internal method populates a list of non-fatal errors and returns a `Result`. This `Result` is then transformed into a `WResult` depending on the presence or absence of non-fatal errors.

crates/weaver_common/src/result.rs
```diff
@@ -0,0 +1,230 @@
...
+            } else {
+                WResult::OkWithNFEs(result, errors)
+            }
+        } else {
+            self
+        }
+    }
+
+    /// Return a [`WResult`] without the non-fatal errors with severity=warning.
+    pub fn ignore_severity_warnings(self) -> WResult<T, E> {
```
jsuereth, line 76: This is fine for now, but we should consider an `ignore(F) -> WResult<T,E> where F: FnOnce(E) -> Bool`

crates/weaver_resolver/src/lib.rs
```diff
@@ -316,40 +301,36 @@ impl SchemaResolver {
...
 
-        let mut error = vec![];
-        let result = result
-            .into_iter()
-            .filter_map(|r| match r {
-                Ok(r) => Some(r),
-                Err(e) => {
-                    error.push(e);
-                    None
+        // Process all the results of the previous parallel processing.
```
jsuereth, line 317: NIt: We should turn this into a helper method eventually.

crates/weaver_semconv/src/registry.rs
jsuereth: This gets around limitations in the current expeirmental custom type `?` support for sure.  I'd still like to go that direction with WResult when it stabilizes, but for now this seems to be our best bet.

I may take some time to write up a guide on using WResult that includes this, but sadly I've been highly limited on time recently (possibly when entities OTEP has progressed past prototyping, I can come back and revisit this)

jsuereth (approved):

crates/weaver_common/src/result.rs
lquerel: That’s something that crossed my mind at some point, but then I forgot about it. I will change that before the merge.

crates/weaver_resolver/src/lib.rs
lquerel: I suggest implementing this helper once we encounter more than one instance of this pattern.

PR #346, opened 2024-08-28
Fix #345 - Build ARM64 image.

### LINKED ISSUE #345: Publish docker image with for `arm64` platform
While incorporating weaver into `semantic-conventions-java`, we discovered that the weaver docker image doesn't have multi platform support, so running it in arm64 requires this extra argument. 

> Apparently the `otel/weaver` image doesn't publish for `arm64` platform yet. 

```suggestion
        "--rm",
        "--platform=linux/x86_64",
```

_Originally posted by @jack-berg in https://github.com/open-telemetry/semantic-conventions-java/pull/70#discussion_r1731439609_

jsuereth:
Follows the docker guide for setting up QEMU to build an ARM + x86 image.

Fixes #345 

Note: This may not be the fastest way to do these builds, it's just the most convenient.

### DIFF
--- a/.github/workflows/publish-docker.yml
+++ b/.github/workflows/publish-docker.yml
@@ -33,7 +33,7 @@ jobs:
       uses: docker/build-push-action@v6.6.1
       with:
           push: false
-          platforms: linux/amd64,linux/arm64
+          load: true
           tags: ${{ env.TEST_WEAVER_TAG }}
           labels: ${{ steps.meta.outputs.labels }}
     - name: Test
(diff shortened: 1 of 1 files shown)


jsuereth:
@lquerel It looks like building both in *test* takes a really long time.  Do we think it's worth having the previous test-image build we had and assuming building for ARM in docker will succeed on release or should we perform it in every pull request?

jsuereth:
Oof - so here' s an alternative: https://docs.docker.com/build/ci/github-actions/multi-platform/#distribute-build-across-multiple-runners

Take a look and let me know what you think - I'll look into the more complicated setup a bit later.

jsuereth:
I updated this so the combined image is only created on release for now.

lquerel (approved):

Issue #215, opened 2024-06-19
Support multiple semantic convention registries


lquerel:
We'd like to be able to combine multiple registries into a single master resolved semantic convention registry. For example, combining the official OpenTelemetry semantic convention with one or several vendor/enterprise registries.

MadVikingGod:
Here are some thoughts on how this might look.  Don't take any of this as how it should be done.

### Base Registry
I think there should be a base registry, this is what dictates what other registries or parts of registries should be included and what should be over written.  These are assumed to not be written by otel, but by customers using this feature.

### Includes (Dependency Management)
Here there is a trade off. We could require that any dependency is fully resolved (into a single json/yaml file), or we could have a more ephemeral reference (to a directory or git repository).  The former will be a usability hurdle, because it puts more work up front to use this feature, while the former will introduce problems like availability and circular dependencies.  I would probably suggest the former at least for prototypes.

### What is Rendered
When rendering documentation or code generation we can either include all of any included registries or we can opt to only include the portions that are referenced.  We should make it easy to include groups from a different registry.

### References
I think everything but an `include` (the high level name I made for repository reference) should be a referenceable type, so right now groups, and attributes.  These should be referenced by both their repository name and the name in that repository, and any field should be overwriteable.  


An example:
```yaml
includes:
- name: "otel_1.27.0"
  file: "includes/otel-v1.27.0.json"
groups:
- ref_repo: otel_1.27.0
  ref: metric.http.server.request.duration
  unit: "ms"
  attributes:
  - id: environment
    type: string
    brief: This is an additional attribute included.
    examples:
    - dev
    - prod
  - ref_repo: otel_1.27.0
    ref: server.address
    brief: This overwrites an included attribute.  It should be the service name of the server.
    examples:
    - foo
    - bar
```

This should yield one metric group, a few attribute_groups that have the different attributes in `metric.http.server.request.duration`

lquerel:
I'm very pleased to see this topic being explored again (for reference, the concept of [app telemetry schema](https://github.com/open-telemetry/oteps/blob/main/text/0243-app-telemetry-schema-vision-roadmap.md) touches on the same subject with a few minor differences). I believe we are getting closer to the moment when Weaver will finally be able to support this type of use case, which, in my view, is the original use case I had in mind when starting this project.

[...]

PR #348, opened 2024-08-29
Semantic Conventions - Multi-Registry Proposal

lquerel:
This PR proposes to support multiple semantic convention registries in OTEL and Weaver.

If you’re like me and prefer to read a rendered version of the markdown spec, it’s available [here](https://github.com/open-telemetry/weaver/blob/46cd09644a1900c839fe4240d5044e71d827cba4/docs/specs/multi-registry/multi_registry.md).

Note: This proposal could eventually be transformed into an OTEP if needed.

See GH issue #215


docs/specs/multi-registry/multi_registry.md
```diff
@@ -0,0 +1,366 @@
...
+
+## Design Principles
+
+- **Independent Ownership**: Individuals or organizations should be able to create and maintain a registry
+  independently, without requiring coordination with the OTEL project.
+- **Registry Accessibility**: Registries should be either private or public, allowing flexibility based on
+  the needs of the owner. Registries should be accessible via a URL.
+- **Common Format**: A published registry should adhere to a common packaging and format, making it easy to
+  consume and integrate with other registries.
+- **Self-Contained**: A published registry should be self-contained and stored in a single file.
```
lmolkova, line 90: agree with-self-contained, but single file sounds weird as a design principle

docs/specs/multi-registry/multi_registry.md
```diff
@@ -0,0 +1,366 @@
...
+  URL.
+- **Community Support Tools**: The OTEL project should provide and maintain an extensible tool, Weaver, to
+  assist the community in creating, validating, resolving, packaging, and publishing registries. Weaver should
+  also support documentation generation and code generation for any registry.
+- **Core Policy Enforcement**: The OTEL project should establish and enforce a set of core policies (e.g.,
+  backward compatibility policies) that all registries must adhere to, ensuring consistency and reliability.
+  These policies should be enforced by Weaver.
+- **Cross-Registry References**: References between different registries should be supported, facilitating
+  interoperability and integration across various registries.
+- **Conflict Avoidance through Scoping**: A scoping mechanism should be implemented to ensure that a signal
```
lmolkova, line 101: I'd like to understand what scoping means. 

There should not be a case when two attributes with the same name are defined within one  schema_url. 

If scoping means that both exist but you can specify which one is imported, than this information is not available to the consumer of this telemetry.

docs/specs/multi-registry/multi_registry.md
```diff
@@ -0,0 +1,366 @@
...
+- Long answer: Although technically it is possible to support multiple versions of an imported registry, it is a
+  good practice to align the versions inside the same registry. This practice make the overall registry more
+  consistent and easy to reason about. Weaver will report an error if different versions are used for the same
+  registry in the scope of a single registry.
+
+#### Do we allow reference of experimental entities across registries?
+- Short answer: No for backward compatibility.
+- Long answer: Experimental entities are not meant to be used in production. Experimental entities are subject to
+  change or removal without any notice in the next version of the registry. The type of an experimental entity can
+  be changed. Weaver will detect and report an error if an experimental entity is referenced across registries.
```
lmolkova, line 196: I don't understand this limitation or reasoning behind it.

It's common to use experimental semconv in production, it's also common to implement them in OTel or outside it in the instrumentations.

It makes sense to require something that depends on experimental stuff to be experimental, but I don't understand why we should limit the ability to import experimental attribute/group.

docs/specs/multi-registry/multi_registry.md
```diff
@@ -0,0 +1,366 @@
...
+  an error if a circular reference is detected between registries.
+
+#### Do we allow references to groups or attributes defined in an imported registry?
+- Short answer: Yes.
+- Long answer: References to groups or attributes defined in an imported registry are allowed. This feature
+  allows a registry to extend or override the attributes of an imported group. Weaver will detect and report an
+  error if a reference is made to an entity that is not defined in the imported registry.
+
+### Open Questions:
+- Can we override any field of a group defined in an imported registry? No for `type` field, what about the other
```
lmolkova, line 211: I'd start with the same limitations as within the registry. You can't change type, stability, deprecation status. Can change anything else.

docs/specs/multi-registry/multi_registry.md
```diff
@@ -0,0 +1,366 @@
...
+  fields?
+- Is there a relationship to define between the instrumentation scope name and version and the registry?
```
lmolkova, line 213: instrumentation scope is a random string - name of the library/class/component that produced the telemetry, I can change which component/class emits my logs/events, but this should have no effect on the consumer of the telemetry.

I think we might change it at some point, but that's the current state of affairs.

docs/specs/multi-registry/multi_registry.md
lquerel: The idea was to make a published registry simple to consume, so in addition to being self-contained, having a single file also simplifies consumption. I’m going to remove the “single file” aspect from the design principles section for now, but I will continue to use the single file approach for the published registry in this proposal and see how things evolve.

docs/specs/multi-registry/multi_registry.md
```diff
@@ -0,0 +1,366 @@
...
+
+
+## Core Policies
+
+These core policies aim to ensure consistency and backward compatibility across registries.
+
+### Group Policies
+- No group ID collisions
+- Only attribute groups are allowed in the attribute registry
+- Ref attributes are not allowed in the attribute registry
```
lmolkova, line 223: I think we need a more formal attribute registry definition than the current one - https://github.com/open-telemetry/semantic-conventions/issues/1269

docs/specs/multi-registry/multi_registry.md
```diff
@@ -0,0 +1,366 @@
...
+- Events cannot be removed
+- Events cannot "degrade" in stability (e.g., stable -> experimental)
+- The set of required/recommended attributes must remain the same
+- Event names must match the following regular expression: `^[a-z][a-z0-9]*([._][a-z0-9]+)*$`
+
+### Spans
+- No span name collisions
+- Spans cannot be removed
+- Spans cannot "degrade" in stability (e.g., stable -> experimental)
+- The set of required/recommended attributes must remain the same
```
lmolkova, line 261: I think we need to come up with more nuanced rules. E.g.
- it's ok to change recommended attribute to required. 
- it's ok to remove recommended (if you have good reasons not to populate recommended, this is your way to tell that you don't populate it) 
- it's ok to change conditionally required to recommended or required (if condition is always true)
- it might be ok to promote opt-in to recommended or even required

maybe those rules are complex enough that we just want to start with a simple one that all required attributes should stay required but some new ones may be added

docs/specs/multi-registry/multi_registry.md
lquerel: To make things clear, the consumer of telemetry, i.e., anyone downstream from signal production (backend, dashboard, etc.), will not see this type of conflict and will see the attribute or signal with its name as defined today. I will add this principle to the design principles.

I am currently writing an entire section on conflict resolution. Conflicts can arise for several reasons. Examples:
- A registry that imports two registries, which in turn import a registry but in different versions.
- A registry that imports two registries, which in turn import the same registry but apply different overrides.

Things should be better defined soon.

docs/specs/multi-registry/multi_registry.md
```diff
@@ -0,0 +1,366 @@
...
+- No event name collisions
+- Events cannot be removed
+- Events cannot "degrade" in stability (e.g., stable -> experimental)
+- The set of required/recommended attributes must remain the same
+- Event names must match the following regular expression: `^[a-z][a-z0-9]*([._][a-z0-9]+)*$`
+
+### Spans
+- No span name collisions
+- Spans cannot be removed
+- Spans cannot "degrade" in stability (e.g., stable -> experimental)
```
lmolkova, line 260: I think this should be reversed.

E.g. I'm adding semconv for Azure messaging library based on stable generic semconv (they are not stable yet, but let's assume they are).

My customizations to generic messaging spans and metrics are experimental, core conventions are stable. Why is it not allowed?

However I should not be able to define stable conventions based on generic experimental ones - this would lead to breaking changes.

docs/specs/multi-registry/multi_registry.md
```diff
@@ -0,0 +1,366 @@
...
+### Events
+- No event name collisions
+- Events cannot be removed
+- Events cannot "degrade" in stability (e.g., stable -> experimental)
+- The set of required/recommended attributes must remain the same
+- Event names must match the following regular expression: `^[a-z][a-z0-9]*([._][a-z0-9]+)*$`
+
+### Spans
+- No span name collisions
+- Spans cannot be removed
```
lmolkova, line 259: why not? what does it even mean to remove a span?

I assume I pick the spans that I want to support by referencing them and I should be free to drop any spans that I don't need.

Moreover, if I'm defining semconv for a specific system (e.g. Kafka library), not all spans declared in generic conventions apply to my library, some of them are defined for scenarios that I don't support and it'd be best if I don't ever mention corresponding spans in my conventions.

docs/specs/multi-registry/multi_registry.md
```diff
@@ -0,0 +1,366 @@
...
+
+### Events
+- No event name collisions
+- Events cannot be removed
+- Events cannot "degrade" in stability (e.g., stable -> experimental)
+- The set of required/recommended attributes must remain the same
+- Event names must match the following regular expression: `^[a-z][a-z0-9]*([._][a-z0-9]+)*$`
+
+### Spans
+- No span name collisions
```
lmolkova, line 258: I think we need to find a better identity parameter for spans than name. Span names is an existing thing - they are usually something like "GET /users/{user-id}" or `SELECT public.my-shop-db` - they are human-friendly summary of what this span represents and are part of the exported telemetry.

docs/specs/multi-registry/multi_registry.md
```diff
@@ -0,0 +1,366 @@
...
+- Spans cannot be removed
+- Spans cannot "degrade" in stability (e.g., stable -> experimental)
+- The set of required/recommended attributes must remain the same
+- Span names must match the following regular expression: `^[a-z][a-z0-9]*([._][a-z0-9]+)*$`
```
lmolkova, line 262: see my comment above - span name is pre-existing thing with different rules (unicode string)

docs/specs/multi-registry/multi_registry.md
lquerel: I added the following item in the design principles list:
- **Transparency for Telemetry Consumers**: The downstream consumers of telemetry data, such as backends and dashboards, should never be exposed to conflicts within or between registries. They should see attributes and signals as they are defined, without any scope or conflict resolution directives, ensuring a consistent and reliable data experience.

docs/specs/multi-registry/multi_registry.md
```diff
@@ -0,0 +1,366 @@
...
+
+Open Questions:
+
+- Can we enforce the presence of the schema URL in the resource and instrumentation scope?
+- How do we convey the schema URL to the SDKs? Could that be part of the codegen done by Weaver?
+
+## Resolved Semantic Convention Registry Format
+
+> Note: A resolved registry is self-contained and does not include any complex constructs
+> like `imports`, `ref`, `extends`, etc. Their **structure is less subject to change**, making them good
```
lmolkova, line 315: if I'm defining a span or a metric, I have to reference some attribute and maybe extend some attribute group, so this may need clarification.

docs/specs/multi-registry/multi_registry.md
```diff
@@ -0,0 +1,366 @@
...
+> candidates for publication and easier to consume.
+
+The following properties are proposed for a resolved registry:
+
+- Resolved registry must be easy to consume and to publish
+  - Accessible via a URL.
+  - Self-contained, i.e. a single file.
+  - No `ref`, no `extends`, no `imports`, no alias, no other complex constructs.
+  - Yaml or JSON format so resolved registries can be easily consumed by any tool.
```
lmolkova, line 324: nitpicking: I think single file and this point are mutually exclusive - I get some warnings from VSCode on resolved attribute registry alone due to its size.

docs/specs/multi-registry/multi_registry.md
```diff
@@ -0,0 +1,366 @@
...
+More specifically, a resolved registry contains:
+- All the attributes registry specified locally in the registry.
+- All the groups specified locally in the registry.
+- All the attributes and groups imported but not re-exported locally are not included in the resolved registry. A
+  re-exported entity is an entity that is imported and referenced in the local registry with some overriding.
+
+Open Questions:
+
+- Do we keep track of the imported registries in the resolved registry? If yes, how? Lineage?
+- Can we leverage the attribute deduplication mechanism to simplify the merging of imported registries? ToDo -> Explain
```
lmolkova, line 348: can we start by disallowing attribute/span/event/metric duplication and only add it along with scoping only if there is actual demand?

docs/specs/multi-registry/multi_registry.md
```diff
@@ -0,0 +1,366 @@
...
+
+## Priorities
+
+TBD
+
+## General Open Questions
+
+- What about introducing a new type of semconv file that will let end-users define global overrides and global redact
+  directives? For example, the requirement level of attributes such as `client.address` or `server.address` will be
+  better defined by the end-user than by the library author, or a vendor. A similar approach could be used for redact
```
lmolkova, line 365: future concern: we should be crystal clear that user modifying requirement level in their registry has no effect on instrumentation libraries or produced telemetry.

docs/specs/multi-registry/registries/acme-http-server-lib/metric.yaml
```diff
@@ -0,0 +1,16 @@
+imports:
+  - schema_url: https://opentelemetry.io/schemas/1.27.0
+    name: otel
+
+groups:
+  - ref: otel:http.server.request.duration
```
lmolkova, line 6: I *think* that reusing the same name across different signals is allowed and is a feature rather than a bug (e.g. `dns.lookup.duration` is a metric name, but if DNS lookup is reported as an event, it'd make a perfect event name, or if it's reported as attribute on something, it could be an attribute name).  

So we need `ref:` to either reference a group id (and then give them some meaning) or reference a specific signal (`ref_metric`).

lmolkova (commented):
Thanks for the detailed spec!

I really like the direction it goes into, here are some suggestions:
- we need to define what it means to reference a group (metric/span/event) within semantic conventions - there are a lot of small questions I left on this, but I think this discussion needs to happen in semconv before anything is done in weaver to support it. We should start by supporting it within the same registry.
- The current scope of this proposal is huge and while I agree it's important to design and prototype how the eventual thing would look like, I'd suggest to start with attributes and attribute registry for the initial implementation once we have a basic idea how it'd work for groups.

docs/specs/multi-registry/multi_registry.md
lquerel: The approach was to support only the most essential/simple features at the beginning, with the option of adding such things later if the need arose. It seems that the need is already here :-)

Technically, I don’t think there’s a problem with supporting experimental definitions across registries. Even if we decide that this is supported by default, I believe we should, at a minimum, make it easier for users who don’t want to allow this. This could simply be implemented as an OTEL policy that prevents this type of dependency.

docs/specs/multi-registry/multi_registry.md
lquerel: Thanks

docs/specs/multi-registry/multi_registry.md
lquerel: I agree, that should be allowed, and I will make the change.

docs/specs/multi-registry/multi_registry.md
lquerel: In the OTEL registry, once a span has been defined, can its definition be removed in a later version? I thought this wasn't allowed if the span had a stability status set to `stable`. If that's the case, I think we should apply the same rule for all registries, whether OTEL or not.

docs/specs/multi-registry/multi_registry.md
lquerel: I agree. However, for symmetry with other signal types, we should ideally keep `name` to identify things, as we do with attributes and metrics, and use a separate identifier for the human-friendly summary of what this span represents.

docs/specs/multi-registry/multi_registry.md
lquerel: The next version of this document includes a complete revision of the resolution process and eliminates the need for scoping in the most common cases. I think the best approach at this stage is to postpone the discussion on scoping.

docs/specs/multi-registry/multi_registry.md
lquerel: Yes, in a general context. I will add this point to the document. However, I also believe that Weaver can enable the development of a new type of type-safe Client SDK, where a change in the `requirement_level` would have a direct impact on what is reported by the applications using this SDK. Personally, I am a strong believer in this approach.

docs/specs/multi-registry/registries/acme-http-server-lib/metric.yaml
lquerel: I think this is not something currently allowed or even supported. You’re mentioning in this comment something you’d like us to explore. Is that correct?

docs/specs/multi-registry/multi_registry.md
lquerel: Added your comment in the spec for discussion

docs/specs/multi-registry/multi_registry.md
lquerel: Reference experimental entities is now allowed by default.

docs/specs/multi-registry/multi_registry.md
lmolkova: Let's say I work on a library that supports multiple protocols (HTTP and something else). At some point my library removes support for one of these protocols. I should be allowed to stop referencing corresponding protocol spans - I no longer report them.

For metrics, for example, we at least support informal requirement levels and we could say that if you report e.g. http metrics, you must at least report at least required ones. Even this is tough to say - how do you know that they are going to report HTTP metrics if they don't reference them?

For spans it's even more complicated since span identity is not defined and spans don't have requirement levels. 
We don't define anything about spans in yaml beyond what's defined for the attribute_groups - we need to change this first.

----------

I think there are two separate discussions here:
1. What's allowed when I reference conventions from registry A in registry B (or within one registry). 
2. What changes are allowed in the the registry B from version N to N+1

I'm mostly talking about p1 (you can reference or not reference any group), and I think you're talking about p2. 

Maybe it'd be more clear if policies were separated into two groups: within-one-version and backward-compatibility?

docs/specs/multi-registry/multi_registry.md
lmolkova: Span name is used everywhere - in the spec - https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#span-creation, ion all semantic conventions. I don't think we can change how span name property is called - it'll be a huge source of confusion.

docs/specs/multi-registry/multi_registry.md
lmolkova: I love the idea, but it needs all instrumentation libraries to change. 

I.e. if my library reports `http.request.header.foo` as opt-in (or doesn't report it at all), it's not in the OTel SDK power to make my instrumentation report it.

docs/specs/multi-registry/registries/acme-http-server-lib/metric.yaml
lmolkova: I'm talking about current state.

I can easily define an attribute with the same name as metric, there are no checks and nothing going to stop me. And I consider it as a feature, not a bug.

lquerel: My bad. It's now fixed. See [5d68d7d](https://github.com/open-telemetry/weaver/pull/348/commits/5d68d7dfb79f88bb90c2318d8e760f69ade5b31c)

docs/specs/multi-registry/multi_registry.md
lquerel: I don’t think we need to change anything in the current instrumentation libraries. In my view, we can follow a two-step approach. The first step involves proposing a library that offers a type-safe API on top of the existing SDK clients. Those who prefer the current approach can continue using the current SDK client as they always have. Those who want the pure type-safe API approach will only interact with this additional layer. It is even conceivable that a user might want to adopt a mixed approach, using one or the other APIs depending on the need.

In step 2, the idea is to no longer base the type-safe API on the SDK clients, but to provide a fully optimized version by taking advantage of the fact that a type-safe API doesn’t need all the overhead of a generic approach. Hashmaps, abstraction layers, etc., can either be optimized or completely removed.

The advantage of this two-step approach is that we can test the type-safe API approach with the community at a lower cost. I already have a proof of concept for step 1 in Rust.

docs/specs/multi-registry/multi_registry.md
```diff
@@ -0,0 +1,493 @@
...
+
+> [!NOTE]
+> This proposal describes the overall changes required to achieve the multi-registry vision long term. However, these
+> changes can be introduced incrementally. For example, we could start by supporting attributes only and gradually add
+> support for metrics, events, spans, etc.
+
+> [!NOTE]
+> In this document, “semantic convention registry” refers to a collection of semantic convention entities (attributes,
+> groups, signals, etc.) that define the semantics of the data model used in OpenTelemetry. The terms “registry” and
+> “semantic convention registry” are used interchangeably. The term “entity” refers to any semantic convention
```
jsuereth, line 34: Nit: "Entity" may have overlap in name in OpenTelemetry. I can't think of a better term just yet, but we may need one.

docs/specs/multi-registry/multi_registry.md
```diff
@@ -0,0 +1,493 @@
...
+
+### `weaver_registry.yaml` File
+
+This file is used to describe the registry and its dependencies. It is optional for standalone registries
+but required for published registries or registries with dependencies on other registries.
+
+```yaml
+name: <registry_name>
+description: <registry_description>
+version: <registry_version>
```
jsuereth, line 152: One concern I have here is the same I have, e.g. with cargo crate versions.

What does CI/CD look like?  My ideal is releasing a version is as close to `git tag <version>` as possible.  

This is a bit of a nit, but also a foundational question - What does release and maintenance of these look like?  Should version be external, but everything else be in this file?

docs/specs/multi-registry/multi_registry.md
```diff
@@ -0,0 +1,493 @@
...
+Based on the `dependencies` section, Weaver can automatically update dependencies to the latest version when a new
+version becomes available (see the [Registry Update](#registry-update) section). It also allows registry authors to use
+registry names as preferred, without concerns about conflicts.
+
+Sections such as [Registry Resolution](#registry-resolution), [Registry Packaging](#registry-packaging), and 
+[Registry Publication](#registry-publication) provide more details on how the `weaver_registry.yaml` file
+is used in these processes.
+
+Open Questions:
+- Should we follow SemVer 2 for registry versions? It seems advisable, as Weaver can detect breaking changes. However,
```
jsuereth, line 183: This is something we likely need to discuss on the specification itself. cc @tigrannajaryan.

I do think we should just adopt semconv as it alleviates a lot of problems with a path forward for handling them, all while keeping the versioning "simple".

docs/specs/multi-registry/multi_registry.md
```diff
@@ -0,0 +1,493 @@
...
+reference by prefixing it with the registry name (e.g., `ref: otel:client.address`). The prefix corresponds
+to the registry name as defined in the `weaver_registry.yaml` file under the `dependencies` section. A colon is used
+as a separator between the name and the attribute or group name. References to groups or attributes defined in the
+local registry are never prefixed.
+
+In a group, the `extends` section can now reference an imported group using this new reference syntax. Similarly, the
+`attributes` section of a group can reference an imported attribute.
+
+In the current semantic conventions specification, referencing a group is currently unsupported. Uniqueness within
+groups is scoped by the type of group. It is entirely possible to have an event and a metric identified by the same ID.
```
jsuereth, line 210: Maybe we should fix this first -  This is something we could do BEFORE allowing multiple registries and I think would make our lives better overall.

cc @lmolkova for thoughts on that.

docs/specs/multi-registry/multi_registry.md
```diff
@@ -0,0 +1,493 @@
...
+However, when a conflict cannot be resolved consistently and predictably, the user must disambiguate by explicitly
+choosing the imported registry to resolve the conflict.
+
+The process hinges on the concept of mergeable entities (attributes or groups). For two entities to be considered
+mergeable, they must, at least, meet the following basic criteria:
+1. Both entities must be of the same type (e.g., attribute, metric, event, span, etc.).
+2. They must share the same ID.
+3. They must have a common ancestor (i.e., both were originally defined in the same registry).
+
+Since entities are complex objects consisting of multiple fields, and override mechanisms are allowed, one of the
```
jsuereth, line 249: I like this definition - but we should think about debuggability / o11y of these scenarios.  I may just need to read further, but how are you planning to call out / report these issues to users?

docs/specs/multi-registry/multi_registry.md
MadVikingGod: I don't think that we should force semconv on all repositories.  I think it's a good thing to follow for the otel repositories but might not be appropriate everywhere.

I think automatically upgrading to newer versions might be a bad idea in general.  If I pinned to a specific version getting a newer version will eventually lead to unexpected results.

docs/specs/multi-registry/multi_registry.md
lquerel: There are several elements throughout the document related to debuggability, but I consolidated the key points by adding two new paragraphs in the “Registry Resolution” section to summarize the ideas. See changes in [58d4f9d](https://github.com/open-telemetry/weaver/pull/348/commits/58d4f9d774b86b7be31157a2be80412d94731dd0)

docs/specs/multi-registry/multi_registry.md
jsuereth: Sorry, I meant semver

lquerel: @jsuereth I agree. We should use SemVer for OTel but not strictly adhere to all semantic versioning rules. We could simply define an order relation between versions (e.g., 1.2.0 > 1.1.15 > 0.9.32), which would be sufficient to determine if one registry is newer than another.

@MadVikingGod Regarding the automatic upgrade, to clarify, I’m not suggesting automatically upgrading registry versions. Instead, I’m proposing a command that allows the user to explicitly choose whether to automatically upgrade registry versions, so it’s done in a controlled manner.

docs/specs/multi-registry/multi_registry.md
lquerel: Here’s the English version of your text:

Using `git tag <version>` and not defining the version number in the YAML file is fine with me, as long as we can maintain the self-contained aspect of the artifact that will be downloadable by the users of the corresponding registry. This artifact could be hosted on a GitHub repo, a simple web server, or a CDN. In the end, Weaver should be able to retrieve the version either from the content, the file name, headers, etc. So, I feel like we’ll need a bit more than just a `git tag <version>`. Any suggestions to achieve something like this are welcome.

docs/specs/multi-registry/multi_registry.md
lquerel: I understand. Should I use `telemetry object` instead of entity?

docs/specs/multi-registry/multi_registry.md
lmolkova: @jsuereth I think it's a feature, not a bug - https://github.com/open-telemetry/weaver/pull/348#discussion_r1742858393

E.g. `http.request.body.size` can in theory be an attribute and a metric name. Or `messaging.message.time_in_queue`. 

I don't believe we have real examples of this but there were some discussions in [the past](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/32486#discussion_r1575584907) where it would make sense to use the same attribute name on an event/span as some existing metric name.

It could be a good feature for spans/events->metrics pipeline (take specific attribute and convert it to metric) and I'd like to check if we can keep this door open.

docs/specs/multi-registry/multi_registry.md
jmacd: +1 to starting with a simple rule (about required).

as a case in point, when the metric instrument is a monotonic counter there are very few problems associated with optional attributes: some systems will be OK, some will not. could an 3rd party library registry declare recommended attributes on a metric (say, with a configuration to enable/disable) and then an application registry pin down which attributes will or will not be used (e.g., by not enabling the choice in the 3rd party library)? that makes sense to me.

jmacd:
I don't feel qualified to approve this, but the motivation and vision look good to me. Thanks @lquerel!

docs/specs/multi-registry/multi_registry.md
jsuereth: Yes - I'm fine if `git tag <version>` kicks off a workflow that fills out this version, just want to make sure we keep it flexible so we can do that. I agree there needs to be a final file that has the version filled out.  I just want to make sure that's a "produced" artifact not an "input source" requiement.

jsuereth:
@lquerel Is it time to revisit this and get it merged?

lquerel:
> @lquerel Is it time to revisit this and get it merged?

@jsuereth Yes, it would be good to have some additional feedback.

jerbly:
I've read through this a few times now and this proposal seems sound to me. What do we need to get this proposal merged? Do we just need all the Open Questions resolved? Do we need to move any items out of scope?

Separately, Multi-Registry widens the user-base for semconv authoring and Weaver beyond those few who are very invested today. There is a steep learning curve not only to writing good quality definitions but all the periphery items like rego policies, minijinja templates, jq, etc. So, in parallel with this work I think we need to consider how to stabilize and simplify the authoring process. Perhaps we can make Weaver more "batteries-included" too so you get a standard set of document and code generation templates and validation policies out-of-the-box and so on.

docs/specs/multi-registry/multi_registry.md
lmolkova: >  It is entirely possible to have an event and a metric identified by the same ID.

I think it's no longer the case - we have unique group ids enforced across signals. So perhaps this discussion can be resolved?

docs/specs/multi-registry/registries/acme-http-server-lib/metric.yaml
lmolkova: I still think this one should be `otel:metric.http.server.request.duration`

https://github.com/open-telemetry/semantic-conventions/blob/06395f47d5ffc01d91ef6abe238fe56e941e7e9f/model/http/metrics.yaml#L39

so the `ref` is based on the group id (which is guaranteed to be unique) and not on the name (which could be reused across signals)

docs/specs/multi-registry/multi_registry.md
```diff
@@ -0,0 +1,501 @@
...
+```plaintext
+registry_root/
+  domain_1/
+  domain_2/
+  ...
+  resources/
+  weaver_registry.yaml
+```
+
+### `weaver_registry.yaml` File
```
lmolkova, line 144: I think we should decide if/how it aligns with the 3-file-schema-approach we discussed in semconv tooling call:

* 3 files  
  * Existing file \- marker  
    * contains pointer to diff  
    * contains pointer to def  
  * Model file  
     * Contains the full set of defined semantic conventions for that version  
  * Migration file  
     * Contains transformations you can use to keep version compatible \*when they are non breaking differences\*  
      * Should be cohesive/stand-alone for transformation use case.

There seem to be a lot of intersection with what'd we put into the 2.0 of `schema.yaml` and we should figure out if we need another one for registry manifest.

docs/specs/multi-registry/multi_registry.md
```diff
@@ -0,0 +1,501 @@
...
+
+In the current semantic conventions specification, referencing a group is currently unsupported. Uniqueness within
+groups is scoped by the type of group. It is entirely possible to have an event and a metric identified by the same ID.
+Therefore, the following approach is proposed to reference a group:
+
+```yaml
+groups:
+  - metric_ref: <reference_to_a_metric_group>
+  - event_ref: <reference_to_an_event_group>
+  - span_ref: <reference_to_a_span_group>
```
lmolkova, line 217: we enforce group id uniqueness within otel registry and have a policy to ensure group ids look like `metric.{metric_name}`, `event.{event_name}`, etc  so I think we'd be fine with `ref:`

docs/specs/multi-registry/multi_registry.md
```diff
@@ -0,0 +1,501 @@
...
+  - ...
+```
+
+Similar to attribute references, group references cannot override the type, stability, or deprecation status of the
```
lmolkova, line 221: I'd prefer to discuss it, maybe we can just say 

```suggestion
Similar to attribute references, group references would have limitations on what can be updated in scope of reference"
```

I can easily see that reference group is stable and something that overrides it is experimental - why not?

docs/specs/multi-registry/multi_registry.md
```diff
@@ -0,0 +1,501 @@
...
+imported group.
+
+References to experimental entities across registries are allowed under certain conditions:
+
+- An entity referencing an experimental entity must also be marked as experimental.
+- The flag `allow_experimental_ref` in the `weaver_registry.yaml` file must be set to `true`.
```
lmolkova, line 227: `allow_experimental_ref` - could we instead recommend (and include) a set of default policies to run when referencing things? I assume there will be tons of other policy-related things (charset, naming, deprecation/removal/back-compat, etc) that depend on the project needs and it'd be nice to make this configurable/expandable via a policy set rather than config options in the manifest

lmolkova:
I really like the direction it's going to! 

There are a lot of details to discuss/agree which are impl details and can be solved in one way or another.

I don't know if merging this specific PR is necessary though - if it is, I'd minimize it to cover the essential parts and would remove  implementation details that need discussion.

E.g. I think it'd be great to define 
- how the manifest file would look like https://github.com/open-telemetry/weaver/pull/348#discussion_r1922793699 both for  muilti- and single- registry
- how to reference group (both within one registry and across them)

But I feel like having smaller focused PRs would be easier.

docs/specs/multi-registry/multi_registry.md
lquerel: Replaced all references to entit(y|ies) with telemetry object(s).

docs/specs/multi-registry/multi_registry.md
lquerel: The last proposal doesn't have this limitation.

docs/specs/multi-registry/multi_registry.md
lquerel: Done.

docs/specs/multi-registry/multi_registry.md
lquerel: Made some clarifications in the proposal and removed these specific open questions.

docs/specs/multi-registry/multi_registry.md
lquerel: I have added the following note in the proposal.

> The way this version number for the registry is managed and injected into the manifest file is still up for
> discussion. For example, it could be part of a workflow triggered by a simple git tag <version> or any similar
> mechanism.

docs/specs/multi-registry/multi_registry.md
lquerel: @jsuereth I updated the proposal as follow.

The resolution process for supporting multiple registries does not rely specifically on semantic versioning SemVer 2 (we rely on the concept of mergeable telemetry objects sharing a common ancestor). However, it is entirely acceptable and supported for registry authors to adopt this convention if they wish to do so.

docs/specs/multi-registry/multi_registry.md
lquerel: @lmolkova I removed the part mentioning the possible lack of uniqueness between different types of signals. However, we still have an issue when a registry wants to extend the `attribute_group` of another registry. The `attribute_group` ID, I believe, is unique within a specific version of the registry. However, group IDs, in general, are not truly persistent across registry versions, unlike the `name` fields (used for the signals). I added a note in the proposal to keep track of this. Please let me know if you disagree.

docs/specs/multi-registry/multi_registry.md
lquerel: Yes I agree. I updated the proposal.

docs/specs/multi-registry/multi_registry.md
lquerel: Agree and applied.

lquerel:
@jsuereth @lmolkova @jerbly I’ve made a significant number of changes to the document to address most of your feedback. Regarding the open questions initially present in the document, I’ve either answered them by adding to the proposal or removed them because they were not necessarily important within the specific scope of this proposal.

Unless you see major issues to address in this new version, I’d like to get your approval and merge this proposal. This proposal has been sitting in the list of PRs for six months (mostly due to me). I believe it now provides a good representation of the direction we’re heading in terms of supporting multiple registries and the implications for the ecosystem. Having a merged version in the repo will also simplify communication. On a personal note, it would be nice to no longer see it in the list of PRs :-)

My next focus will be finalizing the PR on schema diffs to then begin the first round of implementation for this multi-registry support. What do you think?

jerbly (approved):
I think this should be merged now - there's more than enough here to start iterating. There's growing interest in this area with more people working out their own workarounds.

PR #357, opened 2024-09-06
Fix clippy issues.

jsuereth:
It looks like some new checks/improvements are made of this variety:

- Use `y.clone_into(x)` instead of `x = y.to_owned()`
- Don't borrow unless you need to.
- Use character literals instead of string literals of size 1


jsuereth:
Another quick note: I disabled some clone vs. `to_owned` checks in the comment filter, as we would need a mutable-borrow version of `trim` for no great reason.  We COULD create such a function, but I think the current performance hint is minimal for codegen and we can address that in the future.

lquerel (approved):

PR #365, opened 2024-09-12
feat: Add AnyValue proto support to semconv crate only

MSNev:


crates/weaver_semconv/src/any_value.rs
```diff
@@ -0,0 +1,336 @@
...
+use serde::{Deserialize, Serialize};
+
+use crate::attribute::Examples;
+use crate::stability::Stability;
+
+/// The AnyValueTypeSpec is a specification of a value that can be of any type.
+#[derive(Serialize, Deserialize, Debug, Clone, Eq, PartialEq, Hash, JsonSchema)]
+#[serde(rename_all = "snake_case")]
+#[serde(tag = "type")]
+pub enum AnyValueSpec {
```
MSNev, line 19: This structure is mostly unchanged from the file @lquerel provided back (added comments, and some additional tests).

Creating this Draft PR so that we can have an isolated discussion about the "AnyValue" support only.

I'm going to start merging this (and any updates we discuss) into #297

crates/weaver_semconv/data/expected/any_value.yaml
```diff
@@ -0,0 +1,72 @@
+body:
+  type: map
+  id: id
```
MSNev, line 3: The only "possible" issue with the current structure is that the `id` is not always required (for simple maps) at this location, it make sense at the later (nested) levels saying hey I have this field which is a map.

But here this is "body" specific. Assuming that we make the "body" just an `any_value` which would be the other way to change this -- except it make sense for `body` to be an any_value

MSNev: We could just make it "optional" (default to empty) and then "enforce" the presence as part of resolving the schema once it loaded...

MSNev: Actually, just started rolling this into my PR and with the `AnyValueSpec` this will completly replace all of the `BodySpec` stuff -- as `body` is now just the name of the `AnyValue` 😄

crates/weaver_semconv/src/any_value.rs
```diff
@@ -0,0 +1,336 @@
...
+    pub note: String,
+    /// Specifies the stability of the value.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub stability: Option<Stability>,
+    /// Sequence of examples for the value or single example
+    /// value. If only a single example is provided, it can
+    /// directly be reported without encapsulating it
+    /// into a sequence/dictionary.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub examples: Option<Examples>,
```
MSNev, line 132: I'm going to add the requirement_level to this on the "basis" that as `AnyValue` is effectively become what is called the `BodyField` in my current PR, it's "missing" these "Common" definitions in this version of the Spec

crates/weaver_semconv/Cargo.toml
```diff
@@ -16,6 +16,7 @@ weaver_common = { path = "../weaver_common" }
 
 serde.workspace = true
 serde_yaml.workspace = true
+serde_json.workspace = true
```
lquerel, line 19: If JSON is only used in unit tests, which I believe is the case, then you should move this dependency to the `[dev-dependencies]` section.

crates/weaver_semconv/src/any_value.rs
```diff
@@ -0,0 +1,336 @@
...
+                let entries = fields
+                    .iter()
+                    .map(|v| format!("{}", v))
+                    .collect::<Vec<String>>()
+                    .join(", ");
+                write!(f, "map{{{}}}", entries)
+            }
+            AnyValueSpec::Bytes { .. } => write!(f, "byte[]"),
+            AnyValueSpec::Undefined { .. } => write!(f, "undefined"),
+            _ => write!(f, "{:?}", self),
```
lquerel, line 149: In the example I provided, I left this part unimplemented, but I believe we should provide a dedicated rendering for each variant.

lquerel (changes requested):
In my opinion, there are two minor changes to make. Otherwise, it looks good to me.

crates/weaver_semconv/src/any_value.rs
MSNev: Yeah, I made this PR with mostly just your changes so that we could let everyone else see the changes, I've got several updates (currently just locally in my event PR) which provides all the "extra" stuff to support resolving.

MSNev: Still working on the test details -- I'd like to push the changes back into this PR by end of day -- I may (or may not) make that goal at this point

lquerel (approved):
LGTM, based on your last comment

crates/weaver_semconv/data/event.yaml
```diff
@@ -0,0 +1,181 @@
+groups:
```
MSNev, line 1: This file is not used in this PR, just including to show "why" I've added the Enum definition (even though we talked about excluding it).

Short Answer: is because the existing `device.app.lifecycle` has defined the usage of enum type in the `fields` 
BUT: the format is slightly different than currently defined for `AttributeType`

crates/weaver_semconv/data/event.yaml
```diff
@@ -0,0 +1,181 @@
...
+    type: event
+    name: device.app.lifecycle
+    brief: >
+      This event represents an occurrence of a lifecycle transition on Android or iOS platform.
+    note: >
+      This event identifies the fields that are common to all lifecycle events for android and iOS using
+      the `android.state` and `ios.state` fields. The `android.state` and `ios.state` attributes are
+      mutually exclusive.
+    body:
+      id: device_lifecycle
```
MSNev, line 13: The `id` is now (currently) required at this location (because we want the `fields` to have an `id`)

crates/weaver_semconv/data/event.yaml
```diff
@@ -0,0 +1,181 @@
...
+      type: map
+      requirement_level: optional
```
MSNev, line 15: added `requirement_level` and it is not required at each level (based on the same logic as all others)

crates/weaver_semconv/data/event.yaml
```diff
@@ -0,0 +1,181 @@
...
+        - id: ios.state
+          stability: experimental
+          requirement_level:
+            conditionally_required: if and only if `os.name` is `ios`
+          note: >
+            The iOS lifecycle states are defined in the [UIApplicationDelegate documentation](https://developer.apple.com/documentation/uikit/uiapplicationdelegate#1656902),
+            and from which the `OS terminology` column values are derived.
+          brief: >
+            This attribute represents the state the application has transitioned into at the occurrence of the event.
+          type: enum
```
MSNev, line 26: This is the `enum` example the main difference between an `any_value` enum type and the `AttributeType` enum type is

- you need to provide the "type" name (ie. `enum`)
- The `allow_custom_values` and `members` are now at the same "level" (not indented)

If this was the same as an AttributeType enum it would look like (which is a little problematic)
```yaml
brief: >
            This attribute represents the state the application has transitioned into at the occurrence of the event.
          type:
            allow_custom_values: false
            members:
              - id: active
                value: 'active'
                brief: >
                  The app has become `active`. Associated with UIKit notification `applicationDidBecomeActive`.
              - id: inactive
                value: 'inactive'
                brief: >
                  The app is now `inactive`. Associated with UIKit notification `applicationWillResignActive`.
             # ... Pruned ...
```

crates/weaver_semconv/src/any_value.rs
```diff
@@ -0,0 +1,690 @@
...
+
+    /// The value type is not specified.
+    Undefined {
+        /// The common value specification
+        #[serde(flatten)]
+        common: AnyValueCommonSpec
+    },
+
+    /// An enum definition type.
+    Enum {
```
MSNev, line 104: Add (even though we talked about) see additional comments in the `event.yaml` in the `data` folder.

crates/weaver_semconv/src/any_value.rs
```diff
@@ -0,0 +1,690 @@
...
+    /// directly be reported without encapsulating it
+    /// into a sequence/dictionary.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub examples: Option<Examples>,
+    /// Specifies if the field is mandatory. Can be "required",
+    /// "conditionally_required", "recommended" or "opt_in". When omitted,
+    /// the field is "recommended". When set to
+    /// "conditionally_required", the string provided as <condition> MUST
+    /// specify the conditions under which the field is required.
+    pub requirement_level: RequirementLevel,
```
MSNev, line 152: added this and `deprecated` as I have then in the Events PR

crates/weaver_semconv/src/any_value.rs
```diff
@@ -0,0 +1,690 @@
...
+impl Display for AnyValueSpec {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        match self {
+            AnyValueSpec::Map { fields, .. } => {
+                let entries = fields
+                    .iter()
+                    .map(|v| format!("{}", v))
+                    .collect::<Vec<String>>()
+                    .join(", ");
+                write!(f, "map<{}>{{ {} }}", self.id(), entries)
```
MSNev, line 170: Discussion point: 
Slight difference from the previous as I'm using the `id` to decorate the "name" of the `map` (and `enum`)

crates/weaver_semconv/src/any_value.rs
```diff
@@ -0,0 +1,690 @@
...
+            AnyValueSpec::Undefined { .. } => write!(f, "undefined"),
+            AnyValueSpec::Enum { .. } => write!(f, "enum<{}>", self.id()),
+        }
+    }
+}
+
+impl AnyValueSpec {
+    /// Returns the common value specification for each type.
+    #[must_use]
+    pub fn common(&self) -> &AnyValueCommonSpec {
```
MSNev, line 190: This is used to avoid having this selection "everywhere" especially when resolving the Spec (that part is not included in this PR -- still working on tests in the Events PR)

crates/weaver_semconv/src/any_value.rs
```diff
@@ -0,0 +1,690 @@
...
+
+        let expected_json = fs::read_to_string("data/expected/any_value.json").unwrap().replace("\r\n", "\n");
+        assert_eq!(expected_json, format!("{}", serde_json::to_string(&body).unwrap()), "{}", expected_json);
+
+        assert_eq!(
+            format!(
+                "{}",
+                map
+            ),
+            "map<id>{ enum<id_enum>, map<id_map>{ int, byte[], string, boolean, map<id_nested_map>{ int[], double[], string[], boolean[] } }, int, byte[], string, boolean, double, double[] }"
```
MSNev, line 483: Discussion Point:
This is the "example" of the Display from using the names above.
I went through a few iterations (one included the `id:type` for each "field" (but it made it very long), as the id is mandatory.

eg.
`map<id>{ enum<id_enum>, map<id_map>{ id_int:int, id_bytes:byte[], ...`

crates/weaver_semconv/src/group.rs
```diff
@@ -86,6 +86,10 @@ pub struct GroupSpec {
     pub name: Option<String>,
     /// The readable name for attribute groups used when generating registry tables.
     pub display_name: Option<String>,
+    // /// The event body definition
+    // /// Note: only valid if type is event
+    // #[serde(skip_serializing_if = "Option::is_none")]
+    // pub body: Option<AnyValueSpec>,   <-- This is from my Event PR (Just showing what the change is going to be here)
```
MSNev, line 92: As comment at the end of the line says, this is what it looks like (locally) in my Events PR, uncomment this and "add" the `event.yaml` file to the test and it will (at least it does in my Events PR) "load" this into the registry -- but as thats not the point of the PR I left it commented out (there are other changes I have locally -- not much) around validation.

crates/weaver_semconv/src/any_value.rs
MSNev: Ok, as part of "testing" this in the Events PR this "Display" appears to be used for generated registry `json` so this "can't" use this decoration.

MSNev: to further the discussion point, I've currently kept this Display and added a new `type_name()` to "populate" the resolved JSON with the "short" version of the names.

Will update this PR shortly with the changes.

crates/weaver_resolved_schema/src/any_value.rs
```diff
@@ -0,0 +1,71 @@
...
+pub struct AnyValue {
+    /// AnyValue name.
+    pub name: String,
+    /// Either a string literal denoting the type as a primitive or an
+    /// array type, a template type or an enum definition.
+    pub r#type: String,
+    /// A description of the type of the AnyValue
+    /// e.g. "string", "string[]", "int", "enum<enum_id>", "map<id>{ int, string }"
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub type_display: Option<String>,
```
jsuereth, line 24: How is this different than `name` in the enum case?

For non-map descriptions, why is this needed?

For map-descriptions, should this be derived from the structure?

I'd prefer not to add this, find it tedious and remove it.  What's the rationale for having it?

crates/weaver_resolved_schema/src/any_value.rs
```diff
@@ -0,0 +1,71 @@
...
+    /// specify the conditions under which the value is required.
+    pub requirement_level: RequirementLevel,
+    /// A more elaborate description of the any value.
+    /// It defaults to an empty string.
+    #[serde(skip_serializing_if = "String::is_empty")]
+    #[serde(default)]
+    pub note: String,
+    /// Specifies the stability of the any value.
+    /// Note that, if stability is missing but deprecated is present, it will
+    /// automatically set the stability to deprecated. If deprecated is
```
jsuereth, line 49: I'm not sure we physically can, but let's just make "deprecated" not allowed on stability for this new thing.  We want to remove it from everywhere else when we can.

crates/weaver_resolved_schema/src/any_value.rs
```diff
@@ -0,0 +1,71 @@
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub deprecated: Option<String>,
+    /// Identifies the definition of the "fields" of the value when the type is "map".
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub fields: Option<Vec<AnyValue>>,
+    /// Used when the type is "enum".
+    /// Set to false to not accept values other than the specified members.
+    /// It defaults to true.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub allow_custom_values: Option<bool>,
```
jsuereth, line 66: Let's not add this.  We effectively removed allow_custom_values form semconv, it's implicitly true everywhere now, and any setting of it is ignored.  I don't want to duplicate that mistake for events.

crates/weaver_resolved_schema/src/registry.rs
```diff
@@ -126,6 +126,10 @@ pub struct Group {
     /// The readable name for attribute groups used when generating registry tables.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub display_name: Option<String>,
+    // /// The body of the event.
+    // /// This fields is only used for event groups.
+    // #[serde(skip_serializing_if = "Option::is_none")]
+    // pub body: Option<AnyValue>,
```
jsuereth, line 132: Why is this commented out?

crates/weaver_semconv/data/event.yaml
jsuereth: Adding event makes sense to me, let's not add the deprecated "allow_custom_values" thing though

crates/weaver_semconv/data/event.yaml
jsuereth: What's the difference between "id" and "name"?

Let's be explicit in the comments which can be unique and how you reference something.

crates/weaver_semconv/data/event.yaml
jsuereth: Do you have a default in the model? (i.e. your requirement level can be set above you in the hierarchy and then SHOULD be excluded below?)  That would make sense to me but we can add in more checks over time.

jsuereth (commented):
Love the approach but have some nits on the structure.

Let's not add things that we're trying to remove elsewhere.

Let's not add things that we don't NEED yet (type descriptions), or can be derived from existing data.

crates/weaver_semconv/data/event.yaml
MSNev: Currently, the AnyValue is required to be present (as per the same behavior from Attributes). But we can make it optional and apply the same logic.

crates/weaver_resolved_schema/src/any_value.rs
MSNev: I agree, to the "Why is this needed". I accidently created this and it is an "interesting" short descriptive representation of the nested definition. And this is one of the major discussion topics about whether we do or do not want this (it's not really needed)

crates/weaver_resolved_schema/src/registry.rs
MSNev: In this PR as it's intent was to "just" show the AnyValue, this was showing "how" it is used in the Event PR (prior to me updating the Event PR)

crates/weaver_semconv/data/event.yaml
MSNev: Based on how weaver handles the `id` for everything else the `id` in the yaml becomes the `name` in the resolved definition.... Sooo. they are the same.

crates/weaver_resolved_schema/src/any_value.rs
MSNev: And actually, thinking about this is a different way, it should be possible to "process" this in any template implementation (if required) and re-create this exact same thing by parsing the structure (might need some minor "helpers" to do things like "get" the nested object or "has" nested objects. But we can grow those later (if required).

Really, I found this as a interesting representation more from a documentation representation (for something like markdown) perspective rather than 'just" the tables.

crates/weaver_semconv/data/event.yaml
jsuereth: gotcha - Sorry I missed that.

crates/weaver_resolved_schema/src/any_value.rs
jsuereth: I agree having a descriptive type is interesting, but I don't think it should be user-input (especially given what you're showing it as examples). I think we can synthesize something that will be consistent.  We can provide that synthesized string to templates via  filter or function, it should NOT be in the model.

Please remove, and we can figure out how to expose it in a follow up PR.

MSNev:
Event PR is now also updated with the changes on this PR and includes additional tests and validation.

crates/weaver_resolved_schema/src/any_value.rs
```diff
@@ -0,0 +1,57 @@
...
+
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+use weaver_semconv::attribute::{EnumEntriesSpec, Examples, RequirementLevel};
+use weaver_semconv::stability::Stability;
+
+/// An `AnyValue` definition.
+#[derive(Serialize, Deserialize, Debug, Clone, Eq, PartialEq, Hash, JsonSchema)]
+#[serde(deny_unknown_fields)]
+pub struct AnyValue {
```
lquerel, line 15: If any values remain unchanged during the resolution process, why do we need to redefine `AnyValueSpec` in the `semconv` crate? I recommend removing this struct and using `AnyValueSpec` in the resolved registry instead. If, at some point, we have elements in any value that become resolvable, we can reintroduce this struct without breaking the contract with the jq and Jinja2 engines, as they only interact with the JSON/YAML external representation of the resolved registry.

crates/weaver_resolved_schema/src/registry.rs
lquerel, line 132: This is where we could use AnyValueSpec.

crates/weaver_resolver/src/any_value.rs
```diff
@@ -0,0 +1,44 @@
+// SPDX-License-Identifier: Apache-2.0
+
+//! Functions to resolve a semantic convention body.
+
+use weaver_resolved_schema::any_value::AnyValue;
+use weaver_semconv::{any_value::AnyValueSpec, attribute::EnumEntriesSpec};
+
+/// Resolve a `Body` specification into a resolved `Body`.
+#[must_use]
+pub fn resolve_any_value_spec(value: &AnyValueSpec) -> AnyValue {
```
lquerel, line 10: This resolution looks like a deep copy. See my previous comments, I don't think we need this code.

crates/weaver_resolved_schema/src/registry.rs
MSNev: Correct

crates/weaver_resolver/src/any_value.rs
MSNev: I'll update, as I was just following the existing pattern (and it was previously needed for the body), but I can rework -- the only possible minor issue will be the "r#type" field of the AnyValue which doesn't explicitly exist (as a separate field -- as it's derived from the enum structure and the `Display`), but I'm sure there is some way around that.

MSNev: And yes, this is effective just a deep conversion / copy as it is now.

crates/weaver_resolved_schema/src/any_value.rs
MSNev: @lquerel, @jsuereth
Ok, I started doing this locally and I've hit the following
- `type` is mostly unaffected, the only difference is for the `enum` type as the resolved version (to be the same as the existing enum) includes the id's of the enums (i.e. `enum {active, inactive, background, foreground, terminate}`) where now it becomes just `enum` (probably not a bad thing, and maybe be the preferred approach anyway)
- Bigger (potential) issue, because we use `id` in the semconv and `name` in the resolved, by removing the resolved `AnyValue` it also becomes `id` -- Question is this what we want or do we want the resolved `AnyValue` to contain `name` instead of `id`?

MSNev: Advantages
- Same as `enum` member(s) are they are already just `id` and not `name`
- if we want to introduce a `name` to describe the `AnyValueSpec` later to describe the "type" of a `map` etc we could.
- Has more of a separation from the "groups" vs "values" (groups have both "id" and "name" where the name is really just the "id". So the "body" (or any location of an `AnyValue` will just have its original `id`

MSNev: Ok, I've updated this PR with the changes to "remove" the resolved and resolver `any_value.rs` files and usage, I'll wait for the answer to the above before publishing the Event PR with the change (I already have it locally and it's just waiting). I'll add some comments and link here to the "id" / "name" change for the test files for some context.

lquerel: Since we can still introduce names in the body fields later, I’ll go with the simpler version to minimize code duplication.

MSNev:
I think this PR has achieved it's goal of letting us come to a simplified and agreed upon definition of the AnyValue -- so closing as the Event PR #297 will include all of the changes from this PR.

PR #379, opened 2024-09-19
Add `escape_square_brackets` into `comment_formats` markdown configuration

### LINKED ISSUE #374: Add a flag in `comment_formats` to escape invalid markdown links
Some `brief` and `note` fields contain non-escaped square brackets that don't represent valid Markdown links, such as ranges like `[0,n]`. `cargo doc` rejects any links that don't reference internal definitions, even though most markdown linters tolerate them. The proposal is to add a new parameter, `escape_square_brackets`, in the markdown configuration to automatically escape such invalid links.

lquerel:
The `cargo doc` command does not properly handle non-escaped text containing square brackets when these brackets are not intended to represent links (e.g., `This is a [text] with square brackets that doesn't represent a link`). This PR addresses the issue by introducing a new flag `escape_square_brackets` in the markdown configuration of the `comment_formats` section.

This new flag ensures that text with square brackets, which are not intended as links, is properly escaped in the generated documentation.

### Input Text:
```text
In some cases, a [URL] may refer to an [IP](http://ip.com) and/or port directly.
The file \[extension\] is extracted \[from] the `url.full`, excluding the leading dot.
```

### Output Without the Flag:
Without the new flag, the following text is returned by the comment filter:
```text
In some cases, a [URL] may refer to an [IP](http://ip.com) and/or port directly.
The file \[extension\] is extracted \[from] the `url.full`, excluding the leading dot.
```

### Output With the Flag Enabled:
When the new flag is enabled, the output becomes:
```text
In some cases, a \[URL\] may refer to an [IP](http://ip.com) and/or port directly.
The file \[extension\] is extracted \[from\] the `url.full`, excluding the leading dot.
```

Closes #374

### DIFF
--- a/crates/weaver_forge/README.md
+++ b/crates/weaver_forge/README.md
@@ -579,6 +579,7 @@ comment_formats:           # optional
 
     # Fields specific to 'markdown' format
     escape_backslashes: <bool>            # Whether to escape backslashes in markdown (default: false).
+    escape_square_brackets: <bool>        # Whether to escape square brackets in markdown (default: false).
     shortcut_reference_links: <bool>      # Convert inlined links into shortcut reference links (default: false).
     indent_first_level_list_items: <bool> # Indent the first level of list items in markdown (default: false).
     default_block_code_language: <string> # Default language for block code snippets (default: "").
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 All notable changes to this project will be documented in this file.
 
+## [Next] - YYYY-MM-DD
+
+What's changed
+
+* Add `escape_square_brackets` into `comment_formats` markdown configuration. ([#XXX](...) by lquerel).
+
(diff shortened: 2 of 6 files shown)


jsuereth (approved):

crates/weaver_forge/README.md
```diff
@@ -579,6 +579,7 @@ comment_formats:           # optional
 
     # Fields specific to 'markdown' format
     escape_backslashes: <bool>            # Whether to escape backslashes in markdown (default: false).
+    escape_square_brackets: <bool>        # Whether to escape square brackets in markdown (default: false).
```
joaopgrassi, line 582: Since we have brackets already in the conventions, what is happening there today? And then, shouldn't this be `true` by default, as we want to keep the brackets to match the source data? 🤔

lquerel: I agree. It’s probably better to have this escaping enabled by default. I will make this change before merging. Thanks

PR #382, opened 2024-09-20
Check group id, metric name, and group name uniqueness

### LINKED ISSUE #306: Check group id, metric name, and event name uniqueness
We need to check that metric/events names are unique similarly to https://github.com/open-telemetry/semantic-conventions/pull/1302

Also that group-ids are unique

Blocking PRs in semconv:
- https://github.com/open-telemetry/semantic-conventions/pull/1423
- https://github.com/open-telemetry/semantic-conventions/pull/1440

lquerel:
This PR detects duplications in the following semconv concepts:
- Group ID
- Metric name
- Group name (e.g., resource and event names)

Since there are existing duplications in the current registries (see the example below), we return warnings instead of errors. As usual, the `--future` flag must be used for future versions of the registry.

Example: https://github.com/search?q=repo%3Aopen-telemetry%2Fsemantic-conventions+path%3A%2F%5Emodel%5C%2F%2F+feature_flag&type=code

For now, the group name duplication detection is disabled.

Closes #306


crates/weaver_resolver/src/registry.rs
```diff
@@ -203,6 +238,39 @@ fn check_group_any_of_constraints(
...
+        if let Some(key) = key_fn(group) {
+            let provenances = keys.entry(key).or_default();
+            provenances.push(group.provenance().to_owned());
+        }
+    }
+
+    for (key, provenances) in keys {
+        if provenances.len() > 1 {
+            // Deduplicate the provenances.
+            let provenances: HashSet<String> = provenances.into_iter().collect();
```
jsuereth, line 267: Nit: You can use itertools to `.unique` this when collector instead of forcing a hashset.

I think performance wise it's the same, just easier to read.

jsuereth (approved):

Issue #267, opened 2024-07-29
Remove all remote dependencies on the semconv repo (i.e. Git URL or Git Archive URL) for stability purposes.


lquerel:
We should move away from relying on remote APIs like git url on the semconv repo and instead stand up a local server that hosts the file, for stability purposes.

See: `crates/weaver_cache/src/lib.rs` test `test_semconv_registry_git_repo`. Other instances of that exist in the repo.

PR #388, opened 2024-09-24
Mini http server to serve static files in the context of unit tests

lquerel:
This PR implements a simple HTTP server to serve static files, so tests no longer depend on remote files.

Example:

```
#[test]
fn test_semconv_spec_from_url() {
    let server = ServeStaticFiles::from("tests/test_data").unwrap();
    let semconv_url = server.relative_path_to_url("url/common.yaml");
    let result =
        SemConvRegistry::semconv_spec_from_url(&semconv_url).into_result_failing_non_fatal();
    assert!(result.is_ok());
}
```

In this example, the mini HTTP server returns the file located at `tests/test_data/url/common.yaml`. The URL returned by `server.relative_path_to_url` is something like `http://127.0.0.1:<dyn_port>/url/common.yaml`.

Closes: https://github.com/open-telemetry/weaver/issues/267


jsuereth:
I think the windows failure here is legit, and looks like it might be a leaky thread of sorts.

jsuereth:
Closing this as replaced by #429

PR #392, opened 2024-09-30
Allow adding a description when using `opt_in` requirement level

joaopgrassi:
I bumped into this while working on https://github.com/open-telemetry/semantic-conventions/issues/474. I want to change a requirement level to `opt_in` and add a description to it, but I can't because the schema doesn't allow it.


lmolkova (approved):
Schema changes look great to me. Thanks for doing this!

Issue #186, opened 2024-05-29
Automate OTEL Schema Generation and Update Process with Migration Guide Support


lquerel:
To properly support multiple registries with Weaver, we need to automatically generate the OTEL schema for a new version of a registry before packaging and publishing it. This issue focuses on defining the necessary steps to update the OTEL schema in an automated and controlled manner, based on changes described in the SemConv files.

Tasks:
- [x] Identify the best approach to track renamings and deletions in SemConv. See the discussion [here](https://github.com/open-telemetry/semantic-conventions/issues/1426).
- [x] Specify the format and structure of the file used to track the SemConv registry name, version, schema location, and other details. See the related discussion below.
- [ ] Implement `otel-registry-manifest.yaml` and the new format for the [deprecated](https://github.com/open-telemetry/semantic-conventions/issues/1426) field.
- [ ] Create a sub-command to generate a new version of the telemetry schema.
- [ ] Create a sub-command and/or template to generate the migration guide for a specific version. See the discussion on the [migration guide](https://github.com/open-telemetry/semantic-conventions/blob/5c02e73b2f551ce6ccc2c90ef5547fb09c238092/docs/database/migration-guide.md#database-client-span-attributes).

A new file, `otel-registry-manifest.yaml`, located at the root of the models directory, will be used by Weaver to generate the next version of the OTEL schema and update the migration guides. The file will have the following structure:

EDIT: The following file structure has been revised based on the SemConv Tool SIG discussion.

```yaml
# File otel-registry-manifest.yaml
name: <registry_name>                              # e.g., otel
description: <registry_description>
semconv_version: <registry_version>                # e.g., 1.27.0
schema_base_url: <url>                             # e.g., https://opentelemetry.io/schemas/

# This file will be extended in the future to support multiple registries.
# Fields such as `registry_base_url`, `owners`, `maintainers`, `dependencies`, and more will be added later.
```

lmolkova:
```yaml
name: <registry_name>                      # e.g., otel
description: <registry_description>
semconv_version: <registry_version>                # The version number of the current registry.
url: <registry url> # Uniquely identifies a registry source and where schema with `version` can be downloaded from, it has all the versions ever published
```

PR #400, opened 2024-10-03
Registry manifest and Schema diff

lquerel:
Note: The scope of this PR has been reduced to focus only focus on the schema diff feature. Github issues have been created to track the features that have been postponed #482, #483. 

This PR implements the command `registry diff`, see the following example:
```
cargo run -- registry diff -r https://github.com/open-telemetry/semantic-conventions/archive/refs/tags/v1.27.0.zip[model] --baseline-registry https://github.com/open-telemetry/semantic-conventions/archive/refs/tags/v1.26.0.zip[model] --diff-format markdown
```

In this example, the diff is displayed in markdown format. The following formats are supported: json, markdown, ansi, ansi_stats. YAML format will be supported once PR #525 is finalized.
 
A detailed description of the schema diff data model and the diffing process is visible [here](https://github.com/open-telemetry/weaver/blob/4d938f1214ac633258d69bf04339d9ff8024cd06/docs/schema-changes.md).

Notes:
- The crate `weaver_otel_schema` is not essential for this PR; it was initially included as part of the preparations for the `registry schema-update` command. We have decided to implement this command in a future PR. However, for simplicity, I prefer to keep the preparation code in place instead of removing it. Same thing for `all_changes` in `weaver_version`.

**List of modifications to apply to the semantic conventions repository after the release of the Weaver containing the current PR:**
- Add a `registry-manifest.yaml` file with the version of the next release.
- Update all deprecated fields.

Closes: #186 
----

The following command comparing the versions 1.29 and 1.30

```
/weaver registry diff -r 'https://github.com/open-telemetry/semantic-conventions/archive/refs/tags/v1.30.0.zip[model]' --baseline-registry 'https://github.com/open-telemetry/semantic-conventions/archive/refs/tags/v1.29.0.zip[model]' --diff-format markdown
```

produces the following markdown output:

#### New registry_attributes:
- Add aws.extended_request_id
- Add azure.client.id
- Add azure.cosmosdb.connection.mode
- Add azure.cosmosdb.consistency.level
- Add azure.cosmosdb.operation.contacted_regions
- Add azure.cosmosdb.operation.request_charge
- Add azure.cosmosdb.request.body.size
- Add azure.cosmosdb.response.sub_status_code
- Add cassandra.consistency.level
- Add cassandra.coordinator.dc
- Add cassandra.coordinator.id
- Add cassandra.page.size
- Add cassandra.query.idempotent
- Add cassandra.speculative_execution.count
- Add cicd.pipeline.result
- Add cicd.pipeline.run.state
- Add cicd.system.component
- Add cicd.worker.state
- Add code.column.number
- Add code.file.path
- Add code.function.name
- Add code.line.number
- Add db.system.name
- Add elasticsearch.node.name
- Add gen_ai.request.seed
- Add k8s.namespace.phase
- Add network.connection.state
- Add security_rule.category
- Add security_rule.description
- Add security_rule.license
- Add security_rule.name
- Add security_rule.reference
- Add security_rule.ruleset.name
- Add security_rule.uuid
- Add security_rule.version
- Add vcs.repository.name

#### Deprecated registry_attributes:
- code.column (Note: Deprecated, use `code.column.number`)
- code.function (Note: Deprecated, use `code.function.name` instead)
- code.lineno (Note: Deprecated, use `code.line.number` instead)
- db.cassandra.consistency_level (Note: Deprecated, use `cassandra.consistency.level` instead.)
- db.cassandra.coordinator.dc (Note: Deprecated, use `cassandra.coordinator.dc` instead.)
- db.cassandra.coordinator.id (Note: Deprecated, use `cassandra.coordinator.id` instead.)
- db.cassandra.idempotence (Note: Deprecated, use `cassandra.query.idempotent` instead.)
- db.cassandra.page_size (Note: Deprecated, use `cassandra.page.size` instead.)
- db.cassandra.speculative_execution_count (Note: Deprecated, use `cassandra.speculative_execution.count` instead.)
- db.cosmosdb.client_id (Note: Deprecated, use `azure.client.id` instead.)
- db.cosmosdb.connection_mode (Note: Deprecated, use `azure.cosmosdb.connection.mode` instead.)
- db.cosmosdb.consistency_level (Note: Deprecated, use `cosmosdb.consistency.level` instead.)
- db.cosmosdb.regions_contacted (Note: Deprecated, use `azure.cosmosdb.operation.contacted_regions` instead.)
- db.cosmosdb.request_charge (Note: Deprecated, use `azure.cosmosdb.operation.request_charge` instead.)
- db.cosmosdb.request_content_length (Note: Deprecated, use `azure.cosmosdb.request.body.size` instead.)
- db.cosmosdb.sub_status_code (Note: Deprecated, use `azure.cosmosdb.response.sub_status_code` instead.)
- db.elasticsearch.node.name (Note: Deprecated, use `elasticsearch.node.name` instead.)
- db.elasticsearch.path_parts (Note: Deprecated, use `db.operation.parameter` instead.)
- db.system (Note: Deprecated, use `db.system.name` instead.)
- event.name (Note: Identifies the class / type of event.)
- exception.escaped (Note: Indicates that the exception is escaping the scope of the span.)
- gen_ai.openai.request.seed (Note: Deprecated, use `gen_ai.request.seed`.)
- system.network.state (Note: Deprecated, use `network.connection.state` instead.)

#### New metrics:
- Add metric.azure.cosmosdb.client.active_instance.count
- Add metric.azure.cosmosdb.client.operation.request_charge
- Add metric.cicd.pipeline.run.active
- Add metric.cicd.pipeline.run.duration
- Add metric.cicd.pipeline.run.errors
- Add metric.cicd.system.errors
- Add metric.cicd.worker.count
- Add metric.k8s.cronjob.active_jobs
- Add metric.k8s.daemonset.current_scheduled_nodes
- Add metric.k8s.daemonset.desired_scheduled_nodes
- Add metric.k8s.daemonset.misscheduled_nodes
- Add metric.k8s.daemonset.ready_nodes
- Add metric.k8s.deployment.available_pods
- Add metric.k8s.deployment.desired_pods
- Add metric.k8s.hpa.current_pods
- Add metric.k8s.hpa.desired_pods
- Add metric.k8s.hpa.max_pods
- Add metric.k8s.hpa.min_pods
- Add metric.k8s.job.active_pods
- Add metric.k8s.job.desired_successful_pods
- Add metric.k8s.job.failed_pods
- Add metric.k8s.job.max_parallel_pods
- Add metric.k8s.job.successful_pods
- Add metric.k8s.namespace.phase
- Add metric.k8s.replicaset.available_pods
- Add metric.k8s.replicaset.desired_pods
- Add metric.k8s.replication_controller.available_pods
- Add metric.k8s.replication_controller.desired_pods
- Add metric.k8s.statefulset.current_pods
- Add metric.k8s.statefulset.desired_pods
- Add metric.k8s.statefulset.ready_pods
- Add metric.k8s.statefulset.updated_pods
- Add metric.vcs.change.time_to_merge

#### Deprecated metrics:
- metric.db.client.cosmosdb.active_instance.count (Note: Deprecated)
- metric.db.client.cosmosdb.operation.request_charge (Note: Deprecated)

#### New spans:
- Add span.azure.cosmosdb.client


crates/weaver_forge/templates/test/resource.md
```diff
@@ -39,7 +39,7 @@ Brief: {{ resource.brief }}
 - Sampling relevant: {{ attribute.sampling_relevant }}
   {%- endif %}
   {%- if attribute.deprecated %}
-- Deprecated: {{ attribute.deprecated }}
+- Deprecated: {{ attribute.deprecated.note }}
```
jsuereth, line 42: Is/will this be a breaking change?

crates/weaver_version/src/metrics_changes.rs
```diff
@@ -1,11 +1,19 @@
 // SPDX-License-Identifier: Apache-2.0
 
-//! Metrics change definitions.
+//! Metrics version.
```
jsuereth, line 3: nit: I think "change definitions" was better here...

crates/weaver_semconv/src/deprecated.rs
```diff
@@ -0,0 +1,341 @@
...
+
+        // Handle the old format (just a string)
+        fn visit_str<E>(self, value: &str) -> Result<Self::Value, E>
+        where
+            E: de::Error,
+        {
+            // Regex to match "Replaced by `some_field`"
+            // ToDo make this regex global
+            // Replaced by `([\w]+(?:\.[\w]+)*)`?
+            let renamed_regex =
```
jsuereth, line 66: Do we want to keep this or just go update semconv to use structured data?

I'd be concerned we can't remove this once added.

crates/weaver_resolved_schema/src/lib.rs
```diff
@@ -64,44 +70,367 @@ pub struct ResolvedTelemetrySchema {
...
+        }
+
+        if let Some(ref manifest) = baseline_schema.registry_manifest {
+            changes.set_baseline_manifest(weaver_version::schema_changes::RegistryManifest {
+                semconv_version: manifest.semconv_version.clone(),
+            });
+        }
+
+        // Attributes in the registry
+        self.diff_attributes(baseline_schema, &mut changes);
```
jsuereth, line 162: I have two concerns here:

1. overall attribute renames rely on the *registry concept*.  We can only look at renames/added/removed from attribute groups there.  We should probably avoid looking at deprecated on a per-signal basis *in that instance*.
2. On a per-signal basis, we should be tracking added/removed attributes. (We can ignore changed, as those should be caught in overall "registry" renames.


Example:

- There are two attributes `error.type` and `http.status_code`.
- There is an event `http.my.count`.  It currently uses `error.type`.
- Example1: `http.my.count` *event* is updated to use `http.status_code` instead, using some kind of "ref"/"deprecated" combo.
- Example2: `http.my.count` updates `error.type` from `opt-in` to `recommended`.

Perhaps we don't want to handle this yet.  Curious on @lmolkova's thoughts.

crates/weaver_forge/templates/test/resource.md
lquerel: Yes, it is a breaking change at the rendering level because, without the `.note`, Jinja renders the new JSON of the `deprecated` field instead of the text that now corresponds to the `note`. I couldn’t find another way to handle this at the rendering level. However, at the semconv parsing level, it is not a breaking change since both formats remain supported.

crates/weaver_version/src/metrics_changes.rs
lquerel: I don’t recall any strong reasons for making this change. I’ll revert to the previous comment.

crates/weaver_semconv/src/deprecated.rs
lquerel: Currently, Weaver supports semconv registries from version 1.22 and above. For the `registry diff` command, I used a best-effort approach to support the `deprecated` field, but we could consider not doing so and instead fallback to `Deprecated::Deprecated` when the deserializer detects the old format. This would mean that meaningful diffs won’t be possible until we have two consecutive versions of the registry using the new format for `deprecated`.

I’m not opposed to removing this part of the code if you’re not comfortable with it. Please confirm.

crates/weaver_resolved_schema/src/lib.rs
```diff
@@ -64,44 +70,367 @@ pub struct ResolvedTelemetrySchema {
...
+                        .catalog
+                        .attribute(attr_ref)
+                        .expect("Attribute ref not found in catalog. This is a bug.");
+                    (attr.name.as_str(), attr)
+                })
+            })
+            .collect()
+    }
+
+    /// Get the metrics of the resolved telemetry schema.
```
lquerel, line 124: [LQ ToDo] Fix this comment.

crates/weaver_resolved_schema/src/lib.rs
```diff
@@ -64,44 +70,367 @@ pub struct ResolvedTelemetrySchema {
...
+    #[must_use]
+    pub fn groups(&self, group_type: GroupType) -> HashMap<String, &Group> {
+        self.registry
+            .groups
+            .iter()
+            .filter(|group| group.r#type == group_type)
+            .map(|group| match group_type {
+                GroupType::AttributeGroup
+                | GroupType::Event
+                | GroupType::Span
```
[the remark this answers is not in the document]
lquerel: Good question, we still have this thorn in our side with spans. How do you see things from the perspective of semantic conventions?

crates/weaver_resolved_schema/src/lib.rs
lquerel: Concerns 1 and 2 make sense to me. I can remove rename tracking for everything except the attributes in the registry.

Regarding the examples, could you describe more precisely what you’d like to see in the diff?

crates/weaver_resolved_schema/src/lib.rs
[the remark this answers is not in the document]
lmolkova: we can start relying on group.id across all signals now - https://github.com/open-telemetry/semantic-conventions/pull/1512 

I'll keep making a slow progress on spans to improve it further - https://github.com/open-telemetry/semantic-conventions/issues/1513, but it should not be a blocker

crates/weaver_semconv/src/deprecated.rs
```diff
@@ -0,0 +1,341 @@
...
+- deprecated: 
+    action: deprecated
+    note: This attribute is no longer used.
+- deprecated: 
+    action: deprecated
+    note: Should no longer be used.
+- deprecated:
+    action: renamed
+    new_name: foo.unique_id
+    note: This field has been renamed for consistency.
```
lmolkova, line 217: I thought we're going ahead with https://github.com/open-telemetry/semantic-conventions/issues/1426#issuecomment-2388775414

```yaml
- id: foo.id
  deprecated:
    action: renamed
    renamed_to: foo.unique_id
  note: we can write anything here
- id: foo.another_attribute
  deprecated: 
    action: removed
  note: another optional note
```

I.e. note is the attribute property.

This seems to be minimal and sufficient (we tend to remove notes from deprecated attributes, so it's totally fine to rewrite the original note and provide any deprecation info there).  I can see how `deprecated.note` could be nicer, but I'm suggesting to avoid adding it at least for now.

crates/weaver_forge/templates/test/resource.md
lmolkova: Related comment: https://github.com/open-telemetry/weaver/pull/400/files#r1868713601 ( I thought we'd only keep the `note` on the attribute without introducing one on `deprecated`).

But if we do have `deprecated.note`, it'd still optional thing that does not need to describe the `action`.

I think codegen would need to generate something like 

```java
/*
 * ...
 * @deprecated renamed to {@link
 *     io.opentelemetry.semconv.FooAttributes#FOO_BAR}.
 *  Here's an additional note.
 */
```

from 

```yaml
deprecated: 
  action: renamed
  new_name: foo.bar
note: Here's an additional note.
```

so maybe we should eventually provide some helper filters to render those.

In the scope of this PR, if someone uses `{{ attr.deprecated }}`, I wonder if jinja can call `to_string` on the original `deprecated` object to return a custom string? We'd format it similarly to "Renamed to `foo.bar`", but someone who wants to use new structured info would be able to obtain it with `attr.deprecated.action`, etc

WDYT?

crates/weaver_resolved_schema/src/lib.rs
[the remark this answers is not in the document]
lquerel: Okay, now that we have a form of uniqueness for group IDs within the same registry, I’m fine with it. We’ll need to scope these IDs by registry once we add multi-registry support.

crates/weaver_semconv/src/deprecated.rs
lquerel: I forgot this discussion... I'm fine with it (i.e. reusing the original note).

crates/weaver_resolved_schema/src/lib.rs
lmolkova: I think this PR detects renames only (for attributes, metrics, etc) and does not analyze what's changed within a group. 

I can see cases when we want to go deeper, but they are beyond changelog/schema transformation automation. E.g. we do it in compatibility policy checks.

I wonder if good old weaver.yaml + JQ + jinja with access to baseline and new semconv at the same time can cover all possible complex diff scenarios one would want to do.

crates/weaver_semconv/src/deprecated.rs
jsuereth: Barring comments from other @open-telemetry/specs-semconv-maintainers , I'm more comfortable keeping the weaver behavior here simple, and requiring explicit adoption of the new deprecation yaml for diffs to start working.

However if the ergonomics without this is quite bad, we can keep it, perhaps with some kind of flag we can disable it when in `--future` mode

crates/weaver_resolved_schema/src/lib.rs
jsuereth: I think at some point we'll want diffs on what to include in a metric or span.  As Lidumila calls out we can defer that until later .

crates/weaver_semconv/src/deprecated.rs
lquerel: I made this change. Now, the notes taken into account are those at the level of the object containing the deprecated field.

However, I renamed the `removed` action to `deprecated` because, in the diff, there is a distinction between something marked as `deprecated` and something that has disappeared from the registry. An attribute that has disappeared is marked as `removed` in the diff.

```
- id: foo.id
  deprecated:
    action: renamed
    renamed_to: foo.unique_id
  note: we can write anything here
- id: foo.another_attribute
  deprecated: 
    action: deprecated                 <-- was removed in your example
  note: another optional note
```

crates/weaver_resolved_schema/src/lib.rs
lquerel: Fixed

crates/weaver_semconv/src/deprecated.rs
lquerel: I removed the complex inference logic. To fully benefit from the `registry diff` command, we will need to migrate the registry to use the new `deprecated` format. That does seem more reasonable.

crates/weaver_version/src/metrics_changes.rs
lquerel: Fixed

lquerel:
@lmolkova @jsuereth 

Note: I have addressed most of the feedback. The main task remaining is the removal of change detection for elements other than attributes. That will be done soon.

I have a question regarding the format of the new `deprecated` field. In the current version of this PR, a `deprecated` field can take one of the following three forms:

Old approach (still supported for compatibility reasons):
```
deprecated: "deprecation message"
```

or

```
deprecated:
  action: renamed
  renamed_to: attribute_name
```

or

```
deprecated:
  action: deprecated
```

With this, we can handle simple attribute renaming scenarios, as well as merge scenarios (e.g., A and B are renamed to C; Weaver will detect this automatically). However, we currently have no way to represent a split (e.g., A is renamed to B and C). So with the current implementation, the semconv author will need to set `deprecated` to `action: deprecated` and provide a `note` at the object level to explain the split in textual form.

We could make this explicit in the format of the deprecated field and in the diff output. This would allow for migration documentation that more accurately reflects the desired changes. However, it still wouldn’t enable automatic downgrades in the schema processor for the split scenario (at least without logic taking into account some additional context).

Adding such an advanced definition for the deprecated field isn’t particularly complicated, so I don’t mind including it. What do you think? Are there other types of deprecations you’d like to codify?

crates/weaver_resolved_schema/src/lib.rs
lquerel: Point 1 is fixed in https://github.com/open-telemetry/weaver/pull/400/commits/dbe4316822895918eeadbe2459f4bc46fa30c653.

Point 2 is postponed in a future PR.

crates/weaver_resolved_schema/src/lib.rs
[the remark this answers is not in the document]
lquerel: Fixed in https://github.com/open-telemetry/weaver/pull/400/commits/bfdcfa9cb1009add4cadc0de616862f176d50fd7

crates/weaver_resolved_schema/src/lib.rs
```diff
@@ -92,23 +146,375 @@ impl ResolvedTelemetrySchema {
...
+        // The key may refer to an existing attribute in the baseline schema or
+        // a new attribute in the latest schema.
+        let mut renamed_attributes = HashMap::new();
+
+        // ToDo for future PR, process differences at the field level (not required for the schema update)
+
+        // Collect all the information related to the attributes that have been
+        // deprecated in the latest schema.
+        for (attr_name, attr) in latest_attributes.iter() {
+            if let Some(deprecated) = attr.deprecated.as_ref() {
```
jerbly, line 293: I think there needs to be a comparison with the baseline here, otherwise all deprecated attributes will be flagged as changes.
```suggestion
            if let Some(deprecated) = attr.deprecated.as_ref() {
                // is this a change from the baseline?
                if let Some(baseline_attr) = baseline_attributes.get(attr_name) {
                    if let Some(baseline_deprecated) = baseline_attr.deprecated.as_ref() {
                        if deprecated == baseline_deprecated {
                            continue;
                        }
                    }
                }
```

crates/weaver_resolved_schema/src/lib.rs
```diff
@@ -92,23 +146,375 @@ impl ResolvedTelemetrySchema {
...
+        // A map of signal groups that have been renamed to a new signal of same type.
+        // The key is the new name of the signal, and the value is a set of old names.
+        // The key may refer to an existing signal in the baseline schema or
+        // a new signal in the latest schema.
+        let mut renamed_signals: HashMap<String, Vec<&Group>> = HashMap::new();
+
+        // Collect all the information related to the signals that have been
+        // deprecated in the latest schema.
+        for (signal_name, group) in latest_signals.iter() {
+            if let Some(deprecated) = group.deprecated.as_ref() {
```
jerbly, line 411: Again, I think a comparison with the baseline is missing.
```suggestion
            if let Some(deprecated) = group.deprecated.as_ref() {
                // is this a change from the baseline?
                if let Some(baseline_group) = baseline_signals.get(signal_name) {
                    if let Some(baseline_deprecated) = baseline_group.deprecated.as_ref() {
                        if deprecated == baseline_deprecated {
                            continue;
                        }
                    }
                }
```

lquerel: Good catch! Thank you

lquerel: Updated an existing test to catch this issue.

docs/schema-changes.md
```diff
@@ -0,0 +1,119 @@
...
+Example Schema Diff in YAML
+
+```yaml
+head:
+  semconv_version: v1.27.0
+baseline:
+  semconv_version: v1.26.0
+changes:
+  attributes:
+    - type: deprecated
```
lmolkova, line 69: nit and personal taste, so definitely not blocking

having 

```
- name: http.server_name
  type: deprecated
```

looks more readable to me, maybe because it's closer to how we write attribute definitions.

docs/usage.md
```diff
@@ -185,6 +185,69 @@ Options:
...
+
+```
+Generate a diff between two versions of a semantic convention registry.
+
+This diff can then be rendered in multiple formats:
+- a console-friendly format (default: ansi),
+- a structured document in JSON or YAML format,
+- ...
+
+Usage: weaver registry diff [OPTIONS] --baseline-registry <BASELINE_REGISTRY> [OUTPUT]
```
lmolkova, line 198: could I use it with weaver yaml and pipe output to jq/jinja similarly to the registry? 

It would allow generating area-specific migration guides and doing any further customizations in the same way as we do codegen.

(either way, it's not  a blocker)

docs/schema-changes.md
```diff
@@ -0,0 +1,119 @@
...
+- `renamed_to_existing`: One or more schema items in the baseline registry were
+  renamed to an existing item in the head registry. The old names of the items
+  are stored in the old_names attribute, and the existing item name is stored
+  in the current_name attribute.
+- `deprecated`: An item in the baseline registry was deprecated in the head
+  registry. The name of the deprecated item is stored in the name attribute,
+  and the deprecation note is stored in the note attribute.
+- `removed`: An item in the baseline registry was removed in the head
+  registry. The name of the removed item is stored in the name attribute.
+
```
lmolkova, line 54: we'll eventually need a way to describe other (breaking) changes such as attribute type, metric unit or instrument change. 

Maybe we can add `other` category that would be used for change types that we don't formally support yet? This would allow someone to look at the diff and understand that there was some change and maybe some manual intervention is necessary. 
Otherwise, the lack of any mention could be perceived as 'there was no change, all looks good and compatible'.

docs/schema-changes.md
lquerel: Ok I will make this change in the documentation.

docs/schema-changes.md
lquerel: I will add a note and describe how this could be represented as a potential future extension.

docs/usage.md
lquerel: Yes, it's already in place :-)

You can see several examples in the `defaults/diff_templates`.

jsuereth (commented):
Still haven't finished some of the more mechanical parts, but did review the diff algorithm, deserialization core code.

I think we have an issue around attribute identity we may need to solve. PTAL at comments.

crates/weaver_resolved_schema/src/attribute.rs
```diff
@@ -100,3 +101,105 @@ impl Display for AttributeRef {
         write!(f, "AttributeRef({})", self.0)
     }
 }
+
+impl Attribute {
+    /// Creates a new string attribute.
+    /// Note: This constructor is used for testing purposes.
+    #[cfg(test)]
+    pub(crate) fn string<S: AsRef<str>>(name: S, brief: S, note: S) -> Self {
```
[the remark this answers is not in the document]
lquerel: Good catch. I like the proposal.

docs/schema-changes-use-cases.md
```diff
@@ -0,0 +1,605 @@
...
+              case 'server' => attributes['client.address'] = attributes['net.peer.name']
+            }
+          backward: >
+            switch span_kind {
+              case 'client' => attributes['net.peer.name'] = attributes['server.address'],
+              case 'server' => attributes['net.peer.name'] = attributes['client.address']
+            }
+        stability: experimental
+        examples: ['example.com']
+```
```
jerbly, line 147: In this example, the attribute has been deprecated in the attribute_group but the condition is related to its usage in a span. Should the forward / backward instructions be on the span definition where `net.peer.name` is referenced? There may be different instructions for different spans that use `net.peer.name`.

docs/schema-changes-use-cases.md
```diff
@@ -0,0 +1,605 @@
...
+        brief: 'Deprecated, no general replacement at this time. For Elasticsearch, use `db.elasticsearch.node.name` instead.'
+        deprecated:
+          type: conditionally_deprecated
+          forward: >
+            if attributes['db.system'] == 'elasticsearch' then attributes['db.elasticsearch.node.name'] = attributes['db.instance.id']
+            else drop attributes['db.instance.id']
+          backward: >
+            if attributes['db.system'] == 'elasticsearch' then attributes['db.instance.id'] = attributes['db.elasticsearch.node.name']
+        stability: experimental
+```
```
jerbly, line 451: Like my previous comment. When the instructions involve other attributes then the change is dependent on the presence of those attributes in that signal definition. Perhaps defining the instructions on the attribute_group means it's universal for all uses via references? Maybe this can be overridden with further instructions on the span, for example, to deviate from this default?

docs/schema-changes-use-cases.md
```diff
@@ -0,0 +1,605 @@
...
+
+The previous sections focused on renaming and deprecation of top-level items (attributes and signals) in the semantic
+conventions. This section describes how to handle changes in the fields of attributes and signals and changes in the
+attributes of signals.
+
+### Field Changes
+
+There are two types of field changes:
+
+There are two types of field changes:
```
jerbly, line 534: Remove

docs/schema-changes-use-cases.md
lquerel: Good point. It's a good topic for discussion for our next Semantic Conventions Tooling SIG.

docs/schema-changes-use-cases.md
lquerel: Yes, in the `attribute_group`, the setting is intended to be universal. However, I agree that we should demonstrate how this setting can be overridden when dealing with a specific signal.

docs/schema-changes-use-cases.md
lquerel: Fixed

docs/schema-changes.md
lquerel: Done

docs/schema-changes.md
lquerel: Updated the documentation

crates/weaver_common/src/result.rs
```diff
@@ -27,10 +27,50 @@ pub enum WResult<T, E> {
...
+    pub fn has_errors(&self) -> bool {
+        match self {
+            WResult::Ok(_) => false,
+            WResult::OkWithNFEs(_, errors) => !errors.is_empty(),
+            WResult::FatalErr(_) => true,
+        }
+    }
+
+    /// Returns the number of non-fatal errors, or 1 if the result is a fatal error, 0 otherwise.
+    pub fn error_count(&self) -> usize {
```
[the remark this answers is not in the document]
lquerel: Done

crates/weaver_resolved_schema/src/attribute.rs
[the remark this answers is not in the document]
lquerel: Use `impl Into<String>`.

crates/weaver_resolved_schema/src/lib.rs
```diff
@@ -92,23 +146,391 @@ impl ResolvedTelemetrySchema {
...
+                        .expect("Attribute ref not found in catalog. This is a bug.");
+                    (attr.name.as_str(), attr)
+                })
+            })
+            .collect()
+    }
+
+    /// Get the groups of a specific type from the resolved telemetry schema.
+    #[must_use]
+    pub fn groups(&self, group_type: GroupType) -> HashMap<String, &Group> {
```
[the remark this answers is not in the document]
lquerel: No good reason. It's now fixed.

crates/weaver_semconv/src/registry.rs
```diff
@@ -103,20 +110,45 @@ impl SemConvRegistry {
...
-    /// * `registry_id` - The id of the semantic convention registry.
+    /// * `registry_repo` - The semantic convention registry.
     /// * `semconv_specs` - The list of semantic convention specs to load.
     pub fn from_semconv_specs(
-        registry_id: &str,
+        registry_repo: &RegistryRepo,
         semconv_specs: Vec<(String, SemConvSpec)>,
-    ) -> SemConvRegistry {
+    ) -> Result<SemConvRegistry, Error> {
+        static VERSION_REGEX: LazyLock<Regex> =
```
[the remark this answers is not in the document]
lquerel: Created an issue to keep track of this request => https://github.com/open-telemetry/weaver/issues/579

crates/weaver_resolved_schema/src/lib.rs
```diff
@@ -92,23 +146,391 @@ impl ResolvedTelemetrySchema {
...
+                })
+            })
+            .collect()
+    }
+
+    /// Get the "registry" attributes of the resolved telemetry schema.
+    ///
+    /// Note: At the moment (2024-12-30), I don't know a better way to identify
+    /// the "registry" attributes other than by checking if the group ID starts
+    /// with "registry.".
```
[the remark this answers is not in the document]
lquerel: I agree. I created this new issue to keep track of this request. https://github.com/open-telemetry/weaver/issues/580

crates/weaver_resolved_schema/src/lib.rs
```diff
@@ -92,23 +146,391 @@ impl ResolvedTelemetrySchema {
...
         Stats {
-            registry_count: self.registries.len(),
             registry_stats,
             catalog_stats: self.catalog.stats(),
         }
     }
+
+    /// Get the attributes of the resolved telemetry schema.
+    #[must_use]
+    pub fn attribute_map(&self) -> HashMap<&str, &Attribute> {
```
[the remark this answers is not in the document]
lquerel: While waiting for a better identification of registry attributes in the semantic conventions, I have added a method `is_registry_attribute_group` in the `Group` struct with the following definition:

```
pub fn is_registry_attribute_group(&self) -> bool {
        matches!(self.r#type, GroupType::AttributeGroup) && self.id.starts_with("registry.")
}
```

I have also added a method to return only the registry attributes. The “other” `attribute_group` instances are logical groupings of attributes that, by policy, should not introduce new attribute names or modify existing names.

I believe this is sufficient for this version of the schema diff. What do you think?

crates/weaver_version/src/schema_changes.rs
```diff
@@ -0,0 +1,307 @@
...
+        /// The old name of the telemetry object that has been renamed.
+        old_name: String,
+        /// The new name of the telemetry object that has been renamed.
+        new_name: String,
+        /// A note providing further context.
+        note: String,
+    },
+    /// A top-level telemetry object from the baseline registry was marked as deprecated in the head
+    /// registry.
+    Deprecated {
```
jsuereth, line 71: Is this its own change or should it be attached to other changes?

I.e. is it just an implication of change?  

I think this was called out verbally, but it's the one I'm least sure of belonging with other "semantic" changes, especialyl given "uncategorized" as an option.

docs/schema-changes.md
```diff
@@ -0,0 +1,112 @@
...
+Each change in the changes dictionary for any key is represented as a list of
+schema changes, represented by one of the following change types:
+
+- `added`: A top-level telemetry object (e.g., attribute, metric, etc.) was added to the head registry. The new item’s
+  name is stored in the name attribute.
+- `renamed`: A top-level telemetry object from the baseline registry was renamed in the head registry.
+- `deprecated`: A top-level telemetry object from the baseline registry was marked as deprecated in the head registry.
+- `updated`: One or more fields in a top-level telemetry object have been updated in the head registry.
+- `removed`: A top-level telemetry object from the baseline registry was removed in the head registry.
```
jsuereth, line 44: For semconv specifically, we definitely don't want to allow this, instead we deprecate.

Also - my concern with "deprecated" is that when we rename, we're efecctively deprecating the old.

I'm reading this and think "deprecated" is too generic and too much of a catch-all.  I'd rather use "uncategorized", where deprecation is a *consequence* of the change vs. the change itself.

I.e. we almost need a "removed" where we mark the type as deprecated and prevent further usage but don't remove our knowledge it once existed.

lquerel: The `deprecated` type is indeed probably too much of a catch-all. However, I believe these three types are truly distinct, and I probably didn’t do a great job explaining them.

Currently, the general concept of deprecation is used for several types of changes in semantic conventions (renaming, “soft” removal, and other exotic changes). I propose refining my initial suggestion and the corresponding definitions as follows:
- Rename the change type `deprecated` to `obsoleted` to clearly indicate that this change corresponds to an attribute or a signal that is discontinued without a valid replacement.
- In my view, `removed` should exist at the Weaver level, if only to identify that there has been an actual deletion in a registry under validation. This type of change should never be issued for a published registry, but it is clearly a transitional change that can occur during the development of a registry. We could even build a policy leveraging this type of change in the future.
- `uncategorized` is the catch-all change type representing all complex types of changes that we haven’t precisely codified. The idea of this type, as you mentioned during the meeting, is that we should gradually eliminate it from the registry.

Do we agree on this definition of things?

crates/weaver_version/src/schema_changes.rs
lquerel: See my comment below.

docs/schema-changes.md
jsuereth: I like moving `deprecated` to `obsoleted` where `deprecated` can remain as a catch-all for "we changed this thing in a way" and `obsoleted` implies "do not use anymore, here for legacy reasons".

I agree we need to actually model removed in some way.  `obsoleted` as soft-delete works for me.

So yes, I agree on this.

crates/weaver_version/src/schema_changes.rs
jsuereth: Responded.  Once updated, this PR LGTM.

Structurally/rust-wise you have all the pieces I'd look for.  It's just naming/surface syntax at this point.

docs/schema-changes-use-cases.md
lquerel: Following the discussion in the last SIG meeting, it was decided to start with a more minimalist approach initially. As a result, I have updated this PR accordingly, both in the code and the main documentation (not this doc).

However, I am keeping this exploration document in place so that we can revisit this topic later if needed.

docs/schema-changes-use-cases.md
lquerel: See my previous comment.

crates/weaver_forge/expected_output/test/resource/library.md
```diff
@@ -22,7 +22,7 @@ Brief: Span attributes used by non-OTLP exporters to represent OpenTelemetry Sco
 - Examples: [
     "io.opentelemetry.contrib.mongodb",
 ]
-- Deprecated: use the `otel.scope.name` attribute.
+- Deprecated:
```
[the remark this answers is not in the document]
lquerel: I added the following entry in the changelog

What's changed

* **Breaking Change**: Introduced a new `weaver registry diff` command to generate a diff report between two versions of
  the semantic convention registry. This PR introduces a breaking change in the semantic conventions schema. While the
  text-based `deprecated` field is still supported for compatibility reasons, future semantic conventions should use the
  new `deprecated` structured format. ([#400](https://github.com/open-telemetry/weaver/pull/400/) by @lquerel).
  * The `deprecated` field is now a structured field defining the precise reason for deprecation. The semantic
    conventions must be updated to adopt this new format.
  * The changes related to the `deprecated` field (i.e., string → struct) also have a potential impact on certain
    templates that reference the `deprecated` field as containing text. These templates will need to be updated to use
    the `brief` field, which provides a textual explanation of the reasons for the deprecation.

schemas/semconv.schema.json
```diff
@@ -41,6 +41,48 @@
...
+						"reason": {
+							"const": "obsoleted"
+						}
+					},
+					"required": ["reason"],
+					"additionalProperties": false
+				},
+				{
+					"type": "object",
+					"description": "The telemetry object containing the deprecated field has been deprecated for complex reasons (split, merge, ...) which are currently not precisely define in the supported deprecation reasons.",
```
jerbly, line 75:
```suggestion
					"description": "The telemetry object containing the deprecated field has been deprecated for complex reasons (split, merge, ...) which are currently not precisely defined in the supported deprecation reasons.",
```

docs/schema-changes.md
```diff
@@ -27,36 +27,55 @@ The schema changes data model is composed of the following components:
...
 process, it should never be present in the diff report between two versions of
 well-formed semantic convention registries. The policy for semantic convention
 registries is to deprecate items instead of removing them.
 
+Some of these schema change types are based on the new `deprecated` structured format present in the semantic
+conventions. This field is now supporting one of the three variants:
+
+Variant 1
+```yaml
+brief: <text explaining the reason of the renaming>
```
jerbly, line 59: Would it be clearer for the `deprecated` object to have its own note to explain the reason for deprecation. Then the existing brief can remain intact? e.g.
```yaml
brief: Here is the original brief for the attribute 
deprecated:
  explanation: <text explaining the reason of the renaming>
  reason: renamed
  renamed_to: <name of the telemetry object>
```

jerbly (commented):
A few doc nits. Plus a schema suggestion to avoid obscuring the original brief for the item.

schemas/semconv.schema.json
```diff
@@ -41,6 +41,48 @@
 		}
 	},
 	"$defs": {
+		"Deprecated": {
```
lmolkova, line 44: could you please also update https://github.com/open-telemetry/weaver/blob/main/schemas/semconv-syntax.md - an informal and reader-friendly version of it?

docs/schema-changes.md
lmolkova: we currently replace original brief with something like "Deprecated, use 'another.attribute' instead, but nothing stops us from keeping the original brief or adding deprecating details. There is also attribute note and we can decide to keep brief around and add  deprecation details into the note. 

I'd prefer not to add new properties (since we already have breif and note), but it's not a strong opinion.

docs/schema-changes.md
```diff
@@ -0,0 +1,139 @@
...
+```yaml
+head:
+  semconv_version: v1.27.0
+baseline:
+  semconv_version: v1.26.0
+changes:
+  registry_attributes:
+    - name: http.server_name      # attribute name
+      type: obsoleted             # change type
+      note: This attribute is deprecated.
```
lmolkova, line 90: is the note populated from note or brief?

jsuereth (approved):

docs/schema-changes.md
lquerel: @lmolkova @jerbly Currently, the `note` in the schema change event is directly derived from the `brief` field of the telemetry object. Alternatively, we could choose to use the `note` field of the telemetry object. However, what we cannot do is allow users to fill in either one, as Weaver would then not know which one to use.

One possible solution/extension, however, would be to introduce a `note` field within the `deprecated` field (in the semantic conventions). If this field is populated, it would be used to fill the `note` field in the schema change event. If the `deprecated.note` field is not present, Weaver could then fall back to `brief`.

For this PR, I propose using `brief` for now and introducing the `deprecated.note` field if this approach works for you.

docs/schema-changes.md
lquerel: See my comment below.

PR #407, opened 2024-10-09
Add warning that issues when using prefix

### LINKED ISSUE #263: Change what prefix means in the schema

jsuereth:
Fixes #263


lquerel (approved):
LGTM.

I agree that we need to find a way to reduce the amount of yaml files in our unit tests. But on my side, I don't know when I will be able to prioritize that. Thanks again.

PR #429, opened 2024-10-24
Remove race condition from #388 - Use HTTP server for tests

jsuereth:
Updates tests to use http server rather than the internet to test resolution.

Fixes some race conditions found in github test runnners from #388, replaces #388


lquerel (approved):
Thanks for this fix.
The only thing is the debug println! that is still in the code.

Issue #434, opened 2024-10-30
`registry update-markdown` command fails with warning, but returns 0


lmolkova:
Try  https://github.com/open-telemetry/semantic-conventions/commit/5261f966ab4e4102eb454d19d21edc9c46a565a5:

```
/mnt/c/repo/semantic-conventions$ make table-check
docker run --rm \
        --mount 'type=bind,source=/mnt/c/repo/semantic-conventions/templates,target=/home/weaver/templates,readonly' \
        --mount 'type=bind,source=/mnt/c/repo/semantic-conventions/model,target=/home/weaver/source,readonly' \
        --mount 'type=bind,source=/mnt/c/repo/semantic-conventions/docs,target=/home/weaver/target,readonly' \
        otel/weaver:v0.10.0 registry update-markdown \
        --registry=/home/weaver/source \
        --attribute-registry-base-url=/docs/attributes-registry \
        --templates=/home/weaver/templates \
        --target=markdown \
        --dry-run \
        /home/weaver/target

Diagnostic report:

  ⚠ The attribute `process.executable.build_id.profiling` in the group
  │ `registry.process.deprecated` contains an example that will be considered
  │ invalid in the future. This attribute is a string but it does not contain
  │ any examples..
  │ Provenance: "/home/weaver/source/process/deprecated/registry-
  │ deprecated.yaml"



Total execution time: 0.198580721s
/mnt/c/repo/semantic-conventions$ echo $?
0
```

We should either convert this warning into error (it's fine for anything new that we add) or continue the execution

lquerel:
For any new version of the registry, the plan is to use `--future` to convert warnings into errors (resulting in an exit code of 1). Without `--future`, Weaver will emit warnings for this type of check to maintain compatibility with older registry versions.

Please close the issue if my response satisfies you. Thank you.

lmolkova:
I switched to `future` in semconv. The problem that it does not work correctly without a warning - warning stops semconv generation when it occurs but result in 0 code. 

I.e. if I have a warning, I can't generate markdown

lmolkova:
compare the output above with the real md update - the first one didn't update anything, the one below updated everything

```log
neskazu@neskazu:/mnt/c/repo/semantic-conventions$ make table-check
docker run --rm \
        --mount 'type=bind,source=/mnt/c/repo/semantic-conventions/templates,target=/home/weaver/templates,readonly' \
[...]

PR #438, opened 2024-10-31
fix(update-markdown): generate markdown if warnings without errors

lquerel:
This PR implements the following changes in the `registry update-markdown` command:
- If `--future` is present in the arguments, no generation will occur if there are warnings or errors, and the command will return a code of 1.
- If `--future` is not present in the arguments, the markdown generation will proceed only if there are no errors. In this case, warnings will be displayed (if any), markdown will be generated, and the command will return 0. If at least one error is detected, errors (and warnings, if present) will be displayed, no generation will occur, and the command will return a code of 1.

Closes: #434

### DIFF
--- a/crates/weaver_semconv_gen/src/lib.rs
+++ b/crates/weaver_semconv_gen/src/lib.rs
@@ -329,8 +329,8 @@ impl ResolvedSemconvRegistry {
         diag_msgs: &mut DiagnosticMessages,
     ) -> Result<ResolvedSemconvRegistry, Error> {
         let registry_id = "semantic_conventions";
-        let semconv_specs =
-            SchemaResolver::load_semconv_specs(registry_repo).capture_non_fatal_errors(diag_msgs)?;
+        let semconv_specs = SchemaResolver::load_semconv_specs(registry_repo)
+            .capture_non_fatal_errors(diag_msgs)?;
         let mut registry = SemConvRegistry::from_semconv_specs(registry_id, semconv_specs);
         let schema = SchemaResolver::resolve_semantic_convention_registry(&mut registry)?;
         let lookup = ResolvedSemconvRegistry {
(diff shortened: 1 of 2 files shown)


lmolkova (approved):
Thanks a lot!

jsuereth:
it looks like my async fix for the http  server is still broken.

I suggest we find a different library with easier mechanism/ownership model to cleanly shut it down.  

The current one is deadlocking and I can't duplicate it locally

.github/workflows/ci.yml
```diff
@@ -119,7 +119,7 @@ jobs:
         uses: dtolnay/rust-toolchain@stable
         with:
           # cargo check-external-types is only available on this specific nightly
-          toolchain: nightly-2024-05-01
+          toolchain: nightly-2024-06-30
```
jsuereth, line 122: Can you move this into its own fix?

crates/weaver_semconv_gen/src/lib.rs
```diff
@@ -325,10 +326,11 @@ impl ResolvedSemconvRegistry {
     /// Resolve semconv registry (possibly from git), and make it available for rendering.
     fn try_from_registry_repo(
         registry_repo: &RegistryRepo,
+        diag_msgs: &mut DiagnosticMessages,
```
jsuereth, line 329: Nit: this should be called `warnings` or some such to denote that only non-fatal messages will be recorded here.

.github/workflows/ci.yml
lquerel: Yes, it can be done.

lquerel: See https://github.com/open-telemetry/weaver/pull/442

crates/weaver_semconv_gen/src/lib.rs
lquerel: `warnings` may not be the best name, as diagnostic messages can include warnings, advice, or non-fatal errors. Why not stick with diagnostic messages instead?

jsuereth: Mostly I just want to denote that it's only non-fatal errors would show up here.

PR #449, opened 2024-11-06
Move from tinyhttp to rouille

jsuereth:
Move away from TinyHTTP to a framework that provides:

1. A way to server static files via one function.
2. A better concurrency solution that isn't fraught with peril for our usage.


lquerel (approved):
Looks simpler and working on Windows so it's all good!

lquerel:
I didn’t merge because I wasn’t sure if you wanted to continue working on this or not.

jsuereth:
@lquerel No, let's merge this to fix up flaky PR checks and we can move to a new crate in a different PR.

PR #453, opened 2024-11-08
Update comment filter to remove trailing spaces

### LINKED ISSUE #417: No trailing spaces in comments
The comment formatter allows you to specify a prefix for comments. Setting this to something like `// ` results in the following when formatting "foo\n\nbar\n":

```go
// foo
// 
// bar
```

The middle line has a trailing space from the specified prefix.

How can I not have this? For example:

```go
// foo
//
// bar
```

Is there a way to have a prefix for lines that contain characters and another for empty lines?

jsuereth:
When using `trim` remove trailing spaces on multi-line comments.

Fixes #417

Overall this seems more like a "bug fix" than a breaking change.  The only caveat may be python.

### DIFF
--- a/crates/weaver_forge/expected_output/comment_format/example.go
+++ b/crates/weaver_forge/expected_output/comment_format/example.go
@@ -2,30 +2,30 @@
 
   // DEVICE_ID
   // A unique identifier representing the device
-  // 
+  //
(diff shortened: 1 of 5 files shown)


lquerel (approved):
LGTM

PR #454, opened 2024-11-08
Provide max line-length in comment filter.

### LINKED ISSUE #364: Add the ability to word-wrap or limit line length
We should be able to define the maximum line length for comments, for example. Several approaches are possible:  
1. Add a `word_wrap` filter that would take a maximum number of characters per line as a parameter.  
2. Add an optional parameter to the existing `indent` filter to specify a maximum number of characters per line, accounting for indentation.  
3. Add a configuration option in the `comment_formats` section to specify the preferred maximum line length for comments, per language.  

These options are not mutually exclusive. Personally, I would be inclined to start with 2 and 3.

jsuereth:
Provides naive word-wrapping capabilities to the comment filter.

Fixes #364

- Adds new `word_wrap` config section with basic set of features for controlling word-wrapping
- Pulls in `textwrap` library, but only for pieces. Unfortunately, we need dynamic word-wrapping that is syntax-aware of markdown and html.
- Adds new configuration parameter for rendering ordered lists in golang's desired format.
- Adds a configuration parameter to determine if you ignore endlines when word wrapping to repack paragraphs.
  **Note: this feature is naive and likely needs to be reworked, longer term**
- Adds a new set of tests to ensure word-wrapping works

### DIFF
--- a/crates/weaver_forge/src/formats/markdown.rs
+++ b/crates/weaver_forge/src/formats/markdown.rs
@@ -579,9 +579,9 @@ mod tests {
         let renderer = MarkdownRenderer::try_new(&config)?;
         let markdown = r##"In some cases a URL may refer to an IP and/or port directly,
           The file extension extracted from the `url.full`, excluding the leading dot."##;
-        let html = renderer.render(markdown, "go", None)?;
+        let rendered_md = renderer.render(markdown, "go", None)?;
         assert_string_eq!(
-            &html,
+            &rendered_md,
             r##"In some cases a URL may refer to an IP and/or port directly,
 The file extension extracted from the `url.full`, excluding the leading dot.
 "## // ToDo why a new line at the end?
@@ -594,9 +594,9 @@ and specifically the
 
 An example can be found in
 [Example Image Manifest](https://docs.docker.com/registry/spec/manifest-v2-2/#example-image-manifest)."##;
-        let html = renderer.render(markdown, "go", None)?;
+        let rendered_md = renderer.render(markdown, "go", None)?;
         assert_string_eq!(
-            &html,
+            &rendered_md,
             r##"Follows
 [OCI Image Manifest Specification],
 and specifically the
@@ -629,9 +629,9 @@ it's RECOMMENDED to:
 
 * Use a domain-specific attribute
 * Set `error.type` to capture all errors, regardless of whether they are defined within the domain-specific set or not."##;
-        let html = renderer.render(markdown, "go", None)?;
+        let rendered_md = renderer.render(markdown, "go", None)?;
         assert_string_eq!(
-            &html,
+            &rendered_md,
             r##"The `error.type` SHOULD be predictable, and SHOULD have low cardinality.
 
 When `error.type` is set to a type (e.g., an exception type), its
@@ -689,9 +689,9 @@ it's RECOMMENDED to:
         let renderer = MarkdownRenderer::try_new(&config)?;
         let markdown = r##"In some cases a [URL] may refer to an [IP](http://ip.com) and/or port directly,
           The file \\[extension\\] extracted \\[from] the `url.full`, excluding the leading dot."##;
-        let html = renderer.render(markdown, "go", None)?;
+        let rendered_md = renderer.render(markdown, "go", None)?;
         assert_string_eq!(
-            &html,
+            &rendered_md,
             r##"In some cases a [URL] may refer to an [IP] and/or port directly,
 The file \[extension\] extracted \[from] the `url.full`, excluding the leading dot.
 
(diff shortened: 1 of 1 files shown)


jsuereth:
Frustratingly, I'm running into an oddity in `go fmt` that I'm not sure we should make the default for weaver or somehow configure.

Specifically the [go fmt comment spec](https://tip.golang.org/doc/comment) states the following about lists:

> Gofmt reformats bullet lists to use a dash as the bullet marker, two spaces of indentation before the dash, and four spaces of indentation for continuation lines.
>
> Gofmt reformats numbered lists to use a single space before the number, a period after the number, and again four spaces of indentation for continuation lines.

I'm not positive why there's a single-space before number lists and two spaces before dash lists, but we do NOT do this in weaver today.

@lquerel Thoughts?  Should I create a *specific* renderer for markdown for Go that abides by these restrictions? 

Additionally, `go fmt` does not allow nested lists.

@MrAlias What should we do if we have a nested list?  Will Go (and go fmt) ignore this or fail?

crates/weaver_forge/src/extensions/code.rs
```diff
@@ -449,6 +465,68 @@ it's RECOMMENDED to:
...
+ * its own set of error
+ * identifiers (such as HTTP or
+ * gRPC status codes), it's
+ * RECOMMENDED to:
+ * <p>
+ * <ul>
+ *   <li>Use a domain-specific
+ *   attribute
+ *   <li>Set {@code error.type}
+ *    to capture all errors,
```
lquerel, line 522: This is indeed strange.

jsuereth: I need to more formally pull words and remember prefix/postfix space.  The library allows this and uses it, it just doesn't understand *changing* the line prefix over time.


I hope we can, eventually, fix this all up, I just ran out of time to tinker

crates/weaver_forge/src/formats/markdown.rs
```diff
@@ -81,10 +104,19 @@ impl RenderContext {
     /// Add a blank line if the current markdown buffer
     /// does not end already with a double newline.
     fn add_cond_blank_line(&mut self) {
+        // TODO - This is a workaround for not truly
+        // refactoring word-wrap vs. regular add-text.
         if !self.markdown.ends_with("\n\n") && !self.markdown.is_empty() {
-            self.markdown.push('\n');
+            let _ = self
```
lquerel, line 110: Why not using the `add_blank_line` defined just after?

crates/weaver_forge/src/formats/markdown.rs
```diff
@@ -102,17 +134,78 @@ impl RenderContext {
         self.skip_line_prefix_on_first_line = false;
     }
 
+    fn start_unbreakable_block(&mut self, text: &str) {
+        // TODO - check for existing unbreakable.
+        if let Some(buf) = self.unbreakable_buffer.as_ref() {
+            // ToDo - we should error out here.
+            // For now, we just FLUSH this to write to the buffer.
+            let _ = self
```
lquerel, line 142: `add_blank_line` ?

lquerel (approved):
Some minor suggestions to clarify the code in certain places. The number of TODOs is not reassuring, but the test results are reassuring. 
Thank you!

crates/weaver_forge/src/formats/markdown.rs
jsuereth: It adds a prefix, this does not.  `add_blank_line` should be fine, I may need to go update the prefix handling code.

crates/weaver_forge/src/formats/markdown.rs
jsuereth: Good question - when this happens our algorithm is basically broken.  I debated between "panic"-ing and implementing.

Regarding blank lines, we have no idea what we've been buffering or why we need to buffer twice, so I think it's unlikely we have a necessary blank line, it's more likely we somehow have nested code + strong/emphasis blocks.

crates/weaver_forge/src/formats/markdown.rs
```diff
@@ -102,17 +134,78 @@ impl RenderContext {
...
+                self.markdown.push_str(line);
             }
-            self.markdown.push_str(line);
+        }
+    }
+    fn add_unbreakable_text(&mut self, text: &str) {
+        if let Some(buf) = self.unbreakable_buffer.as_mut() {
+            buf.push_str(text);
+        } else {
+            let _ = self
```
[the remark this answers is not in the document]
jsuereth: No, unbreakable text ==  a cohesive word that cannot be split, e.g.

`[name]: link` <- we treat this as unbreakable text.

crates/weaver_forge/src/formats/markdown.rs
```diff
@@ -569,15 +734,498 @@ The file \[extension\] extracted \[from] the `url.full`, excluding the leading d
...
+                .collect(),
+            ),
+            default_comment_format: Some("go".to_owned()),
+            ..WeaverConfig::default()
+        };
+
+        let renderer = MarkdownRenderer::try_new(&config)?;
+        let markdown = r##"In some cases a URL may refer to an IP and/or port directly,
+          The file extension extracted from the `url.full`, excluding the leading dot."##;
+        let html = renderer.render(markdown, "go", None)?;
```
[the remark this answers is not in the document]
jsuereth: Would you like me to do that to the whole file?  This whole file calls it "html" previously,, so I just kept it.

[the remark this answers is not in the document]
jsuereth: Went ahead and updated all of them.

PR #467, opened 2024-11-19
Update semconv schema to define `stability` as optional for attribute groups.

jerbly:
@lmolkova wrote:
> ... the attribute_group stability does not make sense - https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/group-stability.md#group-stability. It'd be really awesome if we changed yaml schema and tooling behind it to reflect this

This change updates the semconv schema and the syntax doc to correctly define `stability` as optional for attribute groups.


schemas/semconv.schema.json
```diff
@@ -444,6 +457,13 @@
 				},
 				"type": {
 					"$ref": "#/$defs/AttributeType"
+				},
+				"stability": {
+					"allOf": [
+						{
+							"$ref": "#/$defs/StabilityLevel"
+						}
+					]
```
jerbly, line 466: I noticed that `AttributeFullSpec` had `stability` as required but without the property!

schemas/semconv-syntax.md
```diff
@@ -103,9 +103,9 @@ specificfields ::= spanfields
                |   eventfields
                |   metricfields
 
-spanfields ::= [events] [span_kind]
+spanfields ::= [events] [span_kind] stability
```
lquerel, line 106: If spans, events, and metrics require the stability field then we should update Weaver accordingly, see the validate method of the `GroupSpec` structure in the `weaver_semconv` crate https://github.com/open-telemetry/weaver/blob/cab3125d958a2caad3b62b1c4ff13ea0fc5905f0/crates/weaver_semconv/src/group.rs#L98

lquerel (changes requested):

schemas/semconv-syntax.md
jerbly: I'll take a look. Note though that prior to this PR all the group types required the stability field.

jerbly: @lquerel - I have added the validity check. Since the otel semantic conventions do not meet the otel spec currently, the CLI test fails here:
https://github.com/open-telemetry/weaver/blob/cab3125d958a2caad3b62b1c4ff13ea0fc5905f0/tests/registry_check.rs#L11-L22

lquerel: I’m adding @jsuereth and @lmolkova to this thread because, depending on the chosen option, we may or may not need to make changes to the release of the next semantic convention registry.

@jerbly When we encounter this type of problem (i.e., something that was valid until now but becomes invalid in future versions of the registry), we follow the following pattern:
- We define a new error with the annotation `#[diagnostic(severity(Warning))]` (see the error `InvalidExampleWarning` to see an existing example).
- Then the new validation mechanism raises this warning.
- By default, warnings are displayed but do not affect the exit code of the Weaver command. So the test you mentioned should be ok.
- The `--future` argument is used to convert warnings into errors, which will affect the exit code of Weaver. This argument is explicitly used in the CI/CD pipeline for semantic conventions to validate the next release.

Thus, we have two options:
1. Follow the usual pattern, which means we need to make the necessary changes in the semconv registry to allow a release that passes the Weaver check.
2. Temporarily remove the check I asked Jeremy to add so as not to delay the release of a possible upcoming semconv release in the next few days.

Personally, I'm in favor of option 1. 

What do you think?

jerbly: @lquerel - Option 1, following the current pattern, makes the most sense IMO. I'm implementing that now, there is a lot of test data to change!

lquerel (approved):
LGTM. Thanks

jsuereth (approved):

PR #468, opened 2024-11-20
[WIP] Add symlink parameter

### LINKED ISSUE #441: Generate attributes from files in symlinked directories
When trying to run `weaver registry generate` on a local path, I'd like the ability to "compose" the `model/` folder via symlinking directories present in the filesystem. Currently if I do symbolic linking (`ln -s`) weaver does not read from the linked directories when generating the attributes.

An option that can be passed to `weaver` that would make it read sym linked folders under `model/` and generate attributes

Right now the only way to achieve what I'm thinking is if I copy specific directories into the `model/` folder. This means if I want to keep an in-house component that need to be coalesced with semantic conventions then changes from either side need be explicitly propagated to the `model/` folder, which seems a little more cumbersome than allowing changes to be propagated via symbolic links.

Leo6Leo:
fixes #441 

**Summary of the change I have made:**
Added an option to follow symbolic links when loading the registry in various parts of the codebase. The primary change is the addition of a `follow_symlinks` parameter, which is passed through multiple functions and structures to control this behavior.

Key changes include:

- Added `follow_symlinks` param as `WeaverArg`  across `SchemaResolver`, `SnippetGenerator`, and `ResolvedSemconvRegistry` to enable optional symbolic link following during registry operations.
-The `WeaverArgs` struct impacted various commands (`check`, `generate`, `resolve`, `update-markdown`, `stats`) to include and handle the `follow_symlinks` option.
- Modified `SchemaResolver` to configure `walkdir::WalkDir` with the `follow_symlinks` parameter and use the `follow_links function`.
- Modified the existing tests so that they will default to use `follow_symlinks` = false.

**The ideal outcome would be:**
After running the command
```
weaver registry generate -r ./semantic-conventions/model -t ./semantic-conventions/templates markdown -f
```
weaver will read symbolic linked folders under `sementic-conventions/model` and generate the artifacts.

PS: I try to raise this PR first to get some early feedback. As I am new to Rust and Otel, I am open to suggestions and feedback.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -45,6 +45,18 @@ impl Default for DiagnosticArgs {
     }
 }
 
+/// Set of Parameters used to specify the extra options for the `weaver` command.
+/// The WeaverArgs will be shared across all commands. So only the general options should be
+/// included here.
+#[derive(Args, Debug)]
+pub struct WeaverArgs {
+   /// Boolean flag to specify whether to follow symlinks when loading the registry.
+   /// Default is false.
+    #[arg(short, long)]
+    pub(crate) follow_symlinks: bool,
+   
+}
+
 /// Result of a command execution.
 #[derive(Debug)]
 pub(crate) struct CmdResult {
--- a/src/registry/check.rs
+++ b/src/registry/check.rs
@@ -17,7 +17,7 @@ use crate::registry::RegistryArgs;
 use crate::util::{
     check_policy, check_policy_stage, init_policy_engine, load_semconv_specs, resolve_semconv_specs,
 };
-use crate::{DiagnosticArgs, ExitDirectives};
+use crate::{DiagnosticArgs, ExitDirectives, WeaverArgs};
 
 /// Parameters for the `registry check` sub-command
 #[derive(Debug, Args)]
@@ -47,6 +47,10 @@ pub struct RegistryCheckArgs {
     /// Parameters to specify the diagnostic format.
     #[command(flatten)]
     pub diagnostic: DiagnosticArgs,
+
+    /// Weaver parameters
+    #[command(flatten)]
+    pub weaver_args: WeaverArgs,
 }
 
 /// Check a semantic convention registry.
@@ -78,7 +82,7 @@ pub(crate) fn command(
 
     // Load the semantic convention registry into a local registry repo.
     // No parsing errors should be observed.
-    let main_semconv_specs = load_semconv_specs(&main_registry_repo, logger.clone())
+    let main_semconv_specs = load_semconv_specs(&main_registry_repo, logger.clone(), args.weaver_args.follow_symlinks)
         .capture_non_fatal_errors(&mut diag_msgs)?;
     let baseline_semconv_specs = baseline_registry_repo
         .as_ref()
@@ -86,7 +90,7 @@ pub(crate) fn command(
             // Baseline registry resolution should allow non-future features
             // and warnings against it should be suppressed when evaluating
             // against it as a "baseline".
-            load_semconv_specs(repo, logger.clone())
+            load_semconv_specs(repo, logger.clone(),args.weaver_args.follow_symlinks)
                 .ignore(|e| matches!(e.severity(), Some(miette::Severity::Warning)))
                 .capture_non_fatal_errors(&mut diag_msgs)
         })
@@ -221,7 +225,7 @@ mod tests {
     use crate::registry::{
         semconv_registry, RegistryArgs, RegistryCommand, RegistryPath, RegistrySubCommand,
     };
-    use crate::run_command;
+    use crate::{run_command, WeaverArgs};
 
     #[test]
     fn test_registry_check_exit_code() {
@@ -243,6 +247,9 @@ mod tests {
                     skip_policies: true,
                     display_policy_coverage: false,
                     diagnostic: Default::default(),
+                    weaver_args: WeaverArgs {
+                        follow_symlinks: false,
+                    },
                 }),
             })),
         };
@@ -269,6 +276,9 @@ mod tests {
                     skip_policies: false,
                     display_policy_coverage: false,
                     diagnostic: Default::default(),
+                    weaver_args: WeaverArgs {
+                        follow_symlinks: false,
+                    },
                 }),
             })),
         };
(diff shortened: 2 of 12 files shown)


src/main.rs
```diff
@@ -45,6 +45,18 @@ impl Default for DiagnosticArgs {
     }
 }
 
+/// Set of Parameters used to specify the extra options for the `weaver` command.
+/// The WeaverArgs will be shared across all commands. So only the general options should be
+/// included here.
+#[derive(Args, Debug)]
+pub struct WeaverArgs {
```
lquerel, line 52: I suggest renaming this struct to `CommonRegistryArgs` and moving it to `src/registry/mod.rs`, as it will encapsulate the common arguments for all the `registry` subcommands. For global arguments (e.g., --debug), Clap already provides a mechanism. However, that mechanism cannot be used here because `--follow-symlinks` does not apply to commands unrelated to the registry.

lquerel (changes requested):
This is going in the right direction.

Leo6Leo:
I have fixed the changes you mentioned above and I manually tested the command with a symbolic link folder, and it worked flawlessly. Please let me know if there are any other major changes needed; otherwise, I’ll proceed with writing the tests! @lquerel

lquerel:
> I have fixed the changes you mentioned above and I manually tested the command with a symbolic link folder, and it worked flawlessly. Please let me know if there are any other major changes needed; otherwise, I’ll proceed with writing the tests! @lquerel

@Leo6Leo Sounds good. Let me know when you are ready for the next review. Thanks

Leo6Leo:
@lquerel The tests have been added, the PR should be ready for review imo, PTAL, thanks!

src/registry/generate.rs
```diff
@@ -402,4 +421,101 @@ mod tests {
 
         assert_eq!(rust_files, expected_rust_files);
     }
+
+    #[test]
+    fn test_registry_generate_with_symbolic_link_cases() {
```
lquerel, line 426: The main issue I’m facing is having a unit test in the main crate that depends on test data from another crate (`crates/weaver_codegen_test`). I think it would be better to have this symbolic link in the `data` directory of the main crate to avoid this kind of dependency, which might cause problems at some point.

src/registry/mod.rs
```diff
@@ -157,3 +157,14 @@ pub fn semconv_registry(log: impl Logger + Sync + Clone, command: &RegistryComma
         ),
     }
 }
+
+/// Set of Parameters used to specify the extra options for the `weaver` command.
```
lquerel, line 161:
```suggestion
/// Set of Parameters used to specify the extra options for the `weaver registry` command.
```

src/registry/mod.rs
```diff
@@ -157,3 +157,14 @@ pub fn semconv_registry(log: impl Logger + Sync + Clone, command: &RegistryComma
...
+/// The CommonRegistryArgs will be shared across all commands. So only the general options should be
```
lquerel, line 162:
```suggestion
/// The CommonRegistryArgs will be shared across all `weaver registry` sub-commands. So only the general options should be
```

src/registry/resolve.rs
```diff
@@ -56,6 +56,10 @@ pub struct RegistryResolveArgs {
     /// Parameters to specify the diagnostic format.
     #[command(flatten)]
     pub diagnostic: DiagnosticArgs,
+
+    // Weaver parameters
```
lquerel, line 60:
```suggestion
    // Common weaver registry parameters
```

src/registry/search.rs
```diff
@@ -52,6 +52,10 @@ pub struct RegistrySearchArgs {
     /// An (optional) search string to use.  If specified, will return matching values on the command line.
     /// Otherwise, runs an interactive terminal UI.
     pub search_string: Option<String>,
+
+    /// Weaver parameters
```
lquerel, line 56:
```suggestion
    /// Common weaver registry parameters
```

src/registry/stats.rs
```diff
@@ -25,6 +25,10 @@ pub struct RegistryStatsArgs {
     /// Parameters to specify the diagnostic format.
     #[command(flatten)]
     pub diagnostic: DiagnosticArgs,
+
+    /// Weaver parameters
```
lquerel, line 29:
```suggestion
    /// Common weaver registry parameters
```

src/registry/update_markdown.rs
```diff
@@ -49,6 +49,10 @@ pub struct RegistryUpdateMarkdownArgs {
     /// Parameters to specify the diagnostic format.
     #[command(flatten)]
     pub diagnostic: DiagnosticArgs,
+
+    /// Weaver parameters
```
lquerel, line 53:
```suggestion
    /// Common weaver registry parameters
```

src/registry/check.rs
```diff
@@ -47,6 +47,10 @@ pub struct RegistryCheckArgs {
     /// Parameters to specify the diagnostic format.
     #[command(flatten)]
     pub diagnostic: DiagnosticArgs,
+
+    /// Weaver parameters
```
lquerel, line 51:
```suggestion
    /// Common weaver registry parameters
```

src/registry/generate.rs
```diff
@@ -73,6 +73,10 @@ pub struct RegistryGenerateArgs {
     /// Parameters to specify the diagnostic format.
     #[command(flatten)]
     pub diagnostic: DiagnosticArgs,
+
+    /// Weaver parameters
```
lquerel, line 77:
```suggestion
    /// Common weaver registry parameters
```

lquerel (changes requested):
Overall, it looks good to me. I just made a few requests for changes, and once those are addressed, it’s good to go for approval.

src/registry/generate.rs
Leo6Leo: Thank you for pointing that out! The reason I created the symbolic link in `crates/weaver_codegen_test` is because, based on the function [test_registry_generate](https://github.com/Leo6Leo/weaver/blob/38154d64478182cc3287aeb5766deb1e8392868c/src/registry/generate.rs#L245), it appears to rely on the registry located in `crates/weaver_codegen_test/semconv_registry/`.

I think this indicates that the test depends on the `weaver_codegen_test` crate to access the registry. To ensure the required dependencies are properly linked, I created the symbolic link there to align with this structure.

But I have created the commit to reflect the suggestion you recommend, PTAL https://github.com/open-telemetry/weaver/pull/468/commits/f985907f7dd92264d1211d453fbc681832287b75 @lquerel

lquerel (approved):
LGTM. Thank you for this contribution!

Issue #470, opened 2024-11-21
Determine and Enforce Correct Minimum Rust Compiler Version in Build Pipeline


lquerel:
A user reported that the project fails to compile with Rust compiler version 1.79.0. Currently, our Cargo.toml specifies version 1.76.0 as the minimum, which seems insufficient based on this feedback. We need to determine the actual minimum compiler version required to compile the project. More importantly, we need to identify the right way to test this, as this issue was not caught in our build pipeline.

unpervertedkid:
@lquerel , can I give a shot at this?

lquerel:
Yes sure. You are welcome.

jsuereth:
I believe this is now fixed.  Thanks @unpervertedkid !

PR #473, opened 2024-11-23
Determine and Enforce Correct Minimum Rust Compiler Version in Build Pipeline #470

unpervertedkid:
# Problem

#470 

# Initial findings

1. Incorrect msrv set in cargo.toml(1.76.0) and the one being used in the step doing the validation(1.70).
2. Project is on wrong msrv, 1.76.0. Should be on 1.81.0

# Approach
1. First validate that the expected toolchain is being used to run the msrv check
2. If step 1 above false, look into ways to fix.
3. If step 1 is true, look into why the check passes yet it's not supposed to.
5. Look into ways we can avoid redefining the same value(msrv) twice, in cargo.toml and in the pipeline step since they will eventually be out of sync.(initial idea - use https://github.com/foresterre/cargo-msrv)

# Findings 
1. The issue was caused by the toolchain supposed to be used being overridden  by the toolchain defined in `rust-toolchain.toml`
```text
info: using existing install for '1.76.0-x86_64-unknown-linux-gnu'
info: default toolchain set to '1.76.0-x86_64-unknown-linux-gnu'

  1.76.0-x86_64-unknown-linux-gnu unchanged - rustc 1.76.0 (07dca489a 2024-02-04)
info: note that the toolchain 'stable-x86_64-unknown-linux-gnu' is currently in use (overridden by '/home/runner/work/weaver/weaver/rust-toolchain.toml')
```

# Solution
1. Ensure the correct toolchain is used to run the test by adding a step to override the toolchain to the msrv specific toolchain.
2. Validating that correct toolchain is being used.

# Results
| With msrv of 1.76.0 | With msrv of 1.81.0 |
| ------------------ | ------------------ |
| image | image |


jsuereth:
Does this fail if our MSRV changes?

unpervertedkid:
@jsuereth by "this" you mean what specifically?

I first used cargo msrv to find the "correct" msrv and then manually tried to compile with lower "msrv"'s.
Currently this was not caught in our msrv check in ci.

unpervertedkid:
@jsuereth, latest iteration should be good for review.

PR #480, opened 2024-11-27
Update acronym filter to ignore word boundaries.

### LINKED ISSUE #415: Custom Capitalizations
The Go team currently uses the following capitalization for known acronyms and initialisms:

- ACL
- AI
- AIX
- AKS
- AMD64
- API
- ARM32
- ARM64
- ARN
- ARNs
- ASCII
- ASPNETCore
- AWS
- CICD
- CPP
- CPU
- CSS
- CosmosDB
- CouchDB
- CronJob
- DB
- DC
- DNS
- DaemonSet
- DragonflyBSD
- DynamoDB
- EC2
- ECS
- EDB
- EKS
- EOF
- FaaS
- FirstSQL
- FreeBSD
- GC
- GCP
- GRPC
- GUID
- HBase
- HPUX
- HSQLDB
- HTML
- HTTP
- HTTPS
- HanaDB
- IA64
- ID
- IP
- InProc
- InstantDB
- JDBC
- JSON
- JVM
- K8S
- LHS
- MSSQL
- MariaDB
- MaxDB
- MongoDB
- MySQL
- NetBSD
- OS
- OTel
- OpenBSD
- PHP
- PID
- PPC32
- PPC64
- PostgreSQL
- QPS
- QUIC
- RAM
- RHS
- RPC
- ReplicaSet
- SDK
- SLA
- SMTP
- SPDY
- SQL
- SSH
- StatefulSet
- TCP
- TLS
- TTL
- UDP
- UI
- UID
- URI
- URL
- UTF8
- UUID
- V8JS
- VCS
- VM
- WebEngine
- XML
- XMPP
- XSRF
- XSS
- ZOS

The current build tooling ensures these capitalizations are used in any naming generated. How can we keep this behavior when using weaver?

jsuereth:
Fixes #415

### DIFF
--- a/crates/weaver_forge/src/extensions/util.rs
+++ b/crates/weaver_forge/src/extensions/util.rs
@@ -8,7 +8,6 @@ use minijinja::{Environment, ErrorKind, Value};
 use regex::Regex;
 use std::borrow::Cow;
 use std::collections::HashMap;
-use std::sync::OnceLock;
 
 /// Add utility filters and tests to the environment.
 pub(crate) fn add_filters(env: &mut Environment<'_>, target_config: &WeaverConfig) {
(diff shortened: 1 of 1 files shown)


lquerel:
Thinking back about this problem, I believe this approach is problematic. For example, let's imagine these two metrics:
- `radios.signal.strength`
- `fluid.pressure`

If we apply these filters `title_case | acronym` to the previous examples, it gives us:
- `RadiOSSignalStrength`
- `FlUIDPressure`

Which is probably not what we want either.

I'm thinking about these two approaches:
- We add an additional filter parameter `_case` that corresponds to the `acronym` flag (false by default) which, if true, must be inserted into the filter logic to produce the desired effect. Whether it's possible to configure the crate we're using for this, I don't know.
- We keep the current filters "as is" but instead of having them produce character strings, they produce an object containing a list of tokens with properties specifying the forbidden transformations per token. This approach is more generic but more complicated. This object must probably implement `Display` for the final rendering.

There might be other simpler approaches.

jsuereth:
> Thinking back about this problem, I believe this approach is problematic. For example, let's imagine these two metrics:
> 
> * `radios.signal.strength`
> * `fluid.pressure`
> 
> If we apply these filters `title_case | acronym` to the previous examples, it gives us:
> 
> * `RadiOSSignalStrength`
> * `FlUIDPressure`
> 
> Which is probably not what we want either.
> 
> I'm thinking about these two approaches:
> 
> * We add an additional filter parameter `_case` that corresponds to the `acronym` flag (false by default) which, if true, must be inserted into the filter logic to produce the desired effect. Whether it's possible to configure the crate we're using for this, I don't know.
> * We keep the current filters "as is" but instead of having them produce character strings, they produce an object containing a list of tokens with properties specifying the forbidden transformations per token. This approach is more generic but more complicated. This object must probably implement `Display` for the final rendering.
> 
> There might be other simpler approaches.

I see two options here:

- we find a way for acronyms to interact directly with `*_case` filters.
- we have configuration options for. Acronyms filter that imply how to split word boundaries.
Which one would your other me explore?

lquerel:
I’m not sure which one will be the easiest to implement. I think I would start by checking if there is a hook mechanism in the library used for casing, and if there isn’t, I would probably try to address the issue in the acronym filter.

Issue #405, opened 2024-03-27
Semconv schema: don't require attributes on metrics


lmolkova:
Json schema validation shows error when metrics is defined without attributes
<img width="356" alt="image" src="https://github.com/open-telemetry/weaver/assets/2347409/4648ec97-c6b8-4d7b-8f65-9458c02e2189">

with `Missing property "attributes".yaml-schema: https://raw.githubusercontent.com/open-telemetry/build-tools/v0.24.0/semantic-conventions/semconv.schema.json`

Metrics don't have to have attributes or may inherit them through extension mechanism.

jerbly:
@lmolkova @jsuereth - I'm happy to take this one on if you would like to assign it to me. It's similar to a PR I just did for group stability: https://github.com/open-telemetry/weaver/pull/467

I will change the EBNF and json schema. However, there is currently a difference in this area between the EBNF and the schema. The EBNF currently defines `attributes` as mandatory for all group types and `extends` optional. Whereas the schema defines `attributes` and `extends` with an `anyOf` meaning it's valid to omit `attributes` provided you have `extends` instead. There's currently one single non-metric defined with `extends` and without `attributes`:

```yaml
  - id: span.db.client
    type: span
    stability: experimental
    brief: This span defines the attributes used to perform database client calls.
    span_kind: client
    extends: trace.db.common.full
```
Is this a valid case? If so the EBNF should be defined like this `( extends | attributes | ( extends attributes ) )`.

jsuereth:
This is a valid use case.

lquerel:
Done in #494

PR #494, opened 2024-11-30
Metrics and Events don't require attributes

jerbly:
Fix for issue #405 
- Updated the EBNF and JSON schema to define the `extends` or `attributes` requirement mandatory for all group types except `metric` and `event`.
- Added a group validity check as a warning.


lquerel (approved):
LGTM. Thanks

lmolkova (approved):
LGTM and thank you!

lquerel:
@jerbly This PR is still in draft mode. Is that intentional? If not, please click “Ready for Review,” and I’ll proceed with merging it. Thanks!

jerbly:
@lquerel it's currently failing unrelated clippy checks. One is a simple fix. The other is documentation and formatting in the generated code in weaver-codegen - I'm not sure what to do with it. Perhaps change clippy settings in that crate?

lquerel:
Regarding Clippy, I suggest implementing the changes proposed in Clippy’s report regarding the possibility of eliding certain lifetimes. Currently, our build pipeline is configured to always use the latest version of the Rust toolchain, which has the advantage of benefiting from the latest changes and improvements but also the drawback of introducing new Clippy errors unrelated to the PR. We could pin the toolchain version to avoid this, but at the same time, the current approach ensures we don’t forget to update it.

jerbly:
The lifetimes one is easy. The doc problems in the generated code was where I had my doubts it was worth the effort. I'll work on it.

jerbly:
@lquerel - I've fixed those issues so this is ready to merge now. :)

crates/weaver_codegen_test/templates/registry/rust/attributes/mod.rs.j2
```diff
@@ -10,11 +9,12 @@
...
 use opentelemetry::{Key, KeyValue, StringValue};
 
+/// A constant indicating whether the attributes are enabled.
 pub const ATTRIBUTES_PARAM: bool = {{ params.attributes }};
 
 {% for attributes_root_namespace in ctx %}
-/// Attributes for the `{{ attributes_root_namespace.root_namespace }}` namespace.
-pub mod {{ attributes_root_namespace.root_namespace | snake_case }};
+//! Attributes for the `{{ attributes_root_namespace.root_namespace }}` namespace.
+//! pub mod {{ attributes_root_namespace.root_namespace | snake_case }};
```
lquerel, line 17: Sorry, I didn’t understand these changes. Why comment out this part of the code generation?

crates/weaver_codegen_test/templates/registry/rust/metrics/mod.rs.j2
```diff
@@ -8,11 +7,12 @@
...
 //! DO NOT EDIT, THIS FILE HAS BEEN GENERATED BY WEAVER
 
+/// A constant indicating whether the metrics are enabled.
 pub const METRICS_PARAM: bool = {{ params.metrics }};
 
 {% for attributes_root_namespace in ctx %}
-/// Metrics for the `{{ attributes_root_namespace.root_namespace }}` namespace.
-pub mod {{ attributes_root_namespace.root_namespace | snake_case }};
+//! Metrics for the `{{ attributes_root_namespace.root_namespace }}` namespace.
+//! pub mod {{ attributes_root_namespace.root_namespace | snake_case }};
```
lquerel, line 15: Same question.

lquerel (commented):
I missed something but I didn’t understand the changes to the code generation section. What was the issue exactly? Why comment out that part of the code generation?

crates/weaver_codegen_test/templates/registry/rust/attributes/mod.rs.j2
jerbly: Clippy was complaining about the documentation and formatting. The jinja template previously created:
```rust
/// Attributes for the `system` namespace.
pub mod system;
```
for example.

Then here: https://github.com/open-telemetry/weaver/blob/42203247a8217ef6909949764ab641956502e8ff/crates/weaver_codegen_test/build.rs#L84-L89 the `pub mod` lines are removed leaving a floating `///` doc comment.

crates/weaver_codegen_test/templates/registry/rust/metrics/mod.rs.j2
jerbly: Same answer ;)

crates/weaver_codegen_test/templates/registry/rust/attributes/mod.rs.j2
lquerel: Ah, ok. I had forgotten about that part of the plumbing in the code :-)

PR #500, opened 2024-12-04
Ensure docker image is built on every pull request to main.

jsuereth:
This is so we can enforce docker image builds successfully on PRs.

Given Docker is a primary mechanism of distributing weaver, we should guarantee it builds on PRs.


lquerel (approved):
LGTM

Issue #502, opened 2024-12-04
Support stability definitions from OTEP 232


lmolkova:
Part of https://github.com/open-telemetry/semantic-conventions/issues/1096

We need to include new maturity levels formally defined in [OTEP 232](https://github.com/open-telemetry/oteps/blob/main/text/0232-maturity-of-otel.md):

- Development - new one that replaces `experimental`. We should keep supporting `experimental`, but treat it as `developement`
- Alpha - new 
- Beta - new 
- Release Candidate - new 
- Stable - no change
- Deprecated - no change (not a stability level from the semconv schema perspective)
- Unmaintained - don't add for now

This would mean:
- extending the [stability enum](https://github.com/open-telemetry/weaver/blob/c7e27b243f09ada21e67d12226c03d8b0ca3ad76/crates/weaver_semconv/src/stability.rs#L12) to cover new members
- update [`is_experimental` filter](https://github.com/open-telemetry/weaver/blob/c7e27b243f09ada21e67d12226c03d8b0ca3ad76/crates/weaver_forge/src/extensions/otel.rs#L283) to cover all maturity levels (i.e. return `true` for all levels except `stable`)
- update [schema definition](https://github.com/open-telemetry/weaver/blob/c7e27b243f09ada21e67d12226c03d8b0ca3ad76/schemas/semconv.schema.json#L531) to list all new possible values

jerbly:
@lmolkova - I can do this one. Set me as the assignee if you like??

lmolkova:
@jerbly thank you, really appreciate your help!

jerbly:
Made #504 - hopefully this is what you had in mind. 🤞

Issue #481, opened 2024-11-27
Validate that Weaver is compatible with older versions of the registry


lquerel:
Currently, Weaver is compatible with registries from version 1.22 to the most recent one. However, there is no specific test dedicated to validating support for older versions. It would be beneficial to add a test modeled after the one that verifies the `registry check` command (see `tests/registry_check.rs`).

This test should execute the `registry check` command on each archive of the registries, starting from version 1.22 up to the latest. To avoid introducing external dependencies on content from GitHub in our tests, I propose copying these archives into a Weaver directory and performing the validation on these local archives (Weaver already supports this mode).

Of course, this means maintaining these archives within Weaver. If there’s a more dynamic way to achieve this without adding external dependencies, it should definitely be explored.

In any case, this test should fail if one or more registry checks fail. Ideally, the test should return the version numbers of all registries that could not be validated.

jerbly:
What is the need for being backwards compatible?

My understanding is that the schema defined in https://github.com/open-telemetry/weaver/blob/main/schemas/semconv-syntax.md is versioned via git tags/releases. To avoid polluting the code with a lot of backwards compatibility logic and tests, perhaps the user should use the appropriate tag/release of weaver for the version of schema they need?

If we take this approach, there is already quite a bit of code and tests which could be removed rather than maintained.

lquerel:
The main need for supporting multiple versions of the registry will become significantly more important when we support the multi-registry scenario (which is our next focus). In this context, it will be difficult to get everyone to agree on the schema versions of different registries. Let's say the beginning of multi-registry support corresponds to semconv registry 1.32. Subsequently, people start creating registries that extend 1.32 and others that extend 1.33. In this case, Weaver will need to be able to read these registries even if they follow different schemas.

jerbly:
I see... Perhaps this task would be simpler and more accurate if the standard included a schema version declaration at the top of each yaml semconv file. Then weaver could use this to ensure compliance with that schema version. This would allow for clear branches in the parsing code.

Without an explicit schema version declaration the parser has to "guess" compatibility, the code becomes more tangled and the compliance is weaker. In effect you're checking compliance against a range of versions.

jerbly:
@lquerel - I've made a draft PR #515 for this. Is this what you had in mind? Let me know what you think.

jerbly:
Implemented in #515

PR #504, opened 2024-12-05
BREAKING CHANGE: Support stability definitions from OTEP 232

jerbly:
BREAKING CHANGE:
* Issue [#502](https://github.com/open-telemetry/weaver/issues/502) - Support stability definitions from [OTEP 232](https://github.com/open-telemetry/oteps/blob/main/text/0232-maturity-of-otel.md)
  * Stability enum now has these variants: `stable`, `development`, `deprecated`, `alpha`, `beta`, `release_candidate`
  * `unmaintained` is not supported yet.
  * `experimental` is still accepted when parsing but aliased to `development`.
  * The minijinja test, `experimental`, now returns true for any variant other than `stable` and `deprecated`.
  * EBNF and JSON schema updated to define the new enum without the `experimental` variant.


crates/weaver_forge/templates/semconv_jq_fn/weaver.yaml
```diff
@@ -1,5 +1,5 @@
 params:
-  exclude_stability: ["experimental"]
+  exclude_stability: ["development"]
```
jsuereth, line 2: This may also be a breaking change.  Should we update the default JQ expressions to treat "experimental" as `!= "stable"` ?

crates/weaver_forge/src/extensions/otel.rs
```diff
@@ -277,15 +277,15 @@ pub(crate) fn is_stable(input: &Value) -> bool {
...
+/// Checks if the input value is an object with a field named "stability" that has any value
+/// other than "stable" or "deprecated". Otherwise, it returns false.
 #[must_use]
 pub(crate) fn is_experimental(input: &Value) -> bool {
     let result = input.get_attr("stability");
 
     if let Ok(stability) = result {
         if let Some(stability) = stability.as_str() {
-            return stability == "experimental";
+            return stability != "stable" && stability != "deprecated";
```
lmolkova, line 288: I think we have "deprecated" in stability levels here only for backward compatibility.

In semantic conventions, `deprecated` is an extra property that can coexist with stability.

We deprecate experimental attributes and there are things that are experimental and deprecated. And there will be stable and deprecated things at the same time.

So I think we should do

```suggestion
            return stability != "stable";
```

@lquerel do you think we can deprecate or remove `Deprecated` stability enum member?

https://github.com/open-telemetry/weaver/blob/7b70d9f8ce975f66957c5544e649b543a07fdc25/crates/weaver_semconv/src/stability.rs#L32

Semconv doesn't use it since [v1.24.0](https://github.com/open-telemetry/semantic-conventions/blob/106f880ccbf26443f115ce1f48c236ec6a0b6f1b/CHANGELOG.md?plain=1#L432) and weaver works with semconv starting 1.26.0

crates/weaver_forge/templates/semconv_jq_fn/weaver.yaml
lmolkova: it would be nice!

lmolkova:
I do agree with @jsuereth that we should call out it is breaking - I will join maintainers call on Monday and socialize it. When SIGs run codegen they hopefully should be able to easily detect it and the fix should be straightforward. 

There are also just a few repos affected - https://github.com/search?q=org%3Aopen-telemetry+exclude_stability+language%3AYAML&type=code&l=YAML

crates/weaver_forge/src/extensions/otel.rs
jsuereth: +1.  We should actually treat `stability="deprecated"` as stability="development" for the purpose of comparisons.  We made that decision back when we deprecated "deprecated" on stability level.

trask: > back when we deprecated "deprecated"

😹

jerbly: Seems like there's two things being debated here:
1. Backwards compatibility: How is compatibility defined? As someone relatively new to this it feels a bit like the tail is wagging the dog. The semantic conventions are versioned e.g. 1.24.0 and then the [schema definition](https://github.com/open-telemetry/weaver/blob/c7e27b243f09ada21e67d12226c03d8b0ca3ad76/schemas/semconv.schema.json) is updated. Perhaps the schema definition should be versioned and the semantic conventions declare the version they're using. I wrote about that [here](https://github.com/open-telemetry/weaver/issues/481#issuecomment-2507681351) too. We could then maintain up to, say, 3 prior versions for weaver backwards compatibility. This is particularly important for folks like me who define their own company-wide semantic conventions, we need to adhere to a schema version rather than figure it out from a release of the otel semconvs.
2. OTEP 232 compatibility: If we're trying to adhere to this standard then `deprecated` should be defined in the `stability` field. (Although arguably now `stability` should be renamed `maturity`). [Deprecated is defined in OTEP 232](https://github.com/open-telemetry/oteps/blob/main/text/0232-maturity-of-otel.md#deprecated). The `deprecated` field could become an optional `deprecation_message` instead.

But... I've not been around for all the history here so this is just my relatively newbie viewpoints. I'll wait for @lquerel  to respond to the question above re removing the deprecated variant and then make the changes.

jsuereth: 1. The question here is if folks start using the new weaver, do their workflows continue working or do they break.  Qualitativately this means:
  - Codegen continues working without change.
  - Rego Policies continues working without change.
  - weaver.yaml configuration continues working without change. (So JQ filter expressions work without change).
  
  Regarding supporting old/new schema versions that's an orthogonal but related concern to what I'm worried about here.  We *do* need to define a more formal "backwards compat" strategy with tests for which old versions we resolve.  However, for now, if we cut a new version we don't want existing usage to break, instead there should be a warning that becomes an error when passing `--future`.   For this PR I think it's acceptable for some breakage, as I don't see good alternatives otherwise.
  
  2. In Semantic conventions, we decided that `deprecated` should track the *previous stability level* of a component.  This is because we *allow* cleanup of non-stable deprecated fields but do not allow this for stable-then-deprecated fields.  This is an additional restriction on semantic conventions.  In practice, we are still specification compliant with how we annotate things due to our markdown templates.  Let's not mix specification compliance with architectural design of weaver.

jsuereth:
> I do agree with @jsuereth that we should call out it is breaking - I will join maintainers call on Monday and socialize it. When SIGs run codegen they hopefully should be able to easily detect it and the fix should be straightforward.
> 
> There are also just a few repos affected - https://github.com/search?q=org%3Aopen-telemetry+exclude_stability+language%3AYAML&type=code&l=YAML

I think we can fix that usage (either in this PR or a follow on). I was more worried about this: https://github.com/search?q=org%3Aopen-telemetry+%22.stability%22+language%3A%22Open+Policy+Agent%22&type=code

However, looking there, it's only used by semconv so far AND we only rely on the string `"stable"`, so I don't think this will break anything there.

@jerbly I'll send you a snippet for default JQ helpers if you don't get to this before tomorrow.

jsuereth:
@jerbly So, in `defaults/jq/semconv.jq`, you want to add this helper function:

```jq
# Expands stability array that previously used "experimental" for equivalent new strings.
def expand_stability($stability):
  if $stability | index("experimental") then
    $stability + [ "development", "alpha", "beta", "release_candidate" ] 
  else
    . 
  end;
```

Then modify the methods `semconv_signal` and `semconv_attributes` so anywhere you see `$options.exclude_stability` replace it with `expand_stability($options.exclude_stability)`.

This should ensure that codegen which splits stable from experimental continues to work.

jerbly:
@jsuereth - I made your change to the default semconv.jq (couldn't use `.` in the else clause for some reason??). I also reverted one of the filters in weaver.yaml to exercise this:
```yaml
- template: semconv_grouped_attributes_without_experimental.json
    filter: >
      semconv_grouped_attributes({
        "exclude_root_namespace": ["url", "network"], 
        "exclude_stability": ["experimental"]
      })
    application_mode: single
```

@lmolkova - committed your change so the experimental test means non-stable.

jsuereth:
> I made your change to the default semconv.jq (couldn't use . in the else clause for some reason??).

Ah right, the `.` is super contextual in jq.  The fix you made is right, sorry about that!

crates/weaver_forge/templates/semconv_jq_fn/weaver.yaml
jsuereth: Thanks for fixing this

jsuereth (approved):
Thanks so much for this @jerbly !

lquerel (approved):
LGTM. Thank you Jeremy!

lmolkova (approved):
Thank you!!!

PR #505, opened 2024-12-05
prepare for 0.11.0 release.

jsuereth:
Note: I updated the versions in crates to use the workspace version.  We can opt-out of using the global version on a per-crate basis if needed later.


lquerel (approved):
LGTM

PR #506, opened 2024-12-05
build: add cross-compile support

bernot-dev:
The build currently relies on QEMU emulation for arm64 builds (#346). [This is slow](https://blog.bernot.io/qemu).

This change uses Rust's ability to cross-compile for different platforms instead of relying on emulation, resulting in much faster builds.

Expected improvement is from [≈1h25m](https://github.com/open-telemetry/weaver/actions/runs/12184896455/job/33989858643) to <10m. Local cross-platform builds took about 3-7 minutes for me.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,2 +1,8 @@
 [alias]
 xtask = "run --package xtask --bin xtask --"
+
+[target.aarch64-unknown-linux-gnu]
+linker = "aarch64-linux-gnu-gcc"
+
+[target.x86_64-unknown-linux-gnu]
+linker = "x86_64-linux-gnu-gcc"
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,8 +1,8 @@
 [alias]
 xtask = "run --package xtask --bin xtask --"
 
-[target.aarch64-unknown-linux-gnu]
+[target.aarch64-unknown-linux-musl]
 linker = "aarch64-linux-gnu-gcc"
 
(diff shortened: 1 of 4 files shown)


.cargo/config.toml
```diff
@@ -1,2 +1,8 @@
 [alias]
-xtask = "run --package xtask --bin xtask --"
\ No newline at end of file
+xtask = "run --package xtask --bin xtask --"
+
+[target.aarch64-unknown-linux-gnu]
+linker = "aarch64-linux-gnu-gcc"
```
jsuereth, line 5: I don't think we want to convert MUSL builds to `libc` here.

Dockerfile
```diff
@@ -1,31 +1,33 @@
-
 # The build image
-FROM rust:1.83.0-alpine3.20 AS weaver-build
-RUN apk add musl-dev
```
jsuereth, line 4: Think we need to keep MUSL library here.

cross-arch-build.sh
```diff
@@ -0,0 +1,20 @@
+set -exu
+
+if [ "${TARGETPLATFORM}" = "linux/amd64" ]; then
+  RUST_TARGET=x86_64-unknown-linux-gnu
```
jsuereth, line 4: These targets should be right - these are the MUSL target for rust, which is right for alpine linux (vs. other linux distros that have legit libc)

jsuereth (approved):
Thanks so much for the fixes!!!

PR #508, opened 2024-12-06
Add test cases to cover a paragraph after a list

trask:
I'm planning to submit a (separate) PR (EDIT see #511) to add a (javadoc) `<p>` tag in between `</ul>` and following text, but first sending this PR to add this scenario to the test cases (with no underlying code changes), so that it's clearer what my fix changes and what it doesn't.

Here's the new semconv content that exposed this issue when updating to v1.29.0: https://github.com/open-telemetry/semantic-conventions-java/pull/118/files#r1872551505


crates/weaver_forge/src/formats/html.rs
```diff
@@ -550,9 +552,9 @@ If a specific domain defines its own set of error identifiers (such as HTTP or g
 it's RECOMMENDED to:
 <ul>
   <li>Use a domain-specific attribute
-  <li>Set {@code error.type} to capture all errors, regardless of whether they are defined within the domain-specific set or not.
+  <li>Set {@code error.type} to capture all errors, regardless of whether they are defined within the domain-specific set or not
 </ul>
-"##
+And something more."##
```
trask, line 557: it was confusing for me at first why it sometimes ended with a newline and sometimes not, which was part of the motivation for #509

lquerel (approved):
OK

Issue #501, opened 2024-12-04
Raise warnings when special tags are misspelled, or missing spaces


ChrsMark:
**Describe the bug**

A small "typo" like `<!-- semconv metric.k8s.hpa.min_pods-->` (missing space before `-->`) makes the tooling producing no markdown tables without raising a warning for this.

Raising a warning would definitely help on troubleshooting cause usually these kind of typos are hard to spot :).

**To Reproduce**

Use a metric tag like `<!-- semconv metric.k8s.hpa.min_pods-->` and try to generate the markdowns (`make fix`).

**Expected behavior**

At least a warning should let users know what wrong with that tag.

**Additional context**

ref: https://cloud-native.slack.com/archives/C041APFBYQP/p1733311696182559

jsuereth:
I looked into this, the issue is my parser isn't accurately pulling HTML comments and then parsing the content within it.

[link to code](https://github.com/open-telemetry/weaver/blob/main/crates/weaver_semconv_gen/src/parser.rs#L82)

Specifically, it's *requiring* space to parse the id.  That is, we're parsing:

- `<!--`: HTML header
- `semconv metric.k8s.hpa.min_pods-->`: Identifier
- ERROR!

I think this could be fixed by created a generic HTML comment parser that captures the content *between* `<--` and `-->`, and then feeding this to the `semconv` controls parser.

I'll try to get to this after my latest Go bug fixes unless someone else has time to address it.

yluobb:
Hello, I’m new to contributing to open source projects and I thought this would be a beginner-friendly issue, can I attempt this?

jsuereth:
@yluobb Feel free!  I did create this PR: #512 which doesn't solve the problem but should provide a good help in solving the problem.

yluobb:
@jsuereth  It looks like the changes in your PR works fine with the current tests. Could you provide a specific example or input where the current solution isn't working as expected? This will help me better understand the issue.

yluobb:
@jsuereth I added the mechanism to raise warning when there are missing spaces [here](https://github.com/open-telemetry/weaver/pull/516), please feel free to take a look, and please let me know what other fix I can help with.

PR #512, opened 2024-12-06
Create better HTML comment parser - Allow more semconv snippet headers.

jsuereth:
Towards: #501 

Create a more flexible HTML comment parser so we don't RELY on spaces to exist to succesfully parse HTML comments.


lquerel (approved):
LGTM

PR #515, opened 2024-12-07
Registry check against historic versions

jerbly:
Issue #481 - Added `cargo xtask history <start_version>` this gets the tag list from the semconv repo and then runs `registry check` on the matching archives. `start_version` defaults to `1.22.0`, semver is used to build the list of versions >= start_version.


lquerel:
I would prefer that the archived files for each version since 1.22 be placed in the `history` directory instead of the main repository as a whole. This would avoid having a collection of massive files in this commit. Weaver is capable of interpreting these zip archives, and it is also possible to specify a subdirectory within the archive to locate the YAML files (in our case, it’s `model`).

`https://github.com/open-telemetry/semantic-conventions/archive/refs/tags/v1.22.0.zip`
...
`https://github.com/open-telemetry/semantic-conventions/archive/refs/tags/v1.29.0.zip`

The command-line syntax to specify the subdirectory in the archive is: `https://github.com/open-telemetry/semantic-conventions/archive/refs/tags/v1.29.0.zip[model]`. But from the crate these 2 pieces of information are separated.

jerbly:
@lquerel - I changed this to use zip files. Rather than pull the whole archive, the script creates zips of the model dirs. This saves some space. For example the archive zip for 1.29.0 is 1.7MB whereas the zipped model dir is 184KB.

jsuereth:
For security reasons, I'm not a huge fan of accepting zipped files in a git repo.  They're opaque and easy for someone to slip something in later, even if they are safe right now.

Can we just move the historic check to be something that happens in its own cargo call (not `cargo test`) and create a dedicated github workflow which downloads and runs the check?  We can also not have this be a blocking check, to adjust for the possible flakiness of downloading.

lquerel:
I find the idea of having an independent workflow interesting. Regarding the security issue with zip files in a repository, I think there’s no reason to regret the initial approach of this PR, which involved including all the files for each semconv version. In terms of review, having thousands of files in a PR is just as opaque as a zip file.

jsuereth:
> I find the idea of having an independent workflow interesting. Regarding the security issue with zip files in a repository, I think there’s no reason to regret the initial approach of this PR, which involved including all the files for each semconv version. In terms of review, having thousands of files in a PR is just as opaque as a zip file.

Agreed.  Ideally we can get to a point where the code review is just "which versions are you checking".  @jerbly let us know if you have an idea how to solve making this a separate workflow.  I was thinking we could help brainstorm this live in the Semantic Convention Tooling SIG Wed.

jerbly:
I'm working on a `cargo xtask` to essentially perform what the `history/get.sh` script followed by the test did so it's outside the normal testing flow. (I'm new to gix though so the shell script was 100 times easier to write!) Would be good to talk through on Wednesday.

lquerel:
@jerbly For this xtask, you could use the GitHub API list-releases (see https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#list-releases) and then check the `zipball_url` field for each release to get the archive URL for each release. The other option is to use gix, as you mentioned. There is an example of how to use gix in the `weaver_cache` crate, which allows retrieving the content of a repository locally. But the difficulty here is to get the repo for a specific tag...

jerbly:
@lquerel - thanks for the suggestions. I got something going earlier today so I've pushed it here now as a PoC. It uses gix to do a fetch (no checkout) of the repo to query for the tags. Then I use these to construct the archive paths which weaver already supports.

An enhancement to weaver would be for it to support branches/tags when providing a repo URL. With that new feature, we could replace what I've done here with building the archive URL with something more robust.

The xtask uses semver to determine the versions you want to test against (defaulting to `1.22.0` if you don't supply a start version). It runs registry check against every version >= the start version you've supplied.

Here's an example output:
```
cargo xtask history 1.24.0
Checking versions: ["v1.24.0", "v1.25.0", "v1.26.0", "v1.27.0", "v1.28.0", "v1.29.0"]
Success for: v1.24.0
Success for: v1.25.0
Success for: v1.26.0
Success for: v1.27.0
Success for: v1.28.0
Success for: v1.29.0
```

jsuereth:
@jerbly This looks good to me.  Could you add the github action to perform the check in this PR as well?

lquerel:
LGTM too

jerbly:
@jsuereth - added to CI in the test job. (Looks good in the action output here.) Slight change to produce an error and therefore non-zero exit code if there's any failures. Also added to justfile in `pre-push-check`.

lquerel (approved):
LGTM. Thanks for this PR.

.github/workflows/ci.yml
```diff
@@ -46,6 +46,8 @@ jobs:
 #        run: cargo test --workspace --all-features
       - name: No-default features
         run: cargo test --workspace --no-default-features
+      - name: History check
```
jsuereth, line 49: Nit: A bit late to the party, but I don't think you need to run this in the test matrix.  I.e. we can just run this ONCE instead of on windows/osx/linux.

Otherwise LGTM - thanks for the fix!!!

PR #516, opened 2024-12-08
Create a more flexible HTML comment parser and raise warning

yluobb:
Towards: https://github.com/open-telemetry/weaver/issues/501

This PR builds on top of the work from [here](https://github.com/open-telemetry/weaver/pull/512). Create a more flexible HTML comment parser so we don't rely on spaces to exist to succesfully parse HTML comments. Also, raise warnings when there are missing spaces.

### DIFF
--- a/crates/weaver_semconv_gen/src/parser.rs
+++ b/crates/weaver_semconv_gen/src/parser.rs
@@ -90,6 +90,17 @@ fn parse_html_comment(input: &str) -> IResult<&str, &str> {
     let (input, _) = tag("<!--")(input)?;
     // Optionally, text, with the additional restriction that the text must not start with the string ">", nor start with the string "->", nor contain the strings "<!--", "-->", or "--!>", nor end with the string "<!-".
     let (input, result) = take_until("-->")(input)?;
+
+    // Check for spacing issues and warn if found
+    if result.starts_with("semconv") && !input.trim().starts_with("semconv ") {
+        // Missing space after "<!--"
+        eprintln!("Warning: Missing space after '<!--' in semconv comment: {}", result);
+    }
+    if result.ends_with("semconv") && !result.trim().ends_with(" semconv") {
+        // Missing space before "-->"
+        eprintln!("Warning: Missing space before '-->' in semconv comment: {}", result);
+    }
+
     // The string "-->".
     let (input, _) = tag("-->")(input)?;
     Ok((input, result))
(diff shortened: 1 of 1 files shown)


crates/weaver_semconv_gen/src/parser.rs
```diff
@@ -78,17 +83,37 @@ fn parse_id(input: &str) -> IResult<&str, &str> {
...
+/// nom parser for HTML comments: `<!--{comment}-->
+fn parse_html_comment(input: &str) -> IResult<&str, &str> {
+    // Comments must have the following format:
+    // The string "<!--".
     let (input, _) = tag("<!--")(input)?;
+    // Optionally, text, with the additional restriction that the text must not start with the string ">", nor start with the string "->", nor contain the strings "<!--", "-->", or "--!>", nor end with the string "<!-".
+    let (input, result) = take_until("-->")(input)?;
+
+    // Check for spacing issues and warn if found
+    if result.starts_with("semconv") && !input.trim().starts_with("semconv ") {
```
jsuereth, line 95: Three things:

1. I don't think we *need* to force spaces here.

2. For the purpose of separation of concerns, the validation logic should go in `parse_semconv_snippet_directive`.  Ideally this function JUST parses HTML headers, which have no restrictions on whitespaces.  

3. We're not using `eprintln!` for warnings, instead you'll want to find a way to return a `Diagnostic`, e.g. using weaver's result [WResult](https://github.com/open-telemetry/weaver/blob/main/crates/weaver_common/src/result.rs#L17).  This might require some wiring to make sure non-fatal-errors (NFE) can make it out of the parser.  I haven't tried to sort out NFE w/ the nom parsers bits yet.  Let me know if you have issues I'll see if I can put a scaffold up for you there.

jsuereth:
A quick thought on warnings here:

1. You can update the [`update_markdown_contents`](https://github.com/open-telemetry/weaver/blob/main/crates/weaver_semconv_gen/src/lib.rs#L200) function to use weaver result.
2. You can update the logic to something like:

```rust
  if parser::is_markdown_snippet_directive(line) {
    ... existing ...
  } else if parser::is_possible_markdown_snippet_directive(line) {
     ... issue warnings ...
 }
```

Where `is_possible_markdown_snippet_directive` is a function you can write re-using the HTML comment parser with some kind of very flexible mispelling finding parser.

lquerel:
I'm closing this PR since it has been inactive for over a year.

PR #517, opened 2024-12-08
Fixed rego typos, attrigute and deprecaded

jerbly:
I noticed that in the examples and tests for rego there were two repeated typos:
* `attrigute` -> `attribute`
* `deprecaded` -> `deprecated`


lquerel (approved):
Thanks!

PR #524, opened 2024-12-14
Moving to JAQ 2.0

jsuereth:
- Update Filter to only remember strings, as we can't safely precompile with new lifetime shenanigans.
- Create new jq file to hide all details of calling JQ
- Update filter/engine to start passing serde-JSON structures by default, instead of JAQ internals.
- Create error-message generation.
- Update tests.

### DIFF
--- a/crates/weaver_forge/src/filter.rs
+++ b/crates/weaver_forge/src/filter.rs
@@ -4,14 +4,11 @@
 
 use crate::error::Error;
 use core::fmt;
-use jaq_interpret::{Ctx, FilterT, RcIter, Val};
-use jaq_syn::Def;
-use std::fmt::Debug;
+use std::{collections::BTreeMap, fmt::Debug};
 
 /// A filter that can be applied to a JSON value.
 pub struct Filter {
     filter_expr: String,
-    filter: jaq_interpret::Filter,
 }
 
 impl Filter {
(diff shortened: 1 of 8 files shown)


crates/weaver_forge/src/filter.rs
```diff
@@ -4,87 +4,31 @@
...
-                Err(e) => errs.push(e),
-            }
-        }
-
-        if values.len() == 1 {
-            return Ok(values.pop().expect("values.len() == 1, should not happen"));
-        }
-
-        Ok(serde_json::Value::Array(values))
+        crate::jq::execute_jq(&ctx, &self.filter_expr, values)
```
lquerel, line 31: Happy to see the updated signature of the `execute_jq` function in minijinja with the BTreeMap. It’s a much better!

crates/weaver_forge/src/jq.rs
```diff
@@ -0,0 +1,221 @@
...
+    let modules = loader
+        .load(&arena, program)
+        .map_err(load_errors)
+        .map_err(|e| Error::FilterError {
+            filter: filter_expr.to_owned(),
+            error: e,
+        })?;
+
+    let (names, values) = prepare_jq_context(params);
+    let funs = jaq_std::funs().chain(jaq_json::funs());
```
lquerel, line 61: nit: I’m not entirely sure, but you could try something like this to make the intent more explicit.

```
fn adjust_lifetime<'a, T: 'a>(iter: impl Iterator<Item = T>) -> impl Iterator<Item = T> + 'a {
    iter
}
 
let funs = adjust_lifetime(jaq_std::funs().chain(jaq_json::funs()));
...
  .with_funs(funs)
  .compile(modules)
  ...
```

lquerel (approved):
LGTM. Just two nit comments. Thank you, that was a significant effort to transition to this new JAQ version.

crates/weaver_forge/src/filter.rs
```diff
@@ -4,87 +4,31 @@
...
 impl Filter {
     /// Create a new filter from a string expression or return an error if the
     /// expression is invalid.
     /// The vars parameter is a list of variable names that can be used in the
     /// filter expression.
-    pub fn try_new(filter_expr: &str, vars: Vec<String>, defs: Vec<Def>) -> Result<Self, Error> {
-        let mut ctx = jaq_interpret::ParseCtx::new(vars);
-        ctx.insert_natives(jaq_core::core());
-        ctx.insert_defs(jaq_std::std());
-        ctx.insert_defs(defs);
```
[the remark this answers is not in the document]
jsuereth: I believe we are.  I'm pulling in the jaq_std defa and json defa

crates/weaver_forge/src/jq.rs
jsuereth: unfortunately, that function doesn't work (`iter` may not live for lifetime of `'a`).  What we need is to trick the compiler to infer a narrower lifetime when seeing both `'a` and `'static` to return `'a`, and we can only do that when they are in contravariant (or return-value) position.  I haven't figured out how to do that yet, but I can look that up for future cleanup. I figure it's not important enough to block this PR.

jsuereth: So i think I have my arrows backwards (using contravariant to mean covariant), but otherwise (while I didn't get a chance to fully read this), the answer probably lies here: https://doc.rust-lang.org/nomicon/subtyping.html

From what I can tell the variance rules are only used to change lifetimes within rust, the trick I'm using with lambda is the only way I could think of.

Issue #423, opened 2024-10-21
Dump a value into YAML


lquerel:
We already have a `tojson` filter to output a value in JSON format, but we don’t have a filter for YAML output. Some people have expressed this need.

unpervertedkid:
@lquerel , can I pick this up?

lquerel:
@unpervertedkid Yes. Thank you

unpervertedkid:
@lquerel , for me to understand how to properly go on with this, kindly help point me on the following:
1. An example of usage of the current tojson filter or how I can create a simple example usage to get the gist of the current implementation.
2. I see that there is a filter in [`weaver_forge/src/filter.rs`](https://github.com/open-telemetry/weaver/blob/main/crates/weaver_forge/src/filter.rs). Is that the toJson filter we are referring to?

If you have any other pointers in the docs or relevant information to get me started on this would be really appreciated.
Thanks.

lquerel:
@unpervertedkid You can add the `toyaml` filter in `crates/weaver_forge/src/extensions/util.rs`. The signature of the filter should be something like `fn to_yaml(value: Value) -> Result<Value, minijinja::Error>`. We also have to register this function in the `add_filters` (same file) with something like `env.add_filter("toyaml", to_yaml)`. In the repo of `minijinja` there is the implementation of the `tojson` filter so you can probably adapt the code from there. See https://github.com/mitsuhiko/minijinja/blob/4249fb3f423eb076317db542f08a00a1fcec39c3/minijinja/src/filters.rs#L1059

You also have to add a unit test. Thanks

unpervertedkid:
@lquerel , is it important to preserve ordering when converting to yaml?

PR #525, opened 2024-12-14
Add a to_yaml filter

unpervertedkid:
# Need
People have expressed a need to dump a value into yaml file. #423 

# Solution

WIP: This pull request adds a `to_yaml` filter that can be used to dump a value into a YAML file.

# Known limitation
The implementation does not currently maintain ordering

### DIFF
--- a/crates/weaver_forge/src/extensions/util.rs
+++ b/crates/weaver_forge/src/extensions/util.rs
@@ -19,6 +19,7 @@ pub(crate) fn add_filters(env: &mut Environment<'_>, target_config: &WeaverConfi
     env.add_filter("flatten", flatten);
     env.add_filter("split_id", split_id);
     env.add_filter("regex_replace", regex_replace);
+    env.add_filter("to_yaml", to_yaml);
 }
 
 /// Add utility functions to the environment.
@@ -131,6 +132,16 @@ pub fn acronym(acronyms: Vec<String>) -> impl Fn(&str) -> String {
     }
 }
 
+fn to_yaml(value: Value) -> Result<String, minijinja::Error> {
+    let value = serde_json::to_value(value)
+        .map_err(|error| minijinja::Error::new(ErrorKind::BadSerialization, error.to_string()))?;
+
+    let yaml = serde_yaml::to_string(&value)
+        .map_err(|error| minijinja::Error::new(ErrorKind::BadSerialization, error.to_string()))?;
+
+    Ok(yaml)
+}
+
 #[cfg(test)]
 mod tests {
     use crate::extensions::util::add_filters;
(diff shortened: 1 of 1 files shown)


crates/weaver_forge/src/extensions/util.rs
```diff
@@ -131,6 +132,16 @@ pub fn acronym(acronyms: Vec<String>) -> impl Fn(&str) -> String {
     }
 }
 
+fn to_yaml(value: Value) -> Result<String, minijinja::Error> {
+    let value = serde_json::to_value(value)
+        .map_err(|error| minijinja::Error::new(ErrorKind::BadSerialization, error.to_string()))?;
```
jerbly, line 137: This is unnecessary, you can pass a reference to `minijinja::Value` directly to `serde_yaml::to_string` because it implements `serde::Serialize`.

crates/weaver_forge/src/extensions/util.rs
```diff
@@ -131,6 +132,16 @@ pub fn acronym(acronyms: Vec<String>) -> impl Fn(&str) -> String {
...
+
+    let yaml = serde_yaml::to_string(&value)
+        .map_err(|error| minijinja::Error::new(ErrorKind::BadSerialization, error.to_string()))?;
```
jerbly, line 140: Perhaps there's a way to write a test to exercise the error condition?

jerbly (changes requested):
Remember to also add a changelog entry. Thanks.

crates/weaver_forge/src/extensions/util.rs
```diff
@@ -131,6 +132,16 @@ pub fn acronym(acronyms: Vec<String>) -> impl Fn(&str) -> String {
     }
 }
 
+fn to_yaml(value: Value) -> Result<String, minijinja::Error> {
```
lquerel, line 135: You should take the `tojson` filter implemented in the standard `minijinja` library as an example. The method’s signature is slightly different, and more importantly, they address the compatibility of the output with HTML (pay attention to the last map function). We don’t need the pretty-printing part since YAML is always pretty-printed and we probably don't need the indent parameter too.

See the permalink below.

https://github.com/mitsuhiko/minijinja/blob/d1ce496e101239bdfba01e29f407f0a57b8e4aed/minijinja/src/filters.rs#L1018

lquerel (changes requested):

crates/weaver_forge/src/extensions/util.rs
lquerel: @unpervertedkid Would you like to continue contributing to this PR, or should we take over?

PR #536, opened 2025-01-04
Refactor: common resolve registry with policy checks

jerbly:
- Refactored CLI registry commands to remove some duplication. Resolving the registry with policy checks is common for `generate`, `resolve` and `check`.
- Added missing `after_resolution` policy checks to `generate` and `resolve` through the common code.
- Removed duplication in CLI args to maintain consistency.
- Removed the deprecated `--registry-git-sub-dir` option.
- Fixed bug in `check` if `--skip-policies` was specified then it would not fail for any validation errors. e.g. this fails `registry check -r "https://github.com/open-telemetry/semantic-conventions/archive/refs/tags/v1.27.0.zip[model]" --future` but this would pass `registry check -r "https://github.com/open-telemetry/semantic-conventions/archive/refs/tags/v1.27.0.zip[model]" --future --skip-policies`

### DIFF
--- a/src/registry/check.rs
+++ b/src/registry/check.rs
@@ -15,7 +15,7 @@ use weaver_semconv::registry::SemConvRegistry;
 
 use crate::registry::{CommonRegistryArgs, RegistryArgs};
 use crate::util::{
-    check_policy, check_policy_stage, init_policy_engine, load_semconv_specs, resolve_semconv_specs,
+    check_policy_stage, load_semconv_specs, prepare_main_registry, resolve_semconv_specs,
 };
 use crate::{DiagnosticArgs, ExitDirectives};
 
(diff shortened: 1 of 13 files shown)


src/registry/check.rs
```diff
@@ -80,100 +60,19 @@ pub(crate) fn command(
...
             // against it as a "baseline".
-            load_semconv_specs(
-                repo,
-                logger.clone(),
-                args.common_registry_args.follow_symlinks,
-            )
-            .ignore(|e| matches!(e.severity(), Some(miette::Severity::Warning)))
-            .capture_non_fatal_errors(&mut diag_msgs)
+            load_semconv_specs(repo, logger.clone(), args.registry.follow_symlinks)
+                .ignore(|e| matches!(e.severity(), Some(miette::Severity::Warning)))
```
jsuereth, line 70: I believe this should actually fail when `--future` is enabled.  I forget how we handled `--future`, so @lquerel can confirm/answer questions if you need, just want to make sure that case is handled.

jsuereth (commented):
Overall, really great cleanup!!! Just want to check on some functionality that may have also not been consistent across commands.

src/registry/check.rs
jerbly: The current behaviour for the baseline registry (unchanged by this refactor) is to ignore all warnings, this includes those that would be promoted to Errors with `--future`. To me this makes sense for the baseline registry since it's likely older and therefore may now fail newer validations.

jsuereth: Ah, I missed this was loading the baseline.  Thanks for clarifying!

lquerel:
Great cleanup, and thank you for doing that!

PR #540, opened 2025-01-08
Add is_array filter and test for AttributeType

arthurDiff:
Added `is_array`  fn to AttributeType enum.


lquerel:
@arthurDiff Thank you for this PR, but I believe this ticket unfortunately lacked detail. What @lmolkova originally wanted was a Jinja filter or test integrated into our template engine based on the Minijinja implementation.

The idea would be to replace this Jinja hack (see the original pb description [here](https://github.com/open-telemetry/semantic-conventions/pull/1088#discussion_r1619270097)):

```jinja
{%- if "[]" in attribute.type and "template" not in attribute.type %}
...
```

with a Jinja test similar to this declaration:

```jinja
{%- if attribute is array %}
...
```

You can find several examples of such tests in the following file (the one you could update):

[crates/weaver_forge/src/extensions/otel.rs](https://github.com/open-telemetry/weaver/blob/main/crates/weaver_forge/src/extensions/otel.rs)

I would be happy to provide more details or any kind of help if needed.

arthurDiff:
> @arthurDiff Thank you for this PR, but I believe this ticket unfortunately lacked detail. What @lmolkova originally wanted was a Jinja filter or test integrated into our template engine based on the Minijinja implementation.
> 
> The idea would be to replace this Jinja hack (see the original pb description [here](https://github.com/open-telemetry/semantic-conventions/pull/1088#discussion_r1619270097)):
> 
> ```
> {%- if "[]" in attribute.type and "template" not in attribute.type %}
> ...
> ```
> 
> with a Jinja test similar to this declaration:
> 
> ```
> {%- if attribute is array %}
> ...
> ```
> 
> You can find several examples of such tests in the following file (the one you could update):
> 
> [crates/weaver_forge/src/extensions/otel.rs](https://github.com/open-telemetry/weaver/blob/main/crates/weaver_forge/src/extensions/otel.rs)
> 
> I would be happy to provide more details or any kind of help if needed.

Hi @lquerel ! Sorry about the delay and thanks for the clarification!  Should be implemented with minijinja now.

crates/weaver_forge/src/extensions/otel.rs
```diff
@@ -454,6 +455,24 @@ pub(crate) fn is_enum(attr: &Value) -> bool {
...
+        attr_type,
+        "string[]"
+            | "int[]"
+            | "double[]"
+            | "boolean[]"
+            | "template[string[]]"
+            | "template[int[]]"
+            | "template[double[]]"
+            | "template[boolean[]]"
+    )
```
lquerel, line 473: @lmolkova This test looks fine to me, but could you confirm: for `semconvs`, should a `template[xxx[]]` be considered an array, or should we treat any template as not being an array, even when templating an array of something?

lquerel (changes requested):
Overall LGTM. Just 2 comments:
- I asked Liudmila to confirm if templates of arrays must be considered arrays. Nothing to do on your side for now.
- You should add a description of the Jinja test in the following [documentation](https://github.com/open-telemetry/weaver/blob/main/crates/weaver_forge/README.md#jinja-tests-reference).

After we should be ready to go :-).

Thank you

arthurDiff:
> Overall LGTM. Just 2 comments:
> 
> * I asked Liudmila to confirm if templates of arrays must be considered arrays. Nothing to do on your side for now.
> * You should add a description of the Jinja test in the following [documentation](https://github.com/open-telemetry/weaver/blob/main/crates/weaver_forge/README.md#jinja-tests-reference).
> 
> After we should be ready to go :-).
> 
> Thank you

Sounds good! Out of curiosity. is there reason for match string weaver's using literal string instead of static const? for example like this?
```

**Comment by lquerel** [comment]
@arthurDiff Regarding your question about strings, I’m not sure I fully understand. Are you suggesting introducing constants instead of using literal strings directly in the code? If so, then yes, they could indeed be replaced with `static const &str`.

Additionally, I didn’t understand why you closed this PR after making changes. It’s probably an error, so I'm reopening it so that we can proceed with the merge once we have received a response from @lmolkova. Thanks

File: crates/weaver_forge/src/extensions/otel.rs (hunk 1 above)
Comment (lmolkova) [reply crates/weaver_forge/src/extensions/otel.rs:473]: thanks for checking! Templates are not really arrays, they are more like a maps - we also have a filter that checks if attribute has a template - `is_template_type`, so I believe `is_array` should be limited to 

```
"string[]"
            | "int[]"
            | "double[]"
            | "boolean[]"
```

lmolkova: other that that, it looks great to me, thank you!

arthurDiff: Gotcha! thanks for the clarification! I just updated to those criteria!

lmolkova (approved):
Thank you, this looks great!

lquerel (approved):
LGTM

lquerel (approved):
LGTM

PR #548, opened 2025-01-11
Add an OTLP receiver to Weaver to prepare for the `weaver registry live-check` command

lquerel:
This PR aims to facilitate the creation of commands such as validating the compliance of an OTLP stream against a semconv registry or generating a semconv registry from an OTLP stream.

Weaver is already used in the context of a CI/CD pipeline, this is simply about extending the possibilities for validation or semconv registry generation in a more live context.

An `otlp-receiver` can now easily be launched within Weaver to implement these commands or any other commands that make sense in this context. A skeleton of the command `weaver registry live-check` is provided to help future contribution.

Notes:
- I vendored the gRPC stubs to avoid a dependency on protoc in our CI/CD pipeline. There is a comment in the `build.rs` explaining to regenerate the vendored gRPC stubs. Most of the files presents in this PR are either proto files or code generated from those proto files.
- Both client and server gRPC stubs are generated. The client stubs are used for test purposes and the server stubs are used to implement the OTLP listener.

Next Step: 
Someone will implement the `weaver registry live-check` command. A great example of this approach was demonstrated by @jerbly during the Semantic Convention Tooling SIG on January 15, 2025.


src/registry/otlp/mod.rs
```diff
@@ -0,0 +1,607 @@
...
+                    inactivity_timeout,
+                    &mut tasks,
+                );
+
+                // Serve the OTLP services
+                let server = Server::builder()
+                    .add_service(LogsServiceServer::new(logs_service))
+                    .add_service(MetricsServiceServer::new(metrics_service))
+                    .add_service(TraceServiceServer::new(trace_service))
+                    .serve(addr)
```
jerbly, line 225: If something is already listening on `addr` there should be bind error here. I'm not getting this in my testing - it just silently fails to bind.

lquerel: Missed this one. I will fix that.

lquerel: After taking a closer look at the code, I do have error handling in the statement that follows. This statement involves sending the error to the channel, which is exposed as an iterator on the consumer side. The fact that you don’t see an error message is probably due to Weaver (or the test) stopping before the consumer has had time to process the error. 

I’ll fix that.

lquerel: Fixed.

jerbly: That doesn't actually fix my problem, but I know why. I didn't spot that you're binding on `0.0.0.0:4317` I have another process bound on `127.0.0.1:4317`. So the second process is successfully binding on other interfaces, just not loopback. Then I send to `127.0.0.1:4317` and of course don't receive. Perhaps the fix here is to allow the user to enter the whole address not just the port - that's probably better for security anyway.

jerbly (approved):
Amazing! This opens up a whole load of new use cases for weaver! :)

build.rs
```diff
@@ -0,0 +1,28 @@
...
+    //     .compile_protos(
+    //         &[
+    //             "src/registry/otlp/proto/opentelemetry/proto/collector/logs/v1/logs_service.proto",
+    //             "src/registry/otlp/proto/opentelemetry/proto/collector/metrics/v1/metrics_service.proto",
+    //             "src/registry/otlp/proto/opentelemetry/proto/collector/trace/v1/trace_service.proto",
+    //         ],
+    //         &["src/registry/otlp/proto"],
+    //     )?;
+
+    Ok(())
```
jsuereth, line 27: Why is this all commented out?

I'm a bit nervous how we pull this in now, would it make more sense to depend on the otel-rust artifact for these?

lquerel: I probably did a poor job explaining the comments above this commented code. The general idea is that to integrate the gRPC stubs for OTLP, we have two options: 1) generate the stubs at every build using tonic-build, which wraps protoc, or 2) generate them once and version the generated code, thereby eliminating the need to depend on protoc in our GitHub pipeline. I chose the vendoring option to have a faster build and avoid this dependency in our build process. Option 1 is not a huge problem either, so if you have a strong preference for that approach, I’m open to it.

Regarding the otel-rust artifact, the issue is that I mainly need the OTLP server part, and to my knowledge, the `opentelemetry-rust` crate only includes the client part. If that’s not the case, I’d be happy to migrate to that crate. For the OTLP receiver tests, I had the choice between using opentelemetry-rust (otlp + sdk) or directly using the client stubs. I went with the latter to reduce dependencies. This decision was made before noticing that Jeremy is using these crates in the other PR currently under review. I can make this change if necessary, but the idea was to minimize dependencies.

Issue #564, opened 2025-01-22
Require attributes and event fields to have stability


lmolkova:
See https://github.com/open-telemetry/semantic-conventions/issues/1777 - we don't validate that stability is present on attributes or event fields

jerbly:
@lmolkova looks like we're also not enforcing `stability` on enum members of event fields: https://github.com/open-telemetry/semantic-conventions/blob/main/model/device/events.yaml#L36 

I assume we want this fixed too?

lmolkova:
@jerbly great catch! yes, explicit stability everywhere it matters is the goal :)

jerbly:
At the moment the Weaver code, EBNF and semconv.schema.json all define the `body` of the event as an `any_value`. Some `any_value`s can contain `any_value`s. Therefore the `body` itself is required to have `stability` defined. This is missing in every use of `body` in the current semconvs.
```yaml
    body:
      id: device_lifecycle_state
      type: map
      requirement_level: required
      stability: experimental # <-------------- This is also required
      fields:
        - id: ios.state
          stability: experimental
          type: enum
          ...
```

PR #568, opened 2025-01-23
Stability on attributes and events

jerbly:
For issue #564 

Added warnings for missing stability on:

- Attributes
- Enum members in attributes
- Event AnyValues
- Enum members in AnyValues


schemas/semconv.schema.json
```diff
@@ -209,7 +209,8 @@
 			"required": [
 				"type",
 				"id",
-				"requirement_level"
+				"requirement_level",
+				"stability"
```
lmolkova, line 213: I can't come up with the example where the stability of the body type would be different from event stability today. 

In the long term though we will need to be able to define a type and then reuse it across different event bodies and attributes.

E.g. if we have a complex attribute/type for geo location, then we could have `client.geo.location = {"lat": 42, "lon": 42 }` and  `user.geo.location = {"lat": 24, "lon": 24 }` which share the same type, but attribute definitions may have different stability levels.  

So it might look excessive today to provide stability on every step when defining event body, but I don't mind it

lquerel (approved):
LGTM. Thank you so much!

schemas/semconv.schema.json
jerbly: I agree, it's more consistent this way. If body is not `map` then it's not excessive.

lmolkova (approved):
Schema changes LGTM!

jsuereth (approved):

Issue #569, opened 2025-01-24
Add `include_stability` config into `semconv_grouped_attributes`


lmolkova:
We support `exclude_stability` but it's proven to be dangerous. The key scenario behind it is separating stable from non-stable signals.

`exclude_stability: "experimental"` is more prone to bugs (e.g. excluding "development" would bring a  bunch of `rc` attributes and does not really make sense). What happened in https://github.com/open-telemetry/semantic-conventions/issues/1777 is that we let attribute without stability in and it appeared as stable in Python and C++ since it was not experimental either. 

If we used `include_stability: ["stable"]` it would be more robust. 

We should consider going even further and just provide `stable_only = true|false` parameter.

lmolkova:
Great suggestion from @jerbly in the chat

> Second hole plug suggestion: We could change the logic of is_experimental to return true by default: https://github.com/open-telemetry/weaver/blob/main/crates/weaver_forge/src/extensions/otel.rs#L280-L292 (edited)

PR #570, opened 2025-01-24
`is_experimental` returns `true` by default

jerbly:
For issue #569 - Add include_stability config into semconv_grouped_attributes: `is_experimental` returns `true` by default.

### DIFF
--- a/crates/weaver_codegen_test/semconv_registry/registry/system.yaml
+++ b/crates/weaver_codegen_test/semconv_registry/registry/system.yaml
@@ -16,6 +16,7 @@ groups:
               value: 'nice'
             - id: idle
               value: 'idle'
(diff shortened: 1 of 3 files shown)


crates/weaver_codegen_test/semconv_registry/registry/system.yaml
```diff
@@ -16,6 +16,7 @@ groups:
               value: 'nice'
             - id: idle
               value: 'idle'
+              stability: stable
```
jerbly, line 19: Note. For our internal codegen test to pass I now had to explicitly set this to stable!

lmolkova (approved):
Thank you!

lquerel (approved):
LGTM

PR #591, opened 2025-02-07
simplify the span exporter

jerbly:
This simplifies the span exporter to remove the atomic lock on shutdown. This is to stop a flaky test.


lquerel (approved):
As a short-term solution, it seems okay to me for now. In the long term, the weaver-to-weaver emit->live-check solution indeed seems more interesting.

Issue #596, opened 2025-02-12
is_deprecated filter in JINJA does not respect new structure for deprecated


jsuereth:
TODO - fill out with trask's report

PR #597, opened 2025-02-12
fix(forge): Fix the Jinja test `is deprecated`.

lquerel:
This issue fixes the problem introduced by the change to the `deprecated` field, which transitioned from a simple string to a struct. The Jinja test `is deprecated` previously expected a string, it now checks for the presence of a struct.  

Note: I didn't need to modify the JQ filter `exclude_deprecated` since it was already checking for the presence or absence of a value.
 
Closes: https://github.com/open-telemetry/weaver/issues/596


jerbly (approved):
LGTM

PR #602, opened 2025-02-13
fix(semconv): Add a field note to the deprecated struct.

lquerel:
This PR adds a `note` field to all deprecated variants.
The content of the field `deprecated.note` depends on the nature of the `deprecated` field.
- For a `deprecated` field of type string, the generated `deprecated.note` field contains the text value of the `deprecated` field and the `reason` is `uncategorized`.
- For a structured `deprecated` field, the value of the `deprecated.note` depends on the following conditions:
  - If the `deprecated.note` field is defined then we keep it as is.
  - If the `deprecated.note` field is not defined
    - If the `deprecated.reason` is `renamed` then the `deprecated.note` is initialized with "Replaced by `<renamed_to>`."
    - If the `deprecated.reason` is `obsoleted` then the `deprecated.note` is initialized with "Obsoleted."
    - If the `deprecated.reason` is `uncategorized` then the `deprecated.note` is initialized with "Uncategorized."

Applying the Jinja expression `{{ attr.deprecated.note }}` to the following yaml attribute...

```yaml
id: attr
type: string
brief: This attribute is replaced by new_name.
note: A note.
requirement_level: required
deprecated: Replaced by new_name.
```

... will produce this output "Replaced by new_name." and `deprecated.reason` will be equals to `uncategorized`.

Applying the Jinja expression `{{ attr.deprecated.note }}` to the following yaml attribute...

```yaml
id: attr
type: string
brief: This attribute is replaced by new_name.
note: A note.
requirement_level: required
deprecated:
    reason: renamed
    renamed_to: new_name
```

... will produce this output "Replaced by 'new_name'.".

Applying the Jinja expression `{{ attr.deprecated.note }}` to the following yaml attribute...

```yaml
id: attr
type: string
brief: This attribute is replaced by new_name.
note: A note.
requirement_level: required
deprecated:
    reason: renamed
    renamed_to: new_name
    note: This attribute is deprecated and replaced by 'new_name'.
```

... will produce this output "This attribute is deprecated and replaced by 'new_name'.".


crates/weaver_resolved_schema/src/lib.rs
```diff
@@ -299,32 +299,32 @@ impl ResolvedTelemetrySchema {
...
-                            ..
+                            note,
                         } => {
                             changes.add_change(
                                 SchemaItemType::RegistryAttributes,
                                 SchemaItemChange::Renamed {
                                     old_name: baseline_attr.name.clone(),
                                     new_name: rename_to.clone(),
-                                    note: attr.brief.clone(),
+                                    note: note.clone().unwrap_or_else(|| attr.brief.clone()),
```
lmolkova, line 309: We don't need to fall back to the attribute brief if we have a note. 

For existing versions of semconv, we just need to use deprecated string value. 

For new conventions we only have a need it for `uncategorized` changes. We should generate the rest

```j2
{% macro deprecated_note(attribute) %}
{% if attribute.deprecated.reason == "renamed" %}
    {{ "Replaced by `" ~ attribute.deprecated.renamed_to | trim ~ "`." }}
{% elif attribute.deprecated.reason == "obsoleted" %}
    {{ "Removed." }}
{% else %}
    {{ attribute.brief }}
{% endif %}
{% endmacro %}
```

If we end up adding an optional note everywhere, we still want a default note to be generated (let's not ask everyone to write jinja for it).

E.g. we'd do

```j2
{% macro deprecated_note(attribute) %}
{% if attribute.deprecated.note %}
    {{ attribute.deprecated.note }}
{% else %}
    {% if attribute.deprecated.reason == "renamed" %}
      {{ "Replaced by `" ~ attribute.deprecated.renamed_to | trim ~ "`." }}
    {% elif attribute.deprecated.reason == "obsoleted" %}
      {{ "Removed." }}
    {% else %}
      {{ attribute.deprecated.note }}
    {% endif %}
{% endif %}
{% endmacro %}
```

----

So, if you'd prefer to add note for all change types, the suggestion is:
- do not fall back to the attribute brief
- if note is not provided, create a default one following the pattern above. We'd need to define a default note for each future transformation


I.e. the codegen/semconv md would be able to do something like this:

```j2
{% if attribute is deprecated %}
  {{ attribute.deprecated.note}}
{% endif %}
```

E.g. for [otel-rust](https://github.com/open-telemetry/opentelemetry-rust/blob/05920759303358cb979b7b5538da2b6b5c0be8ee/opentelemetry-semantic-conventions/scripts/templates/registry/rust/attribute.rs.j2#L21
), the immediate change would be

```diff
-#[deprecated(note="{{ attr.deprecated.strip(" \n\"") }}")]
+#[deprecated(note="{{ attr.deprecated.note.strip(" \n\"") }}")]
```

(and they can always customize the default by exploring `attr.deprecated`)

jerbly (commented):
The new `note` field needs to go in the json schema: https://github.com/open-telemetry/weaver/blob/82d56e5f0b57bd25616e33c24a3da7326f549891/schemas/semconv.schema.json#L44 and the syntax doc: https://github.com/open-telemetry/weaver/blob/82d56e5f0b57bd25616e33c24a3da7326f549891/schemas/semconv-syntax.md?plain=1#L69

crates/weaver_resolved_schema/src/lib.rs
lquerel: > We don't need to fall back to the attribute brief if we have a note.

I'm not. I only fall back to the`brief` field if there is no note.
Now if you prefer, I can remove the entire fall back mechanism even when there is no `deprecated.note`. I'm waiting for your feedback before to make any change.

> For new conventions we only have a need it for uncategorized changes. We should generate the rest

I might have misinterpreted the example you gave in the Slack channel, but I think your example demonstrated the opposite. You requested a note for something that wasn’t `uncategorized` but rather `renamed`. Based on this example, it seems like we need to support `note` for more than just `uncategorized` changes. Did I miss something?

lmolkova: I want us to format a note if it was not provided.
- For `renamed` it would be `f"Renamed to `{renamed_to}`"`
- For `obsoleted` it would be `Removed` (or `Obsoleted`)
- For uncategorized it would be `Removed`

I.e. resolved `deprecated.note` would come from (in the priority it's listed)
1. original string value of `deprecated` property if it was provided
2. `deprecated.note` if it was provided explicitly
3. from the above formatting rules

This way 1) existing semconv and codegen would keep working as is. 2) once we start updating to new deprecation syntax, we won't have to write notes on everything 3) we could keep the original attribute briefs if necessary.

lquerel: The existing code behaves as follows:
- If the semconv file contains a `deprecated` field with a string value, it is automatically converted into:
```
deprecated:your 
  reason: uncategorized
  note: <string value found in the semconv file>
```
- If the semconv file contains a structured `deprecated` field and the `predicated.note` field is defined then we keep it.
- If the semconv file contains a structured `deprecated` field and the `note` sub-field is not defined, this field is initialized with the content of the attribute/signal `brief` field.
- When the `deprecated.note` field is used in a Jinja template, it will always contain one of the two values described above.
- When a template references `deprecated` without the `deprecated | string` filter, Jinja renders the struct itself and does not call the struct’s `to_string` Rust method. As mentioned in the initial release, this introduces a breaking change for templates but does not affect the semconv files themselves (we support the two formats for them).
- When a template calls `deprecated | string`, the `Deprecated::to_string()` Rust method is invoked, and the `deprecated.note` field is used (which should never be empty, as we fall back to `brief` if necessary).

What you’re describing is a bit different. You don’t want to fall back to the `brief` field when `deprecated.note` is missing from the initial semconv file. Instead, you prefer generating a note using one of the patterns specified in your first list (e.g., “Renamed to {renamed_to}”). That works for me, but it’s important to note that there is still a breaking change for existing templates.

When Jinja encounters a struct in a placeholder without a filter, it simply tries to render the struct’s content instead of calling its `to_string` method. I’m not sure whether this is a limitation of the minijinja engine we’re using or just the expected behavior of Jinja (which wouldn’t surprise me).

So, I’ll make the modification you proposed, and I hope these additional details clarify that a breaking change still remains for the template authors.

lmolkova: thanks for the clarification! Yes, it's still a breaking change for the template authors, note-generation saves us from asking everyone to add a note when deprecating. Even today the deprecation message and brief are (usually slightly) different and I feel it's important to keep them this way.

lquerel: Please take a look at the modifications. I also updated the description of this PR with examples coming from my unit tests.

jsuereth (approved):

crates/weaver_forge/src/extensions/otel.rs
```diff
@@ -828,6 +829,162 @@ mod tests {
...
+                renamed_to: new_name
+        "#;
+        let attr: AttributeSpec = serde_yaml::from_str(attr_yaml).unwrap();
+        assert_eq!(
+            env.render_str(
+                "{{ attr.deprecated.reason }}: {{ attr.deprecated.note }}",
+                Ctx { attr },
+            )
+            .unwrap(),
+            "renamed: Replaced by `new_name`."
```
lmolkova, line 880: I don't know if `renamed` helps anyone. Can we do 

```suggestion
            "Replaced by `new_name`."
```

? It'd match what we did in semconv so far

crates/weaver_forge/src/extensions/otel.rs
```diff
@@ -828,6 +829,162 @@ mod tests {
...
+                note: This attribute is deprecated and replaced by `new_name`.
+        "#;
+        let attr: AttributeSpec = serde_yaml::from_str(attr_yaml).unwrap();
+        assert_eq!(
+            env.render_str(
+                "{{ attr.deprecated.reason }}: {{ attr.deprecated.note }}",
+                Ctx { attr },
+            )
+            .unwrap(),
+            "renamed: This attribute is deprecated and replaced by `new_name`."
```
lmolkova, line 902: Can we just use the note as is if it was provided? 

```suggestion
            "This attribute is deprecated and replaced by `new_name`."
```

crates/weaver_forge/src/extensions/otel.rs
```diff
@@ -828,6 +829,162 @@ mod tests {
...
+                reason: obsoleted
+        "#;
+        let attr: AttributeSpec = serde_yaml::from_str(attr_yaml).unwrap();
+        assert_eq!(
+            env.render_str(
+                "{{ attr.deprecated.reason }}: {{ attr.deprecated.note }}",
+                Ctx { attr },
+            )
+            .unwrap(),
+            "obsoleted: Obsoleted."
```
lmolkova, line 922:
```suggestion
            "Obsoleted."
```

crates/weaver_forge/src/extensions/otel.rs
```diff
@@ -828,6 +829,162 @@ mod tests {
...
+                note: This attribute is no longer used.
+        "#;
+        let attr: AttributeSpec = serde_yaml::from_str(attr_yaml).unwrap();
+        assert_eq!(
+            env.render_str(
+                "{{ attr.deprecated.reason }}: {{ attr.deprecated.note }}",
+                Ctx { attr },
+            )
+            .unwrap(),
+            "obsoleted: This attribute is no longer used."
```
lmolkova, line 943:
```suggestion
            "This attribute is no longer used."
```

crates/weaver_forge/src/extensions/otel.rs
```diff
@@ -828,6 +829,162 @@ mod tests {
...
+                reason: uncategorized
+        "#;
+        let attr: AttributeSpec = serde_yaml::from_str(attr_yaml).unwrap();
+        assert_eq!(
+            env.render_str(
+                "{{ attr.deprecated.reason }}: {{ attr.deprecated.note }}",
+                Ctx { attr },
+            )
+            .unwrap(),
+            "uncategorized: Uncategorized."
```
lmolkova, line 963:
```suggestion
            ""
```

crates/weaver_forge/src/extensions/otel.rs
```diff
@@ -828,6 +829,162 @@ mod tests {
...
+                note: This attribute is deprecated for complex reasons.
+        "#;
+        let attr: AttributeSpec = serde_yaml::from_str(attr_yaml).unwrap();
+        assert_eq!(
+            env.render_str(
+                "{{ attr.deprecated.reason }}: {{ attr.deprecated.note }}",
+                Ctx { attr },
+            )
+            .unwrap(),
+            "uncategorized: This attribute is deprecated for complex reasons."
```
lmolkova, line 984:
```suggestion
            "This attribute is deprecated for complex reasons."
```

crates/weaver_forge/src/extensions/otel.rs
lquerel: In this unit test, I'm rendering this Jinja expression `{{ attr.deprecated.reason }}: {{ attr.deprecated.note }}`, so the `renamed: ` is something that is not part of the `deprecated.note` but used to check 2 fields in a unit test.

crates/weaver_forge/src/extensions/otel.rs
lquerel: See my previous comment.

crates/weaver_forge/src/extensions/otel.rs
lquerel: Idem

crates/weaver_forge/src/extensions/otel.rs
lquerel: Idem

crates/weaver_forge/src/extensions/otel.rs
lquerel: idem

crates/weaver_forge/src/extensions/otel.rs
lquerel: Idem

PR #606, opened 2025-02-16
Fix #175

### LINKED ISSUE #175: Detect when the same attribute is referenced more than once
**Describe the bug**
build-tools had a check that prevented referencing the same attribute multiple times within one group.

e.g. 

```yaml
  - id: messaging.servicebus
    type: attribute_group
    extends: attributes.messaging.trace.minimal
    brief: >
      Attributes for Azure Service Bus
    attributes:
      - ref: messaging.operation.name
      - ref: messaging.operation.name
        brief: Azure Service Bus operation name.
```

caused an error like 

```
Error parsing /source/trace/messaging.yaml:
Attribute id messaging.operation.name is already present at line 235 - @251:9 ('messaging.operation.name')
```

**To Reproduce**
See 
https://github.com/open-telemetry/semantic-conventions/pull/1006/commits/3c52fd0e2f0ead12c60a158154374f043c843e38
and
https://github.com/open-telemetry/semantic-conventions/actions/runs/9230121422/job/25397692938?pr=1006 

**Expected behavior**
table check done with weaver should detect it and complain (in the above CI check, it's still back-compat in semconv that raises an issue)

jsuereth:
Add new error and validation code to check if two refs in a resolved group have the same reference id.

This is kept behind `--future` for now

Fixes #175

### DIFF
--- a/crates/weaver_semconv/src/group.rs
+++ b/crates/weaver_semconv/src/group.rs
@@ -5,6 +5,7 @@
 //! A group specification.
 
 use schemars::JsonSchema;
+use std::collections::HashSet;
 use std::fmt::{Display, Formatter};
 
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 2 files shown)


jerbly (approved):
LGTM

PR #607, opened 2025-02-17
Deprecate deprecated stability level

lmolkova:
Semconv don't use `stability: deprecated` for 1+ years (since version 1.24.0) - https://github.com/open-telemetry/semantic-conventions/pull/588. This is necessary to know if deprecated attribute was stable when it was deprecated (and therefore must be generated in semconv artifact) or it was experimental and does not have to be generated.

Having it in weaver is not necessary and misleading. Let's remove it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,8 @@ All notable changes to this project will be documented in this file.
 What's changed
 
 * Add support for annotations on attributes and groups. ([#645](https://github.com/open-telemetry/weaver/pull/645) by @lquerel).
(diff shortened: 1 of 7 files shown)


lmolkova:
@open-telemetry/weaver-approvers  could I get some eyes on this?

jerbly (changes requested):
This makes sense to me if we're happy to raise the minimum supported version. But, we should have an entry in CHANGELOG.md that states this.

jsuereth (approved):
Approval contingent on changelog.

jerbly (approved):
LGTM

PR #611, opened 2025-02-19
Remove constraints and don't default group type to span

lmolkova:
Deprecating a couple more of unused (since v1.26) features: constraints and default group type (span).


crates/xtask/src/history.rs
```diff
@@ -12,7 +12,7 @@ use std::time::Duration;
 const REPO_URL: &str = "https://github.com/open-telemetry/semantic-conventions.git";
 const ARCHIVE_URL: &str =
     "https://github.com/open-telemetry/semantic-conventions/archive/refs/tags/";
-const START_TAG: &str = "v1.22.0";
+const START_TAG: &str = "v1.26.0";
```
lmolkova, line 15: Semconv [document codegen support with weaver from 1.26.0](https://github.com/open-telemetry/semantic-conventions/blob/e70853397b73ca9cb85b1df45182b6a3e299c85e/docs/non-normative/code-generation.md?plain=1#L87), so this should be fine

crates/weaver_forge/src/registry.rs
```diff
@@ -34,7 +34,7 @@ pub struct ResolvedGroup {
     /// The id that uniquely identifies the semantic convention.
     pub id: String,
     /// The type of the group including the specific fields for each type.
-    pub r#type: GroupType,
+    pub r#type: Option<GroupType>,
```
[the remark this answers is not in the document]
jsuereth: If we make it non-optional and someone doesn't fill it out, it means we'll *fail* to parse when it doesn't exist, at the serde level.

I'd rather move directly to that and use a different mechanism in `--future` to make sure we force folks to specify a GroupType going forward.  If you need rust help (this is advanced serde code) to make that help, I'll see if I can find time later in the week to send you a patch.

crates/xtask/src/history.rs
jsuereth: We discussed this a bit ago.  We should have a principle here.  When weaver was first adopted, there was expectation from some SIGs (Go, I believe) that it should be able to go *backwards* beyond just 1.26.0 when it was first adopted.

I'd like to have some principle here.  We discussed using "first stable semconv" as the version we want to support for at least 2 years (as per support policy in OTEL).

lmolkova: I made everything required behind a feature flag and rolled back this change

crates/weaver_forge/src/registry.rs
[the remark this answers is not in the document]
lmolkova: I now default to a new `Undefined` variant allowing to both, have it optional in the yaml, and have validation for the `--future`

crates/weaver_semconv/src/group.rs
```diff
@@ -22,10 +22,11 @@ use weaver_common::result::WResult;
 /// metrics, events, spans, etc.
 #[derive(Serialize, Deserialize, Debug, Clone)]
 #[serde(deny_unknown_fields)]
+#[derive(Default)]
```
lmolkova, line 25: oops

crates/weaver_semconv/src/group.rs
```diff
@@ -61,17 +61,14 @@ pub struct GroupSpec {
     /// List of attributes that belong to the semantic convention.
     #[serde(default)]
     pub attributes: Vec<AttributeSpec>,
-    /// Additional constraints.
-    /// Allow to define additional requirements on the semantic convention.
-    /// It defaults to an empty list.
-    #[serde(default)]
-    pub constraints: Vec<ConstraintSpec>,
+    /// List of constraints
+    pub constraints: Option<Vec<ConstraintSpec>>,
```
lmolkova, line 65: todo: add deprecated annotation?

jsuereth:
Can you bump the baseline version to 1.25 here: https://github.com/open-telemetry/weaver/blob/main/crates/xtask/src/history.rs#L15

That's my only last blocker on this pr.

jsuereth (approved):

PR #615, opened 2025-02-20
Remove 'optional' requirement level

lmolkova:
There was never a formal definition of 'optional' requirement level and it was not used in semconv since quite a while ago (earlier that 1.22.0), let's just remove it.


lquerel (approved):
LGTM. Thanks

lmolkova:
@open-telemetry/weaver-maintainers could we merge this one?

PR #626, opened 2025-02-28
to_yaml filter implementation

vivekprm:
# Need
People have expressed a need to dump a value into yaml file. https://github.com/open-telemetry/weaver/issues/423

# Solution
WIP: This pull request adds a to_yaml filter that can be used to dump a value into a YAML file.


lquerel:
Thank you for this contribution. In the past, we already made a few comments on a similar PR (#525), which also apply to this one. PR #525 is inactive, so there is no issue with submitting a new one to address the same problem. I encourage you to review those comments, fix the CLA issue, and I will proceed with the review of this PR.

vivekprm:
@lquerel Thanks will look into those comments.

crates/weaver_forge/src/extensions/util.rs
```diff
@@ -131,10 +132,20 @@ pub fn acronym(acronyms: Vec<String>) -> impl Fn(&str) -> String {
     }
 }
 
+// Helper filter to dump value (1st parameter) in yaml format.
+fn to_yaml(value: Value) -> Result<String, minijinja::Error> {
+    let yaml = serde_yaml::to_string(&value)
+        .map_err(|e| minijinja::Error::new(ErrorKind::BadSerialization, e.to_string()))?;
+    Ok(yaml)
```
lquerel, line 139: Please take a look at the `tojson` implementation in the minijinja library.

vivekprm: Referred it and modified. However, I am unable to test error conditions..tried few things...

- Using non-UTF 8 bytes..

```rust
fn test_to_yaml_invalid_value() {
        // Now test with an invalid value
        let mut invalid_bytes = Vec::new();
        invalid_bytes.push(0xFF); // Non-UTF-8 byte
        let invalid_value = Value::String(String::from_utf8(invalid_bytes).unwrap());
        let invalid_value = minijinja::value::Value::from_serialize(&invalid_value);
        let result = to_yaml(invalid_value);
        assert!(result.is_err()); // This should pass
}
```
- Using integer key

```rust
fn test_map_with_non_string_key_should_fail() {
        use minijinja::{ErrorKind, Value};
        use std::collections::HashMap;

        let mut bad_map = HashMap::new();
        let _ = bad_map.insert(42, Value::from("Invalid key type")); // Integer key (not allowed in YAML)

        // Convert HashMap to minijinja::Value manually
        let value = Value::from_iter(bad_map); 

        let result = to_yaml(value); 

        // Assert that serialization fails
        assert!(result.is_err());

        if let Err(e) = result {
            assert_eq!(e.kind(), ErrorKind::BadSerialization);
        }
    }
    ```

But in both the cases there is no error any other suggestions?

File: crates/weaver_forge/src/extensions/util.rs [hunk 2]
```diff
@@ -131,10 +132,23 @@ pub fn acronym(acronyms: Vec<String>) -> impl Fn(&str) -> String {
     }
 }
 
+// Helper filter to dump value (1st parameter) in yaml format.
+fn to_yaml(value: Value) -> Result<String, minijinja::Error> {
+    let mut out = Vec::<u8>::new();
+    let mut s = serde_yaml::Serializer::new(&mut out);
+    let yaml = serde::Serialize::serialize(&value, &mut s)
+        .map(|_| unsafe { String::from_utf8_unchecked(out) })
+        .map_err(|e| minijinja::Error::new(ErrorKind::InvalidOperation, e.to_string()))?;
```
jerbly, line 141: Your original implementation simply with `.to_string()` was fine. The important part to add from the minijinja tojson example is the escaping for safety: https://github.com/mitsuhiko/minijinja/blob/main/minijinja/src/filters.rs#L970-L982

vivekprm: Ok..Added..

jerbly: Can you add these special characters to the test example to exercise that code and provide coverage?

vivekprm: Yes thanks...added.

crates/weaver_forge/src/extensions/util.rs
```diff
@@ -131,10 +132,34 @@ pub fn acronym(acronyms: Vec<String>) -> impl Fn(&str) -> String {
     }
 }
 
+// Helper filter to dump value (1st parameter) in yaml format.
+fn to_yaml(value: Value) -> Result<String, minijinja::Error> {
```
jerbly, line 136: The function should take a `&Value` and return a `Result<Value, minijinja::Error>`

crates/weaver_forge/src/extensions/util.rs
```diff
@@ -131,10 +132,34 @@ pub fn acronym(acronyms: Vec<String>) -> impl Fn(&str) -> String {
...
+                    '<' => rv.push_str("\\u003c"),
+                    '>' => rv.push_str("\\u003e"),
+                    '&' => rv.push_str("\\u0026"),
+                    '\'' => rv.push_str("\\u0027"),
+                    _ => rv.push(c),
+                }
+            }
+            rv
+        })?;
+    Ok(yaml)
```
jerbly, line 153: You should use `Value::from_safe_string()` to convert the escaped string to a minijinja value.

vivekprm: Ok..modified

jerbly (approved):
This looks good to me now... thanks for sticking with it and going through the iterations.

@lquerel - can you check this meets your expectations too?

vivekprm:
> This looks good to me now... thanks for sticking with it and going through the iterations.
> 
> @lquerel - can you check this meets your expectations too?

Thanks for your feedbacks and suggestions. Looking to improve and contribute more..

lquerel (approved):

PR #627, opened 2025-03-03
Support for 2 semconv registries

### LINKED ISSUE #604: Support for 2 semconv registries
We propose in this issue an intermediate step before fully supporting the multi-registry concept within the semantics conventions framework (for more details, see the original [multi-registry document](https://github.com/open-telemetry/weaver/blob/main/docs/specs/multi-registry/multi_registry.md)). By limiting multi-registry support to two registries: one custom registry extending the official OTEL registry, we eliminate a number of complexities while addressing a clearly identified, albeit limited, need.  

In this intermediate step, we propose to support:  
- The creation of a custom registry that extends the OTEL registry, allowing the addition of registry attributes, metrics, events, and spans.  
- The creation of an overlay mechanism to override a limited number of fields across the custom registry and the OTEL registry (e.g., `requirement_level`, `tags`).  

This intermediate step does **not** allow:  
- Supporting more than two registries (thus avoiding, among other things, circular dependencies, complex cases of importing different versions within the hierarchy of the same registry).  
- Handling attribute and signal name collisions (via a scoping mechanism).
- Referencing a group to allow a registry to override fields of an imported group without refining a new group with a new id/name.
- Supporting scoped references to disambiguate attribute or signal references coming from multiple other registries.   

### Integration into Weaver:  
- Addition of a new `experimental` flag that will enable this new feature and clearly establish that it is an experimental function. 
- Addition of a `dependencies` section (see example below).  
- Updating the registry loading process to merge the "local" registry with the registry referenced in the dependencies.  
- The merged result then follows the standard/current resolution process, policy enforcement, and generation steps.  

```yaml
name: <custom_registry_name>
description: <registry_description>
version: <registry_version>
...
dependencies:
  - name: otel
    version: <otel_registry_version>
    repository: <url_of_otel_semconv_registry_repository>
```

In this first phase, Weaver will not support more than one dependency.  

This addition to Weaver will essentially replicate what some users are already doing manually to support such a scenario. Furthermore, this integration mode should be compatible in the future with an evolution toward supporting more than two registries.

lquerel:
This PR closes #604.

With this PR, it’s now possible to create a custom SemConv registry that depends on another registry (e.g., the official OTEL SemConv registry). Currently, only a single-level dependency is supported, meaning only two registries can be merged and resolved automatically by Weaver. This PR is an initial step toward full support for multiple registries without these limitations.

For example, let’s imagine a custom registry named `acme` referencing the OTEL registry. The following `registry_manifest.yaml` file should be defined in the directory containing the custom acme registry.

```yaml
name: acme
description: This registry contains the semantic conventions for the Acme vendor.
semconv_version: 0.1.0
schema_base_url: https://acme.com/schemas/
dependencies:
  - name: otel
    registry_path: [data/multi-registry/otel_registry](https://github.com/open-telemetry/semantic-conventions/archive/refs/tags/v1.31.0.zip[model])
```

The `registry_path` can be any valid registry path already supported by Weaver.

The definition of the custom registry `acme` is defined in the following yaml file:

```yaml
groups:
  - id: shared.attributes
    type: attribute_group
    brief: Attributes shared between our signals.
    attributes:
      - id: auction.id
        type: int
        brief: >
          The id of the auction.
        stability: stable
      - id: auction.name
        type: string
        brief: >
          The name of the auction
        stability: stable
        examples: ["Fish for sale"]

  - id: metric.auction.bid.count
    type: metric
    metric_name: auction.bid.count
    stability: stable
    brief: "Count of all bids we've seen"
    instrument: counter
    unit: "{bid}"
    attributes:
      - ref: auction.id
        requirement_level: required
      - ref: auction.name
      - ref: error.type
        requirement_level: required
 ```

The `error.type` attribute is not defined locally, so Weaver will resolve it using the imported registry.

> Notes for reviewers:
> - I’ve extracted the concept of `VirtualDirectory` from `RegistryRepo` in order to (1) avoid a circular reference caused by the registry dependencies feature and (2) prepare for supporting virtual directories for templates and policies in a future PR.
> - The `include_unreferenced` flag has been added to support the use case mentioned by Jeremy. This flag defaults to false.

### CODE REVIEW

**Comment by jerbly** [comment]
This is really good. I tried this on my own company registry where we purposefully want to add attributes into the otel defined namespaces. I got a few of these:
```
The group id `registry.aws.s3` is declared multiple times in the following locations:
  │ [Provenance { registry_id: "otel", path: "https://github.com/open-telemetry/semantic-conventions/archive/refs/tags/v1.31.0.zip[model]/aws/
  │ registry.yaml" }, Provenance { registry_id: "main", path: "model/registry/aws.yml" }]
```
Perhaps we need a _merge_ mode or maybe just group-id naming guidance? Or perhaps a mode where all groups are prepended with the `name` field in the registry_manifest, making: `acme.aws.s3` and `otel.aws.s3`. I'm not sure.

**Comment by jerbly** [comment]
Another usability thing I ran into, not introduced in this PR though, is the discovery of the `registry_manifest.yaml`. Model files are discovered by searching from the path given at the command line. However, the manifest file has to be at the root of the given path. Perhaps the manifest could be discovered in the same way?

Regardless of the above, if the manifest is NOT found it would be useful to have a log line stating it has not been found to help debug. Perhaps I misspelled it or used `.yml`. `Manifest file not found at location ...`

File: crates/weaver_resolver/src/registry.rs [hunk 1]
```diff
@@ -152,9 +153,32 @@ pub fn resolve_semconv_registry(
     );
     check_root_attribute_id_duplicates(&ureg.registry, &attr_name_index, &mut errors);
 
+    gc_unreferenced_objects(registry.manifest(), &mut ureg.registry);
```
jerbly, line 156: Can this be optional? I want my registry to be a superset of all otel semconvs and my private semconvs. Alternatively, we'll need a way to succinctly declare the groups to inherit from the dependency.

lquerel: I added a new flag `include_unreferenced` (false by default) to support this use case.

lquerel:
> This is really good. I tried this on my own company registry where we purposefully want to add attributes into the otel defined namespaces. I got a few of these:
> 
> ```
> The group id `registry.aws.s3` is declared multiple times in the following locations:
>   │ [Provenance { registry_id: "otel", path: "https://github.com/open-telemetry/semantic-conventions/archive/refs/tags/v1.31.0.zip[model]/aws/
>   │ registry.yaml" }, Provenance { registry_id: "main", path: "model/registry/aws.yml" }]
> ```
> 
> Perhaps we need a _merge_ mode or maybe just group-id naming guidance? Or perhaps a mode where all groups are prepended with the `name` field in the registry_manifest, making: `acme.aws.s3` and `otel.aws.s3`. I'm not sure.

I suggest we discuss this during our next SIG meeting.

lquerel:
> Another usability thing I ran into, not introduced in this PR though, is the discovery of the `registry_manifest.yaml`. Model files are discovered by searching from the path given at the command line. However, the manifest file has to be at the root of the given path. Perhaps the manifest could be discovered in the same way?
> 
> Regardless of the above, if the manifest is NOT found it would be useful to have a log line stating it has not been found to help debug. Perhaps I misspelled it or used `.yml`. `Manifest file not found at location ...`

@jerbly Could you create a GitHub issue to add the warning to improve user experience? Regarding the recursive search for the manifest file, I’m less sure about it, as it raises the problem of potentially finding multiple manifests and deciding how to handle that. For now, I think it’s best to keep it simple until we have a clear need for that type of functionality.

Issue #557, opened 2025-01-16
Ensuring Conformance of Application OTLP Streams to Semantic Conventions via Registry Validation


lquerel:
## Problem
An application’s adherence to semantic conventions is currently only ensured by the goodwill of the developers and reviewers of the application. It is possible, and even highly likely, to see divergences and improper uses emerge during the lifecycle of the application and/or the semantic conventions. This lack of consistency between the semantic conventions used by an application and the instrumentation integrated into it is unsatisfactory.

## Solutions

- Use a type-safe client SDK generated from the semantic conventions. This approach is not the subject of this issue.
- Use Weaver to verify the conformity of the OTLP stream emitted by the application (during the build pipeline or emitted by tests) against the registry associated with the application. This solution is described in more detail below.

Command: `registry live-check` usage

```
Usage: weaver registry live-check [OPTIONS]

Options:
      --debug...
          Enable debugging information.

  -r, --registry <REGISTRY>
          Local folder, Git repository URL, or Git archive URL of the semantic convention registry. For Git URLs, a sub-folder can be specified using the `[sub-folder]` syntax after the URL.

          [default: https://github.com/open-telemetry/semantic-conventions.git[model]]

      --quiet
          Enable quiet mode (minimal output).

  -s, --follow-symlinks
          Boolean flag to specify whether to follow symlinks when loading the registry. Default is false.

      --future
          Enable the most recent validation rules for the semconv registry. It is recommended to enable this flag when checking a new registry. Note: `semantic_conventions` main branch should always enable this flag.

  -p, --otlp-grpc-port <OTLP_GRPC_PORT>
          Port used by the gRPC OTLP listener.

          [default: 4317]

  -a, --admin-port <ADMIN_PORT>
          Port used by the HTTP admin port (endpoints: /stop).

          [default: 4320]

  -t, --inactivity-timeout <INACTIVITY_TIMEOUT>
          Maximum inactivity time in seconds before stopping the listener.

          [default: 10]

      --diagnostic-format <DIAGNOSTIC_FORMAT>
          Format used to render diagnostic messages. Predefined formats are: ansi, json, gh_workflow_command.

          [default: ansi]

      --diagnostic-template <DIAGNOSTIC_TEMPLATE>
          Path to the directory where the diagnostic templates are located.

          [default: diagnostic_templates]

  -h, --help
          Print help (see a summary with '-h').
```

[...]

PR #630, opened 2025-03-06
Weaver Live Check System

jerbly:
# Weaver Live Check

Live check is a developer tool for assessing sample telemetry and providing advice for improvement.

A Semantic Convention `Registry` is loaded for comparison with samples. `Ingesters` transform various input formats and sources into intermediary representations to be assessed by `Advisors`. The `Advice` produced is transformed via jinja templates to the required output format for downstream consumption.

```mermaid
flowchart LR
    subgraph Inputs
        file["File"]
        stdin["stdin"]
        otlp["OTLP"]
    end

    subgraph Core["Processing"]
        registry["Registry"]
        ingesters["Ingesters"]

        subgraph advisors["Advisors"]
            builtin["Builtin"]

            subgraph external["External"]
                otel["Otel"]
                custom["Custom"]
            end
        end
    end

    subgraph Outputs
        advice["Advice"]
        templates["Jinja Templates"]
        output["Output Format"]
    end

    file --> ingesters
    stdin --> ingesters
    otlp --> ingesters

    registry -- "Loaded for comparison" --> advisors
    ingesters -- "Intermediary representations" --> advisors

    builtin --> advice
    external --> advice

    advice -- "Transformed via" --> templates
    templates --> output
```

This PR partially covers #557. Details in the [README](https://github.com/open-telemetry/weaver/pull/630/files?short_path=06dbb93#diff-06dbb93928a7cae1e6796a895e12491a357866cf4d76123b1b31c61c87e4f359)

Here's an example ANSI output:
image


crates/weaver_health/src/attribute_advice.rs
```diff
@@ -0,0 +1,106 @@
...
+/// The severity of the advice
+#[derive(Debug, Clone, PartialEq, Serialize)]
+pub enum Severity {
+    /// An error
+    Error,
+    /// A warning
+    Warning,
+    /// Information
+    Info,
+}
```
lquerel, line 19: Is there a good reason not to use the `miette` diagnostic mechanism already in place? Miette already supports the concept of `SeverityLevel` and is already integrated with our `--future` flag.

jerbly: Oh, maybe `Severity` is not the right term here. I want to express whether the `Advice` given is a failure: like using a deprecated attribute, or a hint: like suggesting a similar attribute to use instead.

The output of the analysis is not a simple diagnostic list, it's more structured, like the diff output.

jerbly: I've swapped this out for `Advisory`: `Violation`, `Improvement`, `Information` to better express what the advisory level of the advice.

crates/weaver_health/src/attribute_advice.rs
```diff
@@ -0,0 +1,308 @@
...
+        _attribute: &SampleAttribute,
+        _health_checker: &AttributeHealthChecker,
+        semconv_attribute: Option<&Attribute>,
+    ) -> Option<Advice> {
+        if let Some(attribute) = semconv_attribute {
+            if attribute.deprecated.is_some() {
+                Some(Advice {
+                    key: "is_deprecated".to_owned(),
+                    value: Value::Bool(true),
+                    message: "Is deprecated".to_owned(),
```
lquerel, line 69: Could be more informational to build a text message based on the deprecated field.

crates/weaver_health/src/attribute_advice.rs
```diff
@@ -0,0 +1,308 @@
...
+        _health_checker: &AttributeHealthChecker,
+        semconv_attribute: Option<&Attribute>,
+    ) -> Option<Advice> {
+        if let Some(attribute) = semconv_attribute {
+            match attribute.stability {
+                Some(ref stability) if *stability != Stability::Stable => Some(Advice {
+                    key: "stability".to_owned(),
+                    value: Value::String(stability.to_string()),
+                    message: "Is not stable".to_owned(),
+                    advisory: Advisory::Improvement,
```
lquerel, line 99: How do we decide what constitutes an improvement versus a violation? For instance, this `StabilityAdvisor` generates an improvement when the attribute isn't `Stable`. However, in some scenarios, a user might consider this a violation. Perhaps there's a mechanism elsewhere in the system to handle this in one way or another.

crates/weaver_health/src/attribute_advice.rs
```diff
@@ -0,0 +1,308 @@
...
+    Violation,
+}
+
+/// Represents a health check advice
+#[derive(Debug, Clone, PartialEq, Serialize)]
+pub struct Advice {
+    /// The key of the advice e.g. "is_deprecated"
+    pub key: String,
+    /// The value of the advice e.g. "true"
+    pub value: Value,
```
lquerel, line 32: I'm not sure I understand how this `value` field:
- is used by the reporting system,
- adds something compared to the following `message` field.

crates/weaver_health/src/attribute_advice.rs
```diff
@@ -0,0 +1,308 @@
...
+                    advisory: Advisory::Information,
+                });
+            }
+        } else {
+            // Does not have a namespace
+            return Some(Advice {
+                key: "namespace".to_owned(),
+                value: Value::Bool(false),
+                message: "Does not have a namespace".to_owned(),
+                advisory: Advisory::Improvement,
```
lquerel, line 183: I have a similar question here. Why choose an improvement here rather than a violation?

crates/weaver_health/src/attribute_advice.rs
```diff
@@ -0,0 +1,308 @@
...
+                    key: "namespace".to_owned(),
+                    value: Value::String(existing_namespace),
+                    message: "Extends existing namespace".to_owned(),
+                    advisory: Advisory::Information,
+                });
+            }
+        } else {
+            // Does not have a namespace
+            return Some(Advice {
+                key: "namespace".to_owned(),
```
lquerel, line 180: Why isn't this a `no_namespace` key? This would provide a way to filter by advisor output type. I have the same kind of comment for the other advisor results related to the namespace concept.

crates/weaver_health/src/attribute_advice.rs
```diff
@@ -0,0 +1,308 @@
...
+                        TemplateTypeSpec::Doubles => PrimitiveOrArrayTypeSpec::Doubles,
+                        TemplateTypeSpec::Booleans => PrimitiveOrArrayTypeSpec::Booleans,
+                    },
+                    AttributeType::Enum { .. } => {
+                        // Special case: Enum variants can be either string or int
+                        if attribute_type != &PrimitiveOrArrayTypeSpec::String
+                            && attribute_type != &PrimitiveOrArrayTypeSpec::Int
+                        {
+                            return Some(Advice {
+                                key: "type".to_owned(),
```
lquerel, line 223: Perhaps `type_mismatch` would be more precise.

crates/weaver_health/src/attribute_advice.rs
```diff
@@ -0,0 +1,308 @@
...
+                            }
+                        } {
+                            is_found = true;
+                            break;
+                        }
+                    }
+
+                    if !is_found {
+                        return Some(Advice {
+                            key: "enum".to_owned(),
```
lquerel, line 296: `unknown_enum_variant` ?

lquerel (commented):
This feature will bring a lot to Weaver. I've started adding comments on the advisor section. I'll complete the full review as soon as possible.

Personally, I liked the duality between the commands `weaver registry check` and `weaver registry live-check`. The naming `weaver registry health` doesn't resonate as much with me, but that's purely a personal opinion. It might be good to get input from others on this point.

Additionally, for advisors related to best practices (especially syntax), I'm wondering if we could somehow leverage the `rego` policies that we already maintain for the registry.

Also, in my opinion, the goal of this command is to be integrated into a CI/CD pipeline, and therefore it should be capable of returning a non-zero value when there are violations. In another PR, we might also consider generating a report compatible with the format used by coverage tools.

lquerel:
@jerbly Regarding my comment:  
> "For advisors related to best practices (especially syntax), I'm wondering if we could somehow leverage the rego policies that we already maintain for the registry,"  

one possible approach could be generating a registry from an input (OTLP, JSON, ...) to obtain violations that we already maintain elsewhere. There might be good reasons not to reuse existing policies, and this would require significant changes in this PR, but I think it's worth exploring if we want to avoid maintaining two separate sets of policies.

crates/weaver_health/src/attribute_advice.rs
```diff
@@ -0,0 +1,308 @@
...
+
+/// Provides advice on a sample attribute
+pub trait Advisor {
+    /// Provide advice on a sample attribute
+    fn advise(
+        &self,
+        attribute: &SampleAttribute,
+        health_checker: &AttributeHealthChecker,
+        semconv_attribute: Option<&Attribute>,
+    ) -> Option<Advice>;
```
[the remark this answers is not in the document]
jerbly: The concept is that each advisor has one job. So for similarity we would make a SimilarityAdvisor which has the sole job of returning an Advice suggesting other similar attribute names.

crates/weaver_health/src/attribute_advice.rs
jerbly: Yes. I added that in this commit: https://github.com/jerbly/weaver/commit/649fff400b7907093a8f1dfd37afc7a779a4a6c6

crates/weaver_health/src/attribute_advice.rs
jerbly: I have a task in the PR to add a configuration file. This will allow you to:
- select which advisors you want
- configure those advisors
At the moment, both the Enum and Stability advisors need configurable advisory level. I've chosen "sensible" defaults.

crates/weaver_health/src/attribute_advice.rs
jerbly: My idea here is that:
- Key = the name of the advisor. So `stability` or `enum` or `type` etc. (perhaps `advisor_name` would be better than `key`)?
- Value = a machine readable simple indicator for the advice.
- Message = a human readable phrase more descriptive than the Value but to variable too be machine readable.

crates/weaver_health/src/attribute_advice.rs
jerbly: Yes. I will make all the keys better. https://github.com/open-telemetry/weaver/pull/630#discussion_r1997671662

crates/weaver_health/src/attribute_advice.rs
jerbly: Yes.

jerbly:
> @jerbly Regarding my comment:
> 
> > "For advisors related to best practices (especially syntax), I'm wondering if we could somehow leverage the rego policies that we already maintain for the registry,"
> 
> one possible approach could be generating a registry from an input (OTLP, JSON, ...) to obtain violations that we already maintain elsewhere. There might be good reasons not to reuse existing policies, and this would require significant changes in this PR, but I think it's worth exploring if we want to avoid maintaining two separate sets of policies.

I'm not sure I fully follow this, perhaps we can talk about this during the week. I'm happy to change anything to send us in the right direction. I felt we needed this initial iteration to see something in action, spark conversation and make changes.

crates/weaver_health/src/attribute_advice.rs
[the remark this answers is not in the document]
lquerel: Yes, I understand, but where is the registry that could be used to identify similar attributes? In the context of the advisor, in the HealthChecker? That’s the point I’d like to clarify.

If an advisor wants to return more than one advice, for example, to return two similar attributes, the idea is to return a single advice containing both similar attributes inside? That’s possible. The other option is to replace Option<Advice> with Vec<Advice>. I’m not sure yet if it’s worth it. Something to think about.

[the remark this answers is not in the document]
jerbly: > the idea is to return a single advice containing both similar attributes inside?

Yes, this was my idea. I made `value` flexible. For Similarity, `value` could be an array of similar attributes.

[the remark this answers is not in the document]
jerbly: > ...where is the registry that could be used to identify similar attributes? In the context of the advisor, in the HealthChecker? That’s the point I’d like to clarify.

I was leaving this for phase 2. But to implement this, SimilarityAdvisor would access the registry and hashmaps held by the HealthChecker. I planned to add a new `initialise` fn to the Advisor trait which would be called on each advisor prior to the run. This would allow any additional pre-run setup to be done, e.g. feed a statistical algorithm or LLM with the whole registry.

I'm also adding a `conclude` fn to the trait in this PR for the advisors to provide post-run information. For example, how many violations they made each and other statistics.

crates/weaver_health/src/attribute_advice.rs
jerbly: See https://github.com/open-telemetry/weaver/pull/630/commits/7a86e4c5ccc32d00a5bef9265eb629e285f7a4c0

crates/weaver_health/src/attribute_advice.rs
jerbly: See https://github.com/open-telemetry/weaver/pull/630/commits/7a86e4c5ccc32d00a5bef9265eb629e285f7a4c0

crates/weaver_health/src/attribute_advice.rs
jerbly: See https://github.com/open-telemetry/weaver/pull/630/commits/7a86e4c5ccc32d00a5bef9265eb629e285f7a4c0

crates/weaver_health/src/attribute_advice.rs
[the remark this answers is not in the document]
jsuereth: I also think this should allow returning more than one type of Advice from the same Advisor.

I'm thinking about a "rego" based, custom extension here.

crates/weaver_health/src/attribute_advice.rs
```diff
@@ -0,0 +1,382 @@
...
+        }
+    }
+}
+impl Default for NameFormatAdvisor {
+    fn default() -> Self {
+        Self::new(r"^[a-z][a-z0-9]*([._][a-z0-9]+)*$")
+    }
+}
+
+impl Advisor for NameFormatAdvisor {
```
jsuereth, line 131: Shouldn't this be checked in rego polciies on semantic conventions themsselves?

I'm not sure I understand why Name Format is checked here.

I understand we can have template types, but the format of the name is purposefully *open* there.

crates/weaver_health/src/attribute_advice.rs
```diff
@@ -0,0 +1,382 @@
...
+                advisory: Advisory::Violation,
+            })
+        } else {
+            None
+        }
+    }
+}
+
+/// An advisor that provides advice on the namespace of an attribute
+pub struct NamespaceAdvisor {
```
jsuereth, line 157: Similar question here.

crates/weaver_health/src/attribute_advice.rs
jerbly: Now that I've been through this a few times and had the discussions I have a better sense of where the line is between internally rigid rules built in to the rust code and externally defined. IMO the semconv name format is strongly recommended and should be OOTB but by moving this to rego means it can be overridden. (I do wonder though if we have got this wrong in some places though. `Stability` for example?)

I included a name format check on the incoming samples because this has proven useful in my own tool. It provides guidance to developers for ways they can improve their telemetry whether they go on to define the attribute formally or not.

I will move this to rego.

crates/weaver_health/src/attribute_advice.rs
jerbly: Same answer as above. This was also taken from my own tool. For my developers it has proven useful to see, for example, that a sample attribute extends an existing namespace. This may or may not be intentional so this is a good point of guidance. Likewise, indicating that the sample would collide with a defined attribute or that it's missing a namespace altogether is all just useful guidance. Again, whether you decide to formally define the attribute or not.

I will move this to rego too.

crates/weaver_health/src/attribute_advice.rs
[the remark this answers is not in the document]
jerbly: I agree and will change this.

crates/weaver_health/src/attribute_advice.rs
jerbly: Moved to rego in this commit: [f6b7023](https://github.com/open-telemetry/weaver/pull/630/commits/f6b702395e8c346eafa7626538e5e06b634eddfb)

crates/weaver_health/src/attribute_advice.rs
jerbly: Moved to rego in this commit: [f6b7023](https://github.com/open-telemetry/weaver/pull/630/commits/f6b702395e8c346eafa7626538e5e06b634eddfb)

crates/weaver_health/src/attribute_advice.rs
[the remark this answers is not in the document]
jerbly: Done in this commit: [f2041af](https://github.com/open-telemetry/weaver/pull/630/commits/f2041afe2c82cec14de66db2c05938940b665b6e)

crates/weaver_live_check/src/attribute_live_check.rs
```diff
@@ -0,0 +1,727 @@
...
+            }
+        };
+
+        if semconv_attribute.is_none() {
+            attribute_result.add_advice(Advice {
+                key: "missing_attribute".to_owned(),
+                value: Value::String(sample_attribute.name.clone()),
+                message: "Does not exist in the registry".to_owned(),
+                advisory: Advisory::Violation,
+            });
```
lquerel, line 102: In a future PR, it might be interesting to identify attributes with similar names for the attributes that are missing.

crates/weaver_live_check/src/attribute_live_check.rs
```diff
@@ -0,0 +1,727 @@
...
+        } else {
+            self.highest_advisory = Some(advisory);
+        }
+        self.all_advice.push(advice);
+    }
+}
+
+/// A live check report for a set of attributes
+#[derive(Debug, Clone, PartialEq, Serialize)]
+pub struct LiveCheckReport {
```
lquerel, line 186: Shouldn’t we have the following structures in a separate file that isn’t solely associated with attributes? I’m thinking this `LiveCheckReport` will also be used to report advice on metrics, spans, events, etc.

crates/weaver_live_check/src/attribute_live_check.rs
```diff
@@ -0,0 +1,727 @@
...
+    pub fn has_violations(&self) -> bool {
+        self.statistics
+            .highest_advisory_counts
+            .contains_key(&Advisory::Violation)
+    }
+}
+
+/// The statistics for a live check report
+#[derive(Debug, Clone, PartialEq, Serialize)]
+pub struct LiveCheckStatistics {
```
lquerel, line 205: As mentioned earlier, I think we should add information to obtain or calculate coverage details for attributes (and, in the future, for signals) relative to the reference registry.

src/registry/live_check.rs
```diff
@@ -1,112 +1,340 @@
...
+
+/// The type of ingester to use
+#[derive(Debug, Clone)]
+enum IngesterType {
+    AttributeFile,
+    AttributeStdin,
+    AttributeJsonFile,
+    AttributeJsonStdin,
+    AttributeOtlp,
+    GroupFile,
```
lquerel, line 46: IIRC, this variant is not described in the documentation.

src/registry/mod.rs
```diff
@@ -177,19 +170,11 @@ pub struct PolicyArgs {
...
-            skip_policies: true,
-            display_policy_coverage: false,
-        }
-    }
-}
-
 /// Manage a semantic convention registry and return the exit code.
-pub fn semconv_registry(log: impl Logger + Sync + Clone, command: &RegistryCommand) -> CmdResult {
+pub fn semconv_registry(
+    log: impl Logger + Sync + Clone + 'static,
```
lquerel, line 175: I didn’t understand why there was a need for a `static` lifetime for the logger. The logger is intended to be cloned (with low overhead) when there’s a need to pass it as a parameter to multiple sub-functions or methods.

lquerel (changes requested):
I’ve made a few suggestions/requests. In general, I think it’s really good, and we’re very close to having a mergeable PR. Given the size of the PR, I think it’s important to have a review from @jsuereth  to get a second pair of eyes. Thanks for all the work!

crates/weaver_live_check/README.md
```diff
@@ -0,0 +1,239 @@
...
+
+## Ingesters
+
+Sample data can have various levels of detail; from a simple list of attribute names, to a full OTLP signal structure. This data can come from different sources: files, stdin, OTLP. Therefore you need to choose the appropriate `Ingester` for your job:
+
+- `attribute_file | AF | af` => for attribute name or name=value pair text files
+- `attribute_stdin | AS | as` => for attribute name or name=value pairs on stdin
+- `attribute_json_file | AJF | ajf` => for an array of attribute samples in a json file
+- `attribute_json_stdin | AJS | ajs` => for an array of attribute samples in json on stdin
+- `attribute_otlp | AO | ao` => for attributes extracted from OTLP signals
```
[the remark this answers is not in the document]
jerbly: I will split this into three things:
- `--input-source` : `file`, `stdin`, `otlp`
- `--input-format` : `text`, `json` _(not required for otlp)_
- `--scope` : `attributes`, `signals`

The correct ingester will be chosen from this selection.

crates/weaver_live_check/src/attribute_live_check.rs
jerbly: I do this in my own tool using the strsim crate and Jaro similarity. We could implement the same here but I'm not blown away by the results. An LLM may be better here - but that's a heavyweight and slow alternative.

src/registry/mod.rs
jerbly: I've changed the design to avoid this now. I needed to own a struct that needed a logger. I figured adding the 'static lifetime would be ok since the underlying paris logger is 'static. But, anyway, I was able to move things around to avoid this.

crates/weaver_live_check/README.md
```diff
@@ -0,0 +1,239 @@
...
+
+    advice -- "Transformed via" --> templates
+    templates --> output
+```
+
+## Ingesters
+
+Sample data can have various levels of detail; from a simple list of attribute names, to a full OTLP signal structure. This data can come from different sources: files, stdin, OTLP. Therefore you need to choose the appropriate `Ingester` for your job:
+
+- `attribute_file | AF | af` => for attribute name or name=value pair text files
```
jsuereth, line 53: This seems to put a huge focus on "attributes" vs. "signals".  I'm not positive that's the right focus we should have for Weaver.

Yes, semantic conventions has a huge focus on attributes, by we really want to push for *signals* and signal generation code (i.e. span, metric, event and resource[entity]s).

While I think it's fine if the PR only supports attributes initially, I think we should think carefully about ingesters we recommend.

I'd rather see `{signal}_json_file`, `{signal}_otlp` and `{signal}_json_stdin` where we only pay attention to attributes initially.

crates/weaver_live_check/README.md
```diff
@@ -0,0 +1,239 @@
...
+- `attribute_json_stdin | AJS | ajs` => for an array of attribute samples in json on stdin
+- `attribute_otlp | AO | ao` => for attributes extracted from OTLP signals
+
+Some `Ingesters`, like `attribute_stdin` and `attribute_otlp`, can stream the input data so you receive output at the command line as it comes in. This is really useful in live debugging sessions allowing you to breakpoint, step through your code and see live assessment as the data is received in Weaver.
+
+## Advisors
+
+Sample entities are assessed by the set of `Advisors` to produce a list of `Advice` for each one. Built-ins check for fundamental compliance with the `Registry` supplied, for example `missing_attribute` and `type_mismatch`.
+
+Beyond the fundamentals, external `Advisors` can be defined in Rego policies. The OpenTelemetry Semantic Conventions rules are included out-of-the-box by default. They provide `Advice` on name-spacing and formatting aligned with the standard. These default policies can be overridden at the command line with your own.
```
jsuereth, line 65: Are we going to provide semantic convention rules out of the box by default?

(a) I'm not sure all semantic convention rules actually are useful for general projects.
(b) A *subset* of semantic convention rules (or semantic convention usage rules) would be useful to assert.

I haven't read through what you have in this PR yet, so I'll update this comment if I see something along the lines of (a).  If you don't have any out of the box rules yet, we should mark this aspirational.

crates/weaver_live_check/README.md
```diff
@@ -0,0 +1,239 @@
...
+  "all_advice": [
+    {
+      "advisory": "violation",
+      "key": "missing_attribute",
+      "message": "Does not exist in the registry",
+      "value": "aws.s3.extension.name"
+    },
+    {
+      "advisory": "information",
+      "key": "extends_namespace",
```
jsuereth, line 87: This policy probably needs a link to explain *why* here. 

Right now, if I were unfamiliar with OTel's naming rules, this wouldn't make a lot of sense.

crates/weaver_live_check/README.md
```diff
@@ -0,0 +1,239 @@
...
+make_advice(key, advisory, value, message) := {
+	"type": "advice",
+	"key": key,
+	"advisory": advisory,
+	"value": value,
+	"message": message,
+}
+```
+
+`input` contains the sample entity for assessment. `data` contains a structure derived from the supplied `Registry`. A jq preprocessor takes the `Registry` (and maps for attributes and templates) to produce the `data` for the policy. If the jq is simply `.` this will passthrough as-is. Preprocessing is used to improve Rego performance and to simplify policy definitions. With this model `data` is processed once whereas the Rego policy runs for every sample entity as it arrives in the stream. Preprocessing for the default otel policies lead to a ~10x speed-up.
```
jsuereth, line 128: This is a nit beyond the scope of your PR, but we probably need a link here to the actual *schema* of the JSON for `input` and `data`.

crates/weaver_live_check/src/attribute_advice.rs
```diff
@@ -0,0 +1,302 @@
...
+impl Advisor for TypeAdvisor {
+    fn advise(
+        &mut self,
+        attribute: &SampleAttribute,
+        semconv_attribute: Option<&Attribute>,
+    ) -> Result<Vec<Advice>, Error> {
+        // Only provide advice if the attribute is a match and the type is present
+        match (semconv_attribute, attribute.r#type.as_ref()) {
+            (Some(semconv_attribute), Some(attribute_type)) => {
+                let semconv_attribute_type = match &semconv_attribute.r#type {
```
jsuereth, line 108: I wonder if we should make a helper method on the struct of `Attribute` for this.  WDYT?

crates/weaver_live_check/src/attribute_advice.rs
```diff
@@ -0,0 +1,302 @@
...
+        if let Some(path) = policy_dir {
+            let _ = engine
+                .add_policies(path, "*.rego")
+                .map_err(|e| Error::AdviceError {
+                    error: e.to_string(),
+                })?;
+        } else {
+            let _ = engine
+                .add_policy(
+                    "defaults/policies/advice/otel.rego",
```
jsuereth, line 235: Can we move this in a constant somewhere *or* a method where it's obvious what we're adding by default?

If this isn't just pulling out all the defaults "by fiat" magically, we should very clearly have a method or location where we update this.

defaults/policies/advice/otel.rego
```diff
@@ -0,0 +1,97 @@
...
+derive_namespaces(name) := [
+concat(".", array.slice(parts, 0, i)) |
+	parts := split(name, ".")
+	count(parts) > 1 # Only derive namespaces if there are at least 2 parts
+
+	# Stop at count(parts)-1 to exclude the full attribute name
+	some i in numbers.range(1, count(parts) - 1)
+]
+
+# checks attribute has a namespace
```
jsuereth, line 23: should we enforce this by default on non-otel projects?

We actually allow raw names for users.

I'm on the fence of whether this should be default.  I think it *is* good advice, and I like you have it as improvement.

defaults/policies/advice/otel.rego
```diff
@@ -0,0 +1,97 @@
...
+deny contains make_advice(key, advisory, value, message) if {
+	not contains(input.name, ".")
+	key := "missing_namespace"
+	advisory := "improvement"
+	value := input.name
+	message := "Does not have a namespace"
+}
+
+# checks attribute name format
+deny contains make_advice(key, advisory, value, message) if {
```
jsuereth, line 33: This is also something we do not require non-semconv attributes to have.

We should definitely use this for Semantic conventions, but don't think we need to force everyone to do it.  E.g. think "prometheus metrics"

defaults/policies/advice/otel.rego
```diff
@@ -0,0 +1,97 @@
...
+	not regex.match(name_regex, input.name)
+	key := "invalid_format"
+	advisory := "violation"
+	value := input.name
+	message := "Does not match name formatting rules"
+}
+
+# checks attribute namespace doesn't collide with existing attributes
+deny contains make_advice(key, advisory, value, message) if {
```
jsuereth, line 42: This is also something we only enforce in semconv.

Not only that - why do we need this in advice in addition to testing the repo itself for this check?

I.e. shouldn't this check *only* be done on a registry?

src/registry/otlp/attribute_otlp_ingester.rs
```diff
@@ -0,0 +1,193 @@
...
+            OtlpRequest::Metrics(_metrics) => {
+                // TODO Implement the checking logic for metrics
+                // self.logger.error("Metrics Request received");
+                self.next()
+            }
+            OtlpRequest::Traces(trace) => {
+                // Process and buffer all attributes from the trace
+                for resource_span in trace.resource_spans {
+                    if let Some(resource) = resource_span.resource {
+                        for attribute in resource.attributes {
```
jsuereth, line 81: Should we also be tagging that the attribute came from a resource or span so we can use that in the rego policy?

src/registry/otlp/attribute_otlp_ingester.rs
```diff
@@ -0,0 +1,193 @@
...
+
+impl OtlpAttributeIterator {
+    fn new(otlp_requests: Box<dyn Iterator<Item = OtlpRequest>>) -> Self {
+        Self {
+            otlp_requests,
+            buffer: Vec::new(),
+        }
+    }
+
+    fn extract_attributes_from_request(&mut self, request: OtlpRequest) -> Option<SampleAttribute> {
```
jsuereth, line 65: Nit: I'd find this class less confusing if it was shifted a bit:

- Next very clearly attempts to have a buffer *before* returning next.
- Methods like this are responsible for filling the buffer.

```
 /// Puts sample attributes in self.buffer from a request.
 fn fill_attribute_buffer_from_requests(&mut self, request: OtlpRequest) {
   ...
 }
 ... impl Iterator ...
   fn next(&mut self) -> Option<Self::Item> {
     while (self.buffer.empty()) {
        match self.otlp_requests.next() {
            Some(request) => self.extract_attributes_from_request(request),
            None => return None,
        }
     }
     // We know we have a buffer at this point.
     Some(self.buffer.remove(0));
   }
```

src/registry/live_check.rs
```diff
@@ -1,112 +1,340 @@
...
+    /// Ingester type
+    ///
+    /// - `attribute_file` or `AF` or `af` (default)
+    /// - `attribute_stdin` or `AS` or `as`
+    /// - `attribute_json_file` or `AJF` or `ajf`
+    /// - `attribute_json_stdin` or `AJS` or `ajs`
+    /// - `attribute_otlp` or `AO` or `ao`
+    #[clap(verbatim_doc_comment)]
+    #[arg(short = 'g', long)]
+    ingester: IngesterType,
```
jsuereth, line 91: See my comment on ingestors.

src/registry/live_check.rs
```diff
@@ -1,112 +1,340 @@
...
+    /// Path to the directory where the templates are located.
+    #[arg(long, default_value = "live_check_templates")]
+    templates: PathBuf,
+
+    /// Stream mode. Set to false to disable streaming output.
+    ///
+    /// When the output is STDOUT, Ingesters that support streaming (STDIN and OTLP),
+    /// by default output the live check results for each entity as they are ingested.
+    #[arg(long)]
+    stream: Option<bool>,
```
jsuereth, line 106: Does clap support a default?  I.e. could we default to true here?

src/registry/live_check.rs
```diff
@@ -1,112 +1,340 @@
...
-        match otlp_request {
-            OtlpRequest::Logs(_logs) => {
-                // ToDo Implement the checking logic for logs
-                println!("Logs Request received");
-            }
-            OtlpRequest::Metrics(_metrics) => {
-                // ToDo Implement the checking logic for metrics
-                println!("Metrics Request received");
+    // Create the live checker with advisors
+    let advisors: Vec<Box<dyn Advisor>> = vec![
```
jsuereth, line 173: Nit - Can we move this into a loose fucntion, like `fn default_advisors() -> Vec<Box<dyn Advisor>>`

src/registry/live_check.rs
```diff
@@ -1,112 +1,340 @@
...
+        }
+    };
+
+    // If this is a stream - process the attributes one by one
+    // File output is not supported in stream mode
+    if let OutputDirective::File = output_directive {
+        stream_mode = false;
+    }
+
+    if stream_mode {
```
jsuereth, line 276: I wonder if we can "hide" stream mode behind a trait that Ingestor would implement...

Thoughts for future refactor

jsuereth (commented):
Overall this looks amazing!  Very excited to get this merged.

I have a lot of nits and some important comments.  Important things to consider:

1. Which OTEL conventions are actually general (think the prometheus use case, e.g.)
2. I'm not a huge fan of the focus on pure attributes.  I think we should keep the user-facing surface focused on signals (or bundles if you will).  I think it's fine if we only CHECK attributes immediately, but we should make sure:
  - The surface users see ties to our group mechanisms (span, event, metric, log, resource)
   - For live-check we should know which of these an attribute came from (AttributeSample should have an enum of source?)

crates/weaver_live_check/README.md
jerbly: I have a few commits coming in that change how `Ingesters` are configured on the CLI. 

TBH I just did attributes in this first PR because I was trying to limit scope but show off lots of use-cases to whet the appetite. I envisage the defaults flipping in future PRs.

crates/weaver_live_check/README.md
jerbly: Removed from readme. It was just an example.

crates/weaver_live_check/README.md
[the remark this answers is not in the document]
jerbly: Removed.

crates/weaver_live_check/src/attribute_advice.rs
jerbly: Do you mean the conversion from the definition type to the accepted value type?

defaults/policies/advice/otel.rego
jerbly: We could default to no Rego and then allow the `--advice-policies` to accept a path or the string `otel`. We could even have `prometheus` as an embedded policy too if there's a "standard.

src/registry/otlp/attribute_otlp_ingester.rs
jerbly: Yes, we'll need surrounding context. I think, when we move on to support signals properly, these requirements will pop out.

jerbly:
> I don't like that attribute vs. signal checking is a top level concern, or the number of (hopefully re-usable) pieces of this labelled "attribute". I think we're dividing the world to shrink scope, but the divide isn't where it should be.
> ...

OK. In a very early cut of this I had the Ingesters discovering the incoming scope rather than the user explicitly setting the scope. I'll revisit that.

jerbly:
@jsuereth I'm pretty happy with this now and it checks off everything in your comment. Extending this PR to be signal focussed lead to a big refactor which resulted in a much better framework. It all defaults to OTLP now by default so: `weaver registry live-check` is all you need on the command line for a live streaming OTLP input to stdout.

jsuereth:
@jerbly It's a bit of a sour week for me, I should be able to review this by Thursday, latest.  APologies for delay, looking forward to looking at it again.

crates/weaver_live_check/src/advice.rs
```diff
@@ -0,0 +1,345 @@
...
+
+use crate::{live_checker::LiveChecker, Error, SampleRef};
+
+/// Embedded default live check rego policies
+pub const DEFAULT_LIVE_CHECK_REGO: &str =
+    include_str!("../../../defaults/policies/live_check_advice/otel.rego");
+
+/// Default live check rego policy path - used in error messages
+pub const DEFAULT_LIVE_CHECK_REGO_POLICY_PATH: &str =
+    "defaults/policies/live_check_advice/otel.rego";
```
jsuereth, line 29: Should this be `builtin:/policies/live_check_advice/otel.rego` going forward?

crates/weaver_live_check/src/advice.rs
```diff
@@ -0,0 +1,345 @@
...
+/// An advisor that checks if an attribute is deprecated
+pub struct DeprecatedAdvisor;
+impl Advisor for DeprecatedAdvisor {
+    fn advise(
+        &mut self,
+        sample: &SampleRef<'_>,
+        registry_attribute: Option<&Attribute>,
+        _registry_group: Option<&ResolvedGroup>,
+    ) -> Result<Vec<Advice>, Error> {
+        match *sample {
```
jsuereth, line 54: Is there a way we can avoid having Sample *both* have a known lifetime and use `*` here?

I haven't looked into the details, but if it's been tricky, we can track this as a cleanup opportunity later.

crates/weaver_live_check/src/sample_span.rs
```diff
@@ -0,0 +1,110 @@
...
+    #[serde(default)]
+    pub span_events: Vec<SampleSpanEvent>,
+    /// SpanLinks
+    #[serde(default)]
+    pub span_links: Vec<SampleSpanLink>,
+    /// Live check result
+    pub live_check_result: Option<LiveCheckResult>,
+}
+
+impl LiveCheckRunner for SampleSpan {
```
jsuereth, line 33: Ah, I see, we use this trait to figure out how to nest into possible "sample" things we sent to advice.

I like this.

crates/weaver_live_check/src/lib.rs
```diff
@@ -0,0 +1,341 @@
...
+    /// A sample span link
+    SpanLink(SampleSpanLink),
+    /// A sample resource
+    Resource(SampleResource),
+}
+
+/// Represents a sample entity with a reference to the inner type
+#[derive(Debug, Clone, PartialEq, Serialize)]
+#[serde(rename_all = "snake_case")]
+pub enum SampleRef<'a> {
```
jsuereth, line 108: Given how small this is - You can likely pass-by-value to Advice. I'll add comments elsewhere, where I think it can work.

crates/weaver_live_check/src/advice.rs
```diff
@@ -0,0 +1,345 @@
...
+
+/// Embedded default live check jq preprocessor
+pub const DEFAULT_LIVE_CHECK_JQ: &str = include_str!("../../../defaults/jq/advice.jq");
+
+/// Provides advice on a sample
+pub trait Advisor {
+    /// Provide advice on a sample
+    fn advise(
+        &mut self,
+        sample: &SampleRef<'_>,
```
jsuereth, line 39: This should be able to take `SampleRef<'_>` directly.  It's a common Rust pattern where you have a `{thing}Ref` struct that is efficient to just send as a value.

crates/weaver_live_check/src/sample_attribute.rs
```diff
@@ -0,0 +1,298 @@
...
+                        advice_level: AdviceLevel::Information,
+                    });
+                }
+            }
+        }
+
+        // run advisors on the attribute
+        for advisor in live_checker.advisors.iter_mut() {
+            if let Ok(advice_list) = advisor.advise(
+                &SampleRef::Attribute(self),
```
jsuereth, line 200: Here, you don't need to `&` the `SampleRef` if you pass the value, instead of reference.  You're effectively creating a new Ref for every call, but since `SampleRef` is so small/cheap this is fine.

crates/weaver_live_check/src/sample_resource.rs
```diff
@@ -0,0 +1,37 @@
...
+    pub attributes: Vec<SampleAttribute>,
+    /// Live check result
+    pub live_check_result: Option<LiveCheckResult>,
+}
+
+impl LiveCheckRunner for SampleResource {
+    fn run_live_check(&mut self, live_checker: &mut LiveChecker, stats: &mut LiveCheckStatistics) {
+        let mut result = LiveCheckResult::new();
+        for advisor in live_checker.advisors.iter_mut() {
+            if let Ok(advice_list) = advisor.advise(&SampleRef::Resource(self), None, None) {
```
jsuereth, line 26: Similar comment for `SampleRef` not needing `&` here.

jsuereth (approved):
Really liking how this shaped up!

One rust architecture nitpick we should address for the core Advisor interface, but looks good to me!

crates/weaver_live_check/src/advice.rs
jerbly: I think that would be really neat, yes. Basically almost everywhere where we refer to a file or path for input, we should refer to a scheme like this.

jerbly:
> One rust architecture nitpick we should address for the core Advisor interface, but looks good to me!

@jsuereth - I got carried away with my refs! Fixed that up.

jerbly:
Thanks @jsuereth - I need an approval from @lquerel before I can merge.

lquerel (approved):
Looks great. Thank you!

PR #645, opened 2025-03-21
Add support for annotations on attributes and groups

lquerel:
See https://github.com/open-telemetry/semantic-conventions/issues/1462#issuecomment-2655314451 for more context.

This PR adds an `annotations` field to the definition of attributes and groups. In this initial iteration, annotations are opaque to Weaver and represented as a key→YAML-value map.

Annotations become accessible, after registry resolution, within JQ expressions, Jinja templates, and the Rego engine.

In the future, we may introduce a new group type allowing users to define the structure of annotations directly within the semconv registry.


crates/weaver_resolved_schema/src/registry.rs
```diff
@@ -131,6 +131,10 @@ pub struct Group {
     /// This fields is only used for event groups.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub body: Option<AnyValueSpec>,
+    /// Annotations for the group.
+    #[serde(default)]
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub annotations: Option<HashMap<String, YamlValue>>,
```
lmolkova, line 137: could you please update https://github.com/open-telemetry/weaver/blob/main/schemas/semconv.schema.json and the syntax.md too?

lquerel: Done

lmolkova (approved):
Schema changes look good to me!

(Rust code too, but I'm not qualified)

jerbly (approved):
LGTM

PR #651, opened 2025-03-23
Upgrade regorus to 0.4

jerbly:
This will fix #646.

💥 BREAKING CHANGE 💥 - Upgrade to version 0.4.0 of regorus [requires all v0 policies to be modified](https://github.com/microsoft/regorus/pull/373). Policy upgrade instructions [here](https://www.openpolicyagent.org/docs/latest/v0-upgrade/#upgrading-rego) may help.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,7 +52,7 @@ jobs:
       name: Check MSRV
       runs-on: ubuntu-latest
       env:
-        MSRV: "1.81.0"  # MSRV 
+        MSRV: "1.85.0"  # MSRV 
       steps:
         - name: Checkout repository
           uses: actions/checkout@v4
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ jobs:
     permissions:
       contents: none
     name: CI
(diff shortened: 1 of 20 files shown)


jerbly:
Hit a snag with this PR:

regorus 0.3+ has a dependency on [constant_time_eq-0.4.2 which requires edition 2024](https://github.com/cesarb/constant_time_eq/blob/master/Cargo.toml).

[cargo-check-external-types](https://github.com/awslabs/cargo-check-external-types/blob/main/rust-toolchain.toml) latest version requires a nightly build, the last known working is `nightly-2024-10-29`. This version does not support edition 2024. I've tried later ones and they fail:

>The version of rustdoc being used produces JSON format version 37, but this tool requires format version 30. This can happen if the locally installed version of rustdoc doesn't match the rustdoc JSON types from the `rustdoc-types` crate.
>
>If this occurs with the latest Rust nightly and the latest version of this tool, then this is a bug, and the tool needs to be upgraded to the latest format version.
>
>Otherwise, you'll need to determine a Rust nightly version that matches this tool's supported format version (or vice versa).

So, as written, this is a bug in cargo-check-external-types. This leaves us in a stalemate.

lquerel (approved):
LGTM

PR #671, opened 2025-04-01
docs(changelog): should be deprecated.note to use as of weaver 0.13.0

trentm:
I was updating the semconv package for OTel JS to the new version of weaver, and I believe the mention of `deprecated.brief` in the weaver 0.13.0 change log is wrong: should be `deprecated.note`.


jerbly:
I believe the changelog is accurate. The 0.13.0 release made use of the existing `brief` field. However, we quickly found that this wasn't going to work and introduced `note` under `deprecated` in 0.13.2. This is documented in more detail in the PR mentioned in the changelog: #602

trentm:
Okay, thanks for the clarification.

PR #681, opened 2025-04-05
Cli doc updates

larrys:
This syncs the cli help output with doc page.
Also update a few codeblocks to have a format.


jerbly (approved):
Thanks for the doc update!

PR #682, opened 2025-04-06
Add tab completions from clap_complete.

larrys:
I am not a rust developer, so there might be some syntax that needs work.

I tested this locally with zsh on macOS, and it works.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to this project will be documented in this file.
 
+# [TBD] - TBD
+
+* Add shell completion functionality. ([#682](https://github.com/open-telemetry/weaver/pull/682) by @larrys)
+
 ## [0.14.0] - 2025-04-10
 
 What's changed
@@ -15,7 +19,6 @@ What's changed
 * Make `type` property required on the semantic convention group (behind `--future` flag). ([#611](https://github.com/open-telemetry/weaver/pull/611) by @lmolkova).
 * Exclude attributes declared with `code_generation.exclude` annotations in `semconv_attributes` and other JQ attribute helpers. ([#662](https://github.com/open-telemetry/weaver/pull/662) by @lmolkova)
 * Sort metrics by name in all JQ helpers. ([#573](https://github.com/open-telemetry/weaver/issues/573) by @lmolkova)
-* Add shell completion functionality. ([#682](https://github.com/open-telemetry/weaver/pull/682) by @larrys)
 
 ## [0.13.2] - 2025-02-13
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ What's changed
 * Make `type` property required on the semantic convention group (behind `--future` flag). ([#611](https://github.com/open-telemetry/weaver/pull/611) by @lmolkova).
 * Exclude attributes declared with `code_generation.exclude` annotations in `semconv_attributes` and other JQ attribute helpers. ([#662](https://github.com/open-telemetry/weaver/pull/662) by @lmolkova)
 * Sort metrics by name in all JQ helpers. ([#573](https://github.com/open-telemetry/weaver/issues/573) by @lmolkova)
+* Add shell completion functionality. ([#682](https://github.com/open-telemetry/weaver/pull/682) by @larrys)
 
 ## [0.13.2] - 2025-02-13
 
--- a/src/main.rs
+++ b/src/main.rs
@@ -108,7 +108,13 @@ fn run_command(cli: &Cli, log: impl Logger + Sync + Clone) -> ExitDirectives {
         Some(Commands::Registry(params)) => semconv_registry(log.clone(), params),
         Some(Commands::Diagnostic(params)) => diagnostic::diagnostic(log.clone(), params),
         Some(Commands::Completion(completions)) => {
-            generate_completion(&completions.shell);
+            if let Err(e) = generate_completion(&completions.shell, &completions.completion_file) {
+                log.error(&e);
+                return ExitDirectives {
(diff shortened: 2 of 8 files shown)


jsuereth (commented):
This looks great!

Two larger questions:

- Is there anyway we can test the command works?  
- Can you document how to use this for autocomplete in Bash, Zsh, etc.?  It's fine to have a section in weaver installation that links out to clap_complete docs if needed.

tests/completion_check.rs
```diff
@@ -0,0 +1,18 @@
...
+#[test]
+fn test_generate_completion() {
+    let mut cmd = Command::cargo_bin("weaver").unwrap();
+    let output = cmd
+        .arg("completion")
+        .arg("bash")
+        .timeout(std::time::Duration::from_secs(60))
+        .output()
+        .expect("weaver__completion");
+    assert!(output.status.success());
```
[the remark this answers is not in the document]
larrys: That's the weaver__complete contains part. It does not check the whole output just that part is there and exits cleanly.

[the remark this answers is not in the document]
jerbly: Do you mean `.expect("weaver__completion")` if so, that's not what `.expect()` does: https://doc.rust-lang.org/stable/std/result/enum.Result.html#method.expect

You need to read `output.stdout`. Here's an example: https://github.com/open-telemetry/weaver/blob/b59a73d1aa265940417e1dcb5a1fd9810d83b998/tests/registry_check.rs#L43

src/main.rs
```diff
@@ -77,6 +80,11 @@ impl CmdResult {
 fn main() {
     let cli = Cli::parse();
 
+    if let Some(Commands::Completion(completions)) = &cli.command {
+        generate_completion(&completions.shell);
+        return;
```
lquerel, line 85: Is there a good reason not using the `run_command` match statement to process this specific command?

lquerel (commented):
I agree with the other comments. I think this is a nice addition to Weaver and improves the overall user experience. I have just one minor remark, which I’ve included as a comment.

src/main.rs
larrys: As I stated, I am not a Rust developer. So I am going off help from CoPilot and googling. Thanks for the pointer. I will adjust and update to use that.

src/main.rs
```diff
@@ -104,6 +107,13 @@ fn run_command(cli: &Cli, log: impl Logger + Sync + Clone) -> ExitDirectives {
     let cmd_result = match &cli.command {
         Some(Commands::Registry(params)) => semconv_registry(log.clone(), params),
         Some(Commands::Diagnostic(params)) => diagnostic::diagnostic(log.clone(), params),
+        Some(Commands::Completion(completions)) => {
+            generate_completion(&completions.shell);
+            return ExitDirectives {
+                exit_code: 0,
+                quiet_mode: cli.quiet,
```
jerbly, line 114:
```suggestion
                quiet_mode: true,
```
Quiet mode needs to be true to avoid weaver writing `Total execution time: 0.010793s` to stdout and spoiling the completion script.

tests/completion_check.rs
```diff
@@ -0,0 +1,23 @@
...
+    let output = cmd
+        .arg("completion")
+        .arg("bash")
+        .timeout(std::time::Duration::from_secs(60))
+        .output()
+        .expect("failed to execute process");
+    assert!(output.status.success());
+    let stdout = String::from_utf8(output.stdout).expect("Invalid UTF-8");
+    assert!(
+        stdout.contains("weaver__completion"),
```
jerbly, line 20: To really regression test this well we would need to compare the expected output of `generate_completion` with what was captured on `stdout` and they should be identical.

To do this we would need to add an option to `weaver completion` like `--completion-path` where `generate_completion` would write to directly.

Then in your test you would run `weaver completion --completion-path tempfile.txt`, then read back in `tempfile.txt`, then run your test capturing stdout and compare.

The problem is that weaver uses stdout for other diagnostic information and we have problems separating the artifact delivered to stdout from the diagnostics. See https://github.com/open-telemetry/weaver/issues/184#issuecomment-2781465337

jerbly: Sorry that's not what I meant. We're trying to guard against weaver accidentally sending any text to stdout before or after the completion text. If it does then the completion script is broken.

My idea was to have an optional file path config item, but KEEP the default stdout behaviour. You then run the weaver command TWICE. The first time to capture the expected output by using the file path option. Then the second time to capture the observed output using stdout. Then check expected == observed.

larrys: Ahhh... That makes sense. I'll it updated.

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ What's changed
 * Make `type` property required on the semantic convention group (behind `--future` flag). ([#611](https://github.com/open-telemetry/weaver/pull/611) by @lmolkova).
 * Exclude attributes declared with `code_generation.exclude` annotations in `semconv_attributes` and other JQ attribute helpers. ([#662](https://github.com/open-telemetry/weaver/pull/662) by @lmolkova)
 * Sort metrics by name in all JQ helpers. ([#573](https://github.com/open-telemetry/weaver/issues/573) by @lmolkova)
+* Add shell completion functionality. ([#682](https://github.com/open-telemetry/weaver/pull/682) by @larrys)
```
jerbly, line 18: 0.14.0 was released so this line needs to be moved up to the top under a `## Next` heading please.

larrys: Ah, I missed that when I rebased my branch. Thanks for pointing it out.

jerbly (approved):
Thanks for sticking with me on the testing! :)

This will be a great ux addition to weaver!

PR #683, opened 2025-04-06
Allow signal groups to declare associations on entities.

jsuereth:
First step of cleaning up Resource and modelling for semantic conventions.

Allows `entity_associations` to be expressed for all signals.  This a list of strings, expected to match the `name` field of the appropriate group.

Matches the [proposal](https://docs.google.com/document/d/1GmB0kDx9opHxjj2oy6A4h0fqE8UVW-3ebWgp6zhtbVE/edit?tab=t.0).

Additoinally, this makes JSON-based output tests "pretty print" for easier diff-based verification of changes.

Note: This does NOT rename `resource` to `entity`, that will be a different PR.


lquerel (approved):
LGTM

PR #702, opened 2025-04-16
Use the re-export of crossterm from ratatui

jerbly:
This will fix #684 

Crossterm 0.29.0 is not compatible with ratatui 0.29.0. I've switched this to use the re-export of crossterm instead.


lquerel (approved):
LGTM

Issue #703, opened 2025-04-16
Support `any` type values in log/event attributes


dyladan:
Currently all semantic conventions need to declare what type they are (e.g. `string`, `boolean`, etc). In some cases such as `feature_flag.result.value` we may not know what type the value of the attribute is. In the past, body fields were used to solve this use case, but body is being de-emphasized in favor of attributes. Because we don't know the type, the only thing we can currently do is JSON stringify the value and have another attribute that disambiguates the original type (e.g. is an attribute containing string value the string `"6"` or the number `6`?).

I would like to be able to declare the type of `feature_flag.result.value` to be `any`/`anyvalue`/`undefined`/`unknown` type (i'm fine with any name), indicating that the type of the value is unknown. The type information is carried in OTLP as part of the anyvalue structure, so it should be possible to determine on the backend.

This addresses part of https://github.com/open-telemetry/semantic-conventions/issues/1669 but does not completely solve the complex attribute problem which will need to be addressed later.

We would like to use it in https://github.com/open-telemetry/semantic-conventions/pull/1990

PR #707, opened 2025-04-21
Support for attributes of type `any` and `template[any]`

lquerel:
This PR introduces support for attributes of type `any` and `template[any]`, enabling more flexible handling of attribute values in the semconv.

Closes: #703


jerbly (commented):
I think we need a way to compare compatibility of `PrimitiveOrArrayTypeSpec`. In the TypeAdvisor near your change, we compare the sample type with the semconv type with a simple inequality:
```rust
if attribute_type != semconv_attribute_type {
```
But now if `attribute_type` is `string` and `semconv_attribute_type` is `any` then these types are compatible.

crates/weaver_resolver/data/registry-test-15-attr-any/registry/registry-messaging.yaml
```diff
@@ -0,0 +1,19 @@
+groups:
+  - id: registry.test
+    type: attribute_group
+    brief: 'Used for testing any and template[any]'
+    attributes:
+      - id: test.any
+        stability: stable
+        type: any
```
lmolkova, line 8: we'll probably need to filter out `any` attributes in semconv jinja templates on Java

https://github.com/open-telemetry/semantic-conventions-java/blob/3c7edf34f5b548e64a63793e3772da40754d874c/buildscripts/templates/registry/incubating_java/IncubatingSemanticAttributes.java.j2#L10.

Not sure if there are other languages that have similar API and won't be able to define a constant for `any` attribute (yet).

lmolkova (approved):
Schema changes LGTM!

Let's bring it up on the semconv tooling meeting - we'll need to communicate this to maintainers - some languages need to add support for `any` attributes or would generate invalid code.

jsuereth (approved):

crates/weaver_resolver/data/registry-test-15-attr-any/registry/registry-messaging.yaml
lmolkova: Created https://github.com/open-telemetry/semantic-conventions-java/issues/208 and https://github.com/open-telemetry/opentelemetry-go/issues/6691 to give maintainers a heads up

beeme1mr (approved):

dyladan:
What's the current status of this? Is it waiting more reviews or blocked on something?

lmolkova:
@dyladan ETA to get this one merged is this week and release is planned for the next week.

lquerel:
> I think we need a way to compare compatibility of `PrimitiveOrArrayTypeSpec`. In the TypeAdvisor near your change, we compare the sample type with the semconv type with a simple inequality:
> 
> ```rust
> if attribute_type != semconv_attribute_type {
> ```
> 
> But now if `attribute_type` is `string` and `semconv_attribute_type` is `any` then these types are compatible.

@jerbly I made this change

jerbly (approved):
Thanks! Good to go!

PR #710, opened 2025-04-24
Support for simplified Template Type Examples

jerbly:
Support for simplified Template Type Examples format where the key/attribute name is no longer included.

The outcome of a discussion related to this issue: https://github.com/open-telemetry/semantic-conventions/issues/1740 was to simply remove the keys from the example and therefore treat `template[string]` like `string` and `template[int[]]` like `int[]` etc. See https://github.com/open-telemetry/semantic-conventions/pull/2175

For backwards compatibility, warnings are issued for two previously acceptable combinations:
- `template[string[]]` with `string` example e.g. `"foo"`
- `template[string[]]` with `string[]` example e.g. `["foo","bar"]`

Examples for `template[string[]]` must always be e.g. `[ ["foo","bar"] ]`


lmolkova (approved):
LGTM from semconv side, thank you!

lquerel (approved):
LGTM

PR #714, opened 2025-04-28
Rename `resource` group type to `entity`.

jsuereth:
This is part of the move to Entity as a grouping of attributes on Resource. An Entity represents a named set of attributes with its own versioning defined by weaver / semantic conventions. Combined with the previous `entity_associations` work, this unlocks validating Resource in `live_check`.

Note: This is **breaking** for all templates that directly use `group.type`.
      While JQ helpers have been updated to support the chagne, direct
      group.type usage will see `entity` in the future.


lquerel (approved):
LGTM

lmolkova (approved):

PR #719, opened 2025-04-29
Fix release CI/CD

jsuereth:
Unblock CI/CD by moving to use latest ubuntu release images.

Github Actions is blocking Ubuntu 20.04 as being out of service.


jsuereth:
FYI - https://github.com/cncf/foundation/issues/1007

PR #721, opened 2025-04-30
Switch to log crate

### LINKED ISSUE #184: When resolving to stdout do not display status

jerbly:
All logging and diagnostics now go to `stderr`. Override diagnostics with `--diagnostic-stdout`.

Fixes #184 

This achieves everything but the animated loading lines using `log` and `env_logger` which are basically the rust standard. This means we no longer pass the logger around and just use standard log macros `info!("did a thing")` .
To keep the look and feel of paris, we import the formatting functions and then do things like this: `warn!("{}", colorize_string("<b><green>TEST"));`. For the most common there are now simple functions like: `log_success()`, `log_error()` etc.

The separation of `stderr` and `stdout` also means we don't really need `--quiet` anymore. I've left it in though. I've also changed the `ExitDirective` so you can exit a command positively via an `Ok()` and still print out any collected warnings.


jsuereth (approved):

crates/weaver_forge/src/debug.rs
```diff
@@ -111,7 +110,10 @@ mod tests {
 
     #[test]
     fn test_print_dedup_errors() {
-        let logger = weaver_common::TestLogger::new();
+        // Setup TestLog to capture log messages and keep a reference to it
+        let test_log = weaver_common::TestLog::new();
+        test_log.clone().init().expect("Failed to init logger");
```
lquerel, line 115: This split between the constructor and the init method seems really error-prone to me. Couldn’t we have a constructor that performs the initialization? Otherwise, there’s a high chance someone will misuse the logger.

lquerel (approved):
I made a comment about the logger initialization, but otherwise, it looks good to me.

crates/weaver_forge/src/debug.rs
jerbly: Fixed in #727

PR #722, opened 2025-05-01
Fix live check error handling

jerbly:
Errors from advisors were not propagated up properly. Fixed and added a bad rego policy to regression test.


crates/weaver_live_check/src/lib.rs
```diff
@@ -337,5 +341,9 @@ impl LiveCheckStatistics {
 /// Samples implement this trait to run live checks on themselves
 pub trait LiveCheckRunner {
     /// Run the live check
-    fn run_live_check(&mut self, live_checker: &mut LiveChecker, stats: &mut LiveCheckStatistics);
+    fn run_live_check(
+        &mut self,
+        live_checker: &mut LiveChecker,
+        stats: &mut LiveCheckStatistics,
+    ) -> Result<(), Error>;
```
lmolkova, line 348: it's still a rust complain @jsuereth, if the result with error was more like an exception, it would take an effort to ignore it 😅

lmolkova (approved):
Thanks a lot!!

crates/weaver_live_check/src/lib.rs
jsuereth: It did take effort to ignore it.  All the implementations of this trait were explicitly taking the ok state and ignoring the error state of the result.

Our clippy would fire if someone hadn't explicitly decided to do that and tried to implicitly ignore errors.

This is similar to the issue of "catching too much" with checked exceptions where decide to catch everything rather than think through what to propagate.

The only complaint is have on rust is that "if let Ok(...) =" should probably be something we scrutinize more in code review as the equivalent of ignoring exceptions via "try { } catch(anything) { ...ignore...}"

jsuereth (commented):
My only concern here is one of developer experience:

If I have many live check policies with errors, will I get all of them reported?  Do we assume only one rego based checker that can report errors?

jerbly:
> My only concern here is one of developer experience:
> 
> If I have many live check policies with errors, will I get all of them reported?

One of the reasons I didn't spot this until now is that rego doesn't produce an error of this kind until you call `eval_rule` - whereas syntax errors are produced when you call `parse` way back when the policies are loaded. `eval_rule` only returns the first encountered error.
```rego
deny contains make_advice("foo", "violation", attribute_name, "bar") if {
	attribute_name := "foo"
	not attribu1te_name
	not banana
}
```
Only gives:
```
  │ --> crates/weaver_live_check/data/policies/bad_advice/bad.rego:8:9
  │   |
  │ 8 |         not attribu1te_name
  │   |         ^
  │ error: use of undefined variable `attribu1te_name` is unsafe
```

> Do we assume only one rego based checker that can report errors?

At the moment the design only allows for one Rego advisor to which we send all signals.

lquerel (approved):
LGTM

jsuereth:
> eval_rule

@jerbly Can you open a ticket to look into this and tag it under the UX / ease of use label?

I think we'll want to visit this in the future.

crates/weaver_live_check/src/live_checker.rs
```diff
@@ -535,7 +535,7 @@ mod tests {
 
         let mut stats = LiveCheckStatistics::new(&live_checker.registry);
         for sample in &mut samples {
-            sample.run_live_check(&mut live_checker, &mut stats);
+            let _ = sample.run_live_check(&mut live_checker, &mut stats);
```
jsuereth, line 538: Nit: in the future we should not ignore the failures here, even for tests.  Instead the test can return `Result`, and you can `?` your way out of it.

jsuereth (approved):

crates/weaver_live_check/src/live_checker.rs
jerbly: Fixed in #727

PR #728, opened 2025-05-05
Live check metrics

jerbly:
Adds support for Metrics to live-check. The nested nature of this data brought about further refactoring internally but also a change to the rego interface. `input` now has `sample` and `registry_attribute` or `registry_group` - this allows the policy to work within the nested context. For example, providing some advice on an exemplar in a data-point in a metric by referring to the metric's defined unit. This is documented in the readme and an example policy is included in the tests.

image
image

Includes support for Exemplars:

image


crates/weaver_live_check/src/lib.rs
```diff
@@ -110,6 +117,10 @@ pub enum SampleRef<'a> {
     SpanLink(&'a SampleSpanLink),
     /// A sample resource
     Resource(&'a SampleResource),
+    /// A sample metric
+    Metric(&'a SampleMetric),
+    /// A sample number data point
+    NumberDataPoint(&'a NumberDataPoint),
```
jsuereth, line 123: Interesting - Is the convention going to be check at the highest level, but higher-level things don't include lower level details?

E.g. NumberDataPoint is part of a Metric and has metric identifying things, but SampleMetric wouldn't?


Also, can you rename to `SampledNumberDataPoint`?

jerbly: This area is not quite right yet. A NumberDataPoint without the information from the Metric is not so useful, but I need a way to pass it in around in a consistent way.

Yes, good call on the renaming. Will do.

crates/weaver_live_check/src/sample_metric.rs
```diff
@@ -0,0 +1,164 @@
...
+    Histogram(Vec<SampleHistogramDataPoint>),
+    /// Exponential histogram data points
+    ExponentialHistogram(Vec<SampleExponentialHistogramDataPoint>),
+    /// Summary data points
+    Summary(Vec<SampleSummaryDataPoint>),
+}
+
+/// Represents a single data point of a metric
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub struct SampleNumberDataPoint {
```
jsuereth, line 31: I may be missing something, but I think we need the metric available when we live-check sample points.  For example, checking if the type of point is int vs. double or checking if attributes match required/recommended (or are missing) can only be done here.

defaults/policies/live_check_advice/otel.rego
```diff
@@ -40,6 +40,16 @@ deny contains make_advice(advice_type, advice_level, value, message) if {
 	message := "Does not match name formatting rules"
 }
 
+# checks metric name format
+deny contains make_advice(advice_type, advice_level, value, message) if {
```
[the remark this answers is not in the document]
jerbly: This is already done internally in rust rather than in rego since I assumed this was a fundamental thing. See the `attribute_required` advice under each data_point in the screenshot above. This is at the data_point level since the attributes are provided with each point.

crates/weaver_live_check/src/lib.rs
jerbly: I'm looking into the consistency of the framework. For each data_point, I want to be able to provide at advice at that level (checking attributes are required for example), but I need context from the parent group. I'm hoping to get this change done today. This will change rego so the input is a tuple of (sample,semconv).

jerbly: Just pushed this with docs to explain and a rego example.
```rego
# This example shows how to use the registry_group provided in the input.
# If the metric's unit is "By" the value in this data-point must be an integer.
deny contains make_advice(advice_type, advice_level, value, message) if {
	input.sample.number_data_point
	value := input.sample.number_data_point.value
	input.registry_group.unit == "By"
	value != floor(value) # not a good type check, but serves as an example
	advice_type := "invalid_data_point_value"
	advice_level := "violation"
	message := "Value must be an integer when unit is 'By'"
}
```

lquerel (changes requested):
Thanks for this PR. Adding metrics support will significantly enhance the live-check feature. I’ve made several suggestions. There are certain types we can avoid duplicating (especially the leaf enums). For the other structures, I understand that you’re adding at least one custom field for the results. It’s frustrating to see all this duplicated code, but unfortunately, I don’t see a straightforward alternative solution.

It would also be good to remove instances of `&Rc<T>` or `Rc<Rc<T>>`, as they seem semantically redundant and not very idiomatic.

crates/weaver_live_check/src/advice.rs
```diff
@@ -178,6 +237,83 @@ impl Advisor for TypeAdvisor {
...
+            SampleRef::Metric(sample_metric) => {
+                // Check the instrument and unit of the metric
+                let mut advice_list = Vec::new();
+                if let Some(semconv_metric) = registry_group {
+                    match &sample_metric.instrument {
+                        SampleInstrument::Summary => {
+                            advice_list.push(Advice {
+                                advice_type: "legacy_instrument".to_owned(),
+                                value: Value::String(sample_metric.instrument.to_string()),
+                                message: "`Summary` Instrument is not supported".to_owned(),
```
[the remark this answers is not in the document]
jsuereth: You cannot produce Summary from OTEL SDKs.  It only exists to support legacy metric systems, and it's inconvertible to OpenTelemetry "primary" concepts.

I think for weaver, it's unsupported.  We won't allow folks to define summary metrics in our YAML.

PR #731, opened 2025-05-07
Update semconv JSON schema for entities

smith:
- Add `entity_associations` for metrics
- Update `display_name` description to be more generic
- Replace resource with updated entity schema

### DIFF
--- a/schemas/semconv.schema.json
+++ b/schemas/semconv.schema.json
@@ -18,6 +18,13 @@
 					{
 						"allOf": [
 							{
+								"$ref": "#/$defs/EntitySemanticConvention"
+							}
+						]
+					},
+					{
+						"allOf": [
+							{
 								"$ref": "#/$defs/SpanSemanticConvention"
 							}
 						]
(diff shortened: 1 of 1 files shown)


schemas/semconv.schema.json
```diff
@@ -99,6 +106,44 @@
...
+					"const": "entity"
+				},
+				"attributes": {
+					"type": "array",
+					"items": {
+						"$ref": "#/$defs/Attribute"
+					},
+					"description": "list of identifying attributes for the entity type"
+				},
+				"descriptive_attributes": {
```
jsuereth, line 130: This doesn't exist yet.

Also I'm going to be talking about this in today's tooling sync, but we may actually want descriptive to be an aspect of attribute, vs. a separate bundle, both for the `ref` aspect of weaver and the `diff` aspect.

schemas/semconv.schema.json
```diff
@@ -99,6 +106,44 @@
...
+					"type": "array",
+					"items": {
+						"$ref": "#/$defs/Attribute"
+					},
+					"description": "list of descriptive attributes for the entity type"
+				},
+				"relationships": {
```
jsuereth, line 137: This also doesn't exist in weaver yet, so we should hold off adding it.

schemas/semconv.schema.json
```diff
@@ -458,6 +503,11 @@
 				"unit": {
 					"type": "string",
 					"description": "The unit in which the metric is measured in."
+				},
+				"entity_associations": {
```
jsuereth, line 507: We can actually add this to Event + Span as well.

jsuereth (commented):
Thanks for updating this!  A few comments on what we can/should support right now.

smith:
@jsuereth I'll keep the parts that don't exist yet in this PR and move it to draft. I'll open a new PR with the remaining changes that can be made now. Thanks for looking!

smith:
@jsuereth I added role as an enum in 590113d. Let me know if things look right here now, as I also took out the things that don't exist yet.

jsuereth (commented):
This looks good now, but you probably need to merge in main, where `role` was already added.

jsuereth (approved):

PR #735, opened 2025-05-09
Allow attributes to be identifying or descriptive.

jsuereth:
This is the next phase of https://github.com/open-telemetry/semantic-conventions/issues/2246.

What this does:

- Add `role` field to Attribute with options `Identifying` or `Descriptive`
- Adds lineage tracking for `role` changes

What this does not do:

- Update yaml schema for model files (Didn't have time to learn this one yet)
- Add any weaver enforcement of rules around usage. I plan to add this in semconv initially, but not sure how much limits we'll have in weaver.
- Add any JQ or JINJA helpers for `role`.

## Why role

Why do we call this `role`?  I did some investigation into names and actually thought LLM captured some considerations well, here's (snipped) output:

> Why terms like "Type" or "Kind" might be less suitable:
>
> AttributeType: This is commonly used to refer to the data type of the attribute (e.g., string, integer, boolean) (Source 4.2), and using it for this dimension could cause confusion.
> AttributeKind: OpenTelemetry already uses "Kind" for spans (e.g., SpanKind with values like SERVER, CLIENT), so using it here might lead to ambiguity.
>
> ...
>
> Rationale: "Role" is a widely understood term in data modeling, where attributes play specific roles (e.g., key, descriptor). This name signifies the function or part an attribute plays in defining or characterizing the signal or resource.
>
> ...


crates/weaver_resolver/src/registry.rs
```diff
@@ -793,23 +809,37 @@ mod tests {
...
 
+            // Write observed output.
+            let observed_attr_catalog_file = OpenOptions::new()
+                .create(true)
+                .write(true)
+                .open(observed_output_dir.join("attribute-catalog.json"))
+                .expect("Failed to open observed output file");
+            serde_json::to_writer_pretty(observed_attr_catalog_file, &observed_attr_catalog)
+                .expect("Failed to write observed ouptut.");
+            // Compare values
```
lquerel, line 820: What is the purpose of this? Debugging?

lquerel (approved):
Made 2 minor comments, otherwise LGTM

crates/weaver_resolver/src/registry.rs
jsuereth: no, updating the test values for some of these really large files is easier to look at the new and validate, then copy over.

lquerel: I see. Thanks

crates/weaver_resolved_schema/src/attribute.rs
```diff
@@ -87,6 +89,10 @@ pub struct Attribute {
     /// Note: This is only used in a telemetry schema specification.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub value: Option<Value>,
+    /// Whether the attribute is identifying or descriptive.
```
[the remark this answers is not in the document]
jsuereth: Until opentelemetry.io is up-to-date with latest spec, I'd prefer not to link to the github URL directly, as those can be unstable.

crates/weaver_semconv/src/attribute.rs
```diff
@@ -84,6 +84,10 @@ pub enum AttributeSpec {
         prefix: bool,
         /// Annotations for the attribute.
         annotations: Option<BTreeMap<String, YamlValue>>,
+        /// Whether the attribute is identifying or descriptive.
+        #[serde(default)]
+        #[serde(skip_serializing_if = "Option::is_none")]
+        role: Option<AttributeRole>,
```
lmolkova, line 90: could you please update json schema and md in the https://github.com/open-telemetry/weaver/tree/main/schemas ?

jsuereth: Not sure you read my description, but I need help there if possible.  I actually don't really know/understand JSON schema. I spent several hours this weekend trying to learn but it'll take me a bit to figure that out if someone else can update it quicker.

lmolkova: I'm in the same boat, I'm usually copy-pasting something similar and modifying it to fit. 

I think you could use group type as an example https://github.com/open-telemetry/weaver/blob/78808d5f4eee8cbc66ad3645478b4f80f6c04e5b/schemas/semconv.schema.json#L164-L174

I believe a good way to do it is to generate the schema from rust model, but we don't do it yet.

The `syntax.md` is built on vibes and is meant to be readable, so the only somewhat formal change would be

```diff
- attributes ::= (id type brief examples | ref [brief] [examples]) [tag] stability [deprecated] [requirement_level] [sampling_relevant] [note] [annotations]
+ attributes ::= (id type brief examples | ref [brief] [examples]) [tag] stability [deprecated] [requirement_level] [sampling_relevant] [note] [annotations] [role]
```

in https://github.com/open-telemetry/weaver/blob/78808d5f4eee8cbc66ad3645478b4f80f6c04e5b/schemas/semconv-syntax.md?plain=1#L78

lmolkova (approved):

crates/weaver_semconv/src/attribute.rs
```diff
@@ -233,6 +241,24 @@ impl Display for AttributeType {
     }
 }
 
+/// The different roles for attributes in groups.
+#[derive(Serialize, Deserialize, Debug, Default, Clone, Eq, PartialEq, Hash, JsonSchema)]
+#[serde(rename_all = "snake_case")]
+pub enum AttributeRole {
+    /// The attribute is considered identifying for the signal it is associated with.
+    #[default]
```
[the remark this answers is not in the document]
jsuereth: It's not actually optional.  All attributes are implicitly identifying or descriptive.  This is foundational for metrics, and I think could improve logs/spans.

I think it actually does make sense across signals.  We can discuss more wed. if you'd like.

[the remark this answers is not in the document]
lmolkova: I don't think we should default to identifying then. Look at the HTTP spans - https://github.com/open-telemetry/semantic-conventions/blob/main/docs/http/http-spans.md#http-server-span - I have a hard time thinking about most of them as identifying. Method and route/url could be, but the rest?

Also, having it on the attribute definition makes me pause. The same attribute could be identifying in one place and descriptive in the other. Like user_agent.original is descriptive on http spans, but could  be identifying on browser resource.  I think it falls into the same bucket as req_level - per signal option.

crates/weaver_semconv/src/attribute.rs
jsuereth: Thanks for the pointers! Think it's updated now.

crates/weaver_semconv/src/attribute.rs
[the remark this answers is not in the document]
jsuereth: This is fair argument.  I want to follow up with @tedsuo on his proposed OTEP, but I'm fine treating this exactly like requirement level to begin with.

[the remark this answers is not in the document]
jsuereth: Ok, this is now fixed to be optional.

PR #739, opened 2025-05-14
Allow params to work in update_markdown command.

jsuereth:
This is a step towards deprecating the `--attributes-registry-base-url` command line argument.

This adds support for `--param` and `--params` in addition to actually using `weaver.yaml` `params` configuration in the `registry update-markdown` command.


lquerel (approved):
LGTM

PR #758, opened 2025-05-28
remove deprecated allow_custom_values from the code and yaml files

### LINKED ISSUE #520: Remove `allow_custom_values` flag on enums or change the default value to true
We're reserving the right to add values to enums in the future in the spec

>Things not listed in the above are not expected to remain stable via semantic
convention and are allowed (or expected) to change. A few examples:

> - The values of attributes
>    - An exception are existing values in lists of well-known values. However,
    new values can be added to such lists. Consumers should expect unknown
    values.

https://github.com/open-telemetry/opentelemetry-specification/blob/4f71c1642e541220e293bacf8c0d58f1fbcd3111/specification/versioning-and-stability.md?plain=1#L228-L231

But by default `allow_custom_values` is set to `false` resulting in "MUST be one of the following:" in the markdown and  some confusion.

I suggest removing the flag completely, but given it's a breaking change, we can also consider changing it to `true` by default.

trisch-me:
fix #520 

After final deprecation of build-tools we can finally remove `allow_custom_values` from the code


jerbly (approved):
LGTM

PR #767, opened 2025-06-01
Add support for metrics to emit

### LINKED ISSUE #748: Add Metrics to Emit
Currently `weaver registry emit` only supports spans. Since live-check now supports metrics, emit should too. 

Emit is really useful to see how your defined registry appears in your telemetry and observability tools. It also provides a useful round-trip test from Emit to Live-check.

jerbly:
Closes #748 and #752

Emits a metric for every one defined in the model.

Deficiencies:
- While the unit is defined you cannot tell whether this is `int` or `float`. So integer instruments are created in all cases.
- Examples are not given for the metric value, only attributes. So `1` is used in all cases.

Bug fixes:
This feature highlighted two issues which are fixed here with regression tests.
- Enum samples were always emitted as string, meaning ints, bools and doubles would fail with undefined_enum_variant in live-check.
- Otel illegal_namespace advice was raised by the rego policy when the attribute collision was with a deprecated attribute.


crates/weaver_emit/src/lib.rs
```diff
@@ -113,7 +157,24 @@ pub fn emit(
                 error: e.to_string(),
             })?;
 
-        // TODO Emit metrics
+        // Emit metrics
+        let meter_provider = match tracer_provider_config {
```
lquerel: `tracer_provider_config` could be renamed/generalized into `signal_provider_config`

crates/weaver_emit/src/lib.rs
```diff
@@ -113,7 +157,24 @@ pub fn emit(
...
+            ExporterConfig::Stdout => init_stdout_meter_provider(),
+            ExporterConfig::Otlp { endpoint } => {
+                init_meter_provider(endpoint).map_err(|e| Error::TracerProviderError {
```
lquerel: `TracerProviderError` -> `MetricProviderError` ?

crates/weaver_emit/src/spans.rs
```diff
@@ -35,7 +35,7 @@ fn otel_span_kind(span_kind: Option<&SpanKindSpec>) -> SpanKind {
 /// For the given attribute, return a name/value pair.
 /// Values are generated based on the attribute type and examples where possible.
 #[must_use]
-fn get_attribute_name_value(attribute: &Attribute) -> KeyValue {
+pub fn get_attribute_name_value(attribute: &Attribute) -> KeyValue {
```
lquerel: Should not that be moved into a dedicated module?

lquerel (approved):
I made few suggestions to better separate metrics from spans but otherwise LGTM

PR #769, opened 2025-06-05
[WIP] Add `imports` section in semconv to support references on metrics/events/entities from custom registries

lquerel:


crates/weaver_resolver/data/multi-registry/custom_registry/custom_registry.yaml
```diff
@@ -28,4 +28,7 @@ groups:
       - ref: auction.name
         requirement_level: recommended
       - ref: error.type
-        requirement_level: required
\ No newline at end of file
+        requirement_level: required
+
+imports:
```
lmolkova, line 34: slack discussion: we will have refs to signals within registry as well, so maybe we can either:
- have `metric_ref` under `groups` - there are some implementation constraints there
- have a different section name like `refs`

crates/weaver_resolver/data/multi-registry/custom_registry/custom_registry.yaml
```diff
@@ -28,4 +28,7 @@ groups:
...
+  - metric_ref: metric.example_counter
```
lmolkova: another branch of the slack discussion:
we probably want wildcard imports and cover attribute namespaces rather than groups.

E.g.
```yaml
imports:
  - attribute.db.*
  - attribute.server.*
  - attribute.error.*
  - attribute.http.*
  - attribute.url.*
```

syntax allows users to allowlist attributes in bulk without importing everything or explicitly allowing each attribute. 
Attribute groups are meaningless - they are grouping mechanism and we move attributes between them a lot.

there are not stability guarantees for attribute groups.

A similar syntax can be useful for bulk-importing metrics, entities, events

```yaml
imports:
- metric.db.*
- metric.http.*
- metric.system.*
- entity.host.*
- ...
```

Importing each attribute/span/event/metric/entity one-by-one will be very verbose and hard to use. Bulk import seems to be the main use-case.

crates/weaver_semconv/src/group.rs
```diff
@@ -106,6 +106,45 @@ pub struct GroupSpec {
...
+#[derive(Serialize, Deserialize, Debug, Clone, JsonSchema)]
+#[serde(deny_unknown_fields)]
+#[serde(untagged)]
+#[serde(rename_all = "snake_case")]
+pub enum GroupImport {
+    /// Imports a metric group from the imported registry.
+    MetricRef {
+        /// The ID of the metric group being referenced in the imported registry.
+        metric_ref: String,
+        // Additional overridable fields may be added in the future.
```
lmolkova: related to https://github.com/open-telemetry/weaver/pull/769/files#r2131292750 (bulk import)

importing without modification makes sense, but it would override the original signal definition without any means to distinguish imported from the original one. 

e.g. in service A I report `http.client.request.duration` with `url.template` when it talks to internal services B and C.
Service B reports `http.client.request.duration` without `url.template` because it's not supported or possible. 

How do I document these caveats if I have a singleton `http.client.request.duration` definition?

the proposal:

- `imports` allow to import without modification - original definitions are fully frozen and immutable. Original `id` describes only that signal and not its flavor. 
- supported modifications are done through existing `extends` mechanism or something similar. I.e.

```yaml
- id: my.service_a.metric.http.client.request.duration
  extends: metric.http.client.request.duration
  note: this a service a flavor of http.client.request.duration, it records url.template when talking to internal services
  attributes:
  - ref: url.template
     requirement_level:
       conditionally_required: SHOULD be recorded when calling service B or C

- id: my.service_b.metric.http.client.request.duration
  extends: metric.http.client.request.duration
  note: this a service b flavor of http.client.request.duration, it does not include url.template attribute since we can't record it there.
```

I.e. each modified version needs a unique id that can be used to render docs, generate code, build dashboards, run live checks.
We'd need to change what `extends` does for telemetry signals. Today it just inherits attributes and common group properties, but it can become stricter and require to preserve stability, name, type, etc.

From my point of view, having modifiable, but singleton definition is narrow and confusing.

lquerel: What matters to me is: 1) being able to import signals, and 2) being able to extend them. I like what you’re proposing, it seems clear to me. In this PR, I’ll focus only on the import part with wildcard support. 

@jsuereth and @jerbly are we all in agreement with that?

jerbly: Makes sense to me.

crates/weaver_resolver/data/multi-registry/custom_registry/custom_registry.yaml
lmolkova: based on https://github.com/open-telemetry/weaver/pull/769#discussion_r2131326386, we're moving in the direction of importing meaning importing without modification, so let's close this discussion and talk about bulk import here https://github.com/open-telemetry/weaver/pull/769#discussion_r2131292750

crates/weaver_semconv/src/group.rs
lmolkova: then we don't need signal-specific refs for import - they won't have signal-specific properties. 

I.e. we can have a nicer syntax like 
```yaml
imports:
- metric.db.*
- metric.http.*
- metric.system.*
- entity.*
- ...
```

jsuereth: I'm on board with this direction

lmolkova: one last thought on import by id or name. 

The above approach comes short for attributes - their id does not have `attribute` prefix. what if we imported by name but:

```yaml
- imports:
  - attributes:
    - http.*
    - db.*
    - system.*
  - metrics:
    - http.*
    - db.*
    - system.*
  - entities:
    - host
```

lquerel: I was thinking of accepting wildcard expressions in the `imports` list. Given that, is it really an issue that the attributes don’t start with `attribute.`?

lmolkova: given that attribute name can collide with metric or event name, I assume we need import to be able to differentiate them, or are you thinking that `imports: ["db.*"]` would import all signals and attributes with `db.*` name?

lquerel: Do we have attributes starting with `metric.`, `event.`, ... ?
I’m asking because I had just finished implementing the previous syntax :-)

lquerel: In your latest iteration, we import by name except for attributes, which might be confusing for custom registry authors.

lmolkova: maybe can should rename attribute `id` to `name`? 

In any case, if we do

```yaml
imports:
  - attribute.db.*
  - metric.db.*
```

it'll be confusing for authors because `attribute.` prefix doesn't exist anywhere

if we do 

```yaml
imports:
  - db.*
  - metric.db.*
```

it'll be confusing to authors and readers since it's not obvious that `db.*` only imports attributes.

I'm trying to create non-ambigous and consistent way for those who read or write conventions, with the 

```yaml
- imports:
  - attributes:
    - http.*
    - db.*
    - system.*
  - metrics:
    - http.*
    - db.*
    - system.*
  - entities:
    - host
```

they import by over-the-wire identifier they use in queries/dashboards. The `id` becomes internal semconv impl detail it should be

jerbly: @lquerel - What's the extent of the wildcard capability? Can we have `*.db.*` for example? Or, for people not using namespace dot notation, `db*`?

lquerel: @jerbly everything supported by globset is effectively supported by this imports section.

lquerel: @lmolkova 

```
- imports:
  - attributes:
    - http.*
    - db.*
    - system.*
  - metrics:
    - http.*
    - db.*
    - system.*
  - entities:
    - host
```

Unless I’m misinterpreting something in your previous messages, I see more or less the same level of confusion with this syntax. For attributes, the list under `attributes` contains IDs, while the list under `metrics` contains names.

refactoring_of_group_spec.patch
jerbly, line 1: Did you mean to include this file?

crates/weaver_semconv/src/semconv.rs
```diff
@@ -19,6 +19,10 @@ use weaver_common::result::WResult;
 pub struct SemConvSpec {
     /// A collection of semantic convention groups or [`GroupSpec`].
     pub(crate) groups: Vec<GroupSpec>,
+
+    /// A list of imports referencing groups defined in a dependent registry.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub(crate) imports: Option<Vec<GroupWildcard>>,
```
lmolkova: we should document what import does:
- it is a way to allowlist external specific semconv in yours (that's my understanding, not sure if it's a common one)
- when importing a group, does it also import all attributes the group depends on and all entities it's associated with? or do I  need to explicitly import such attribute before using it in my custom signals?
- is there a way to say "I allow all otel semconv" `imports: ["*"]`?
- what do I do with imported things after? generate docs? - we need an example :)

crates/weaver_semconv/src/semconv.rs
```diff
@@ -172,6 +176,12 @@ impl SemConvSpec {
     pub fn groups(&self) -> &[GroupSpec] {
         &self.groups
     }
+
+    /// Returns the list of imports in the semantic convention spec.
+    #[must_use]
+    pub fn imports(&self) -> Option<&[GroupWildcard]> {
```
lmolkova: (random place)

should we warn/fail when someone tries to import a thing that's in the same registry? it doesn't make sense and people would only do this by mistake, we should probably notify them

refactoring_of_group_spec.patch
lquerel: @jerbly What I’m saying is that the glob syntax used to match signal and attribute IDs (or names, depending on the final version) is the same as the one used in the globset crate.

crates/weaver_semconv/src/group.rs
lmolkova: in my suggestion you import by the formal/public name. You never see `metric.db....` unless you look into semconv repo, it's internal impl  detail. The fact we use different properties to represent id inside semconv is sad.

How do you want to solve 'import attribute' problem otherwise?

refactoring_of_group_spec.patch
jerbly: I understand that... this comment above was a file level comment for the patch file you've included. Is that a mistake?

lquerel: Oh yes, my bad. I need to remove this patch from the repo. That was the major refactoring I did to completely change GroupSpec from a struct to an enum. I was planning to create a separate PR with the content of this patch.

lquerel: Good catch, thanks

crates/weaver_semconv/src/semconv.rs
lquerel: I had thought about giving this kind of user's feedback, but actually it’s not so straightforward with the wildcard system. For example, I might have an import like `metric.db.*` that imports all the OTEL database metrics, but at the same time have a local metric declaration that matches this wildcard.
So, it seems clear that we should build something robust, except perhaps for imports without wildcards.

lmolkova: I recall we discussed imports from multiple registries and that we'd need to add registry as the namespace.

E.g. if I have OTel registry and Foo registry, when I import `metric.db....` (or `db....`) I actually implicitly import from `otel::` because it's special to us. 

So when a collision happens, I should be able to resolve it by adding registry `foo::db....`.

I don't like the idea of 

```yaml
- imports
  - metrics:
    db.*
```

importing all db metrics from all registries I have without any means to control it and all the collisions that come with it.

lquerel: We could allow `db.*` only if there is no ambiguity, as with references, and when there is ambiguity, require the custom registry author to specify the registry alias, like `otel:db.*`. What do you think?

lmolkova: imho the best way to prevent ambiguity is to make people explicitly mention where they import from.

It's ok if we have one registry as a default one or if you can set it.
When I write `imports: ['db.foo.bar']` and I have two registries I import from, it should be deterministic where `db.foo.bar` comes from - it comes from the default one. You can only import metric defined in non-default registry using `another_registry:db.foo.bar`.

So if you want to import from all registries, write `imports: ['*:db.foo.bar']`, weaver should not try to be too smart.

i.e. the syntax should minimize the chance of ambiguity, runtime check is the last resort

lquerel:
During the semconv tooling SIG, we decided to represent the `imports` sections with sub-sections.

```
- imports:
  - attributes:
    - http.*       <- name with wildcard (id in the semconv right now)
    - db.*
    - system.*
  - metrics:
    - http.*       <- metric_name with wildcard 
    - db.*
    - system.*
  - entities:
    - host        <- name
  - events:     <- name
    - db.*  
 ```
 
 Right now, the "name" of the persistant identifiers across the various group types is non fully consistent. We have a plan to rename `metric_name` to `name`. 
 See #785 for more details.

**Comment by jerbly** [comment]
Should https://github.com/open-telemetry/weaver/blob/main/crates/weaver_resolver/README.md be updated?

**Comment by lquerel** [comment]
> Are we ready to switch to the auto-generated json schema for use in editors? There is a trade-off that I'm not sure we've agreed on. The manual schema encodes some of the rules which shifts-left the warnings for a better experience for the author 

I thought we had decided to switch, but I agree with your argument. I'll roll back this file.

File: schemas/semconv.schema.json [hunk 6]
```diff
@@ -45,6 +45,24 @@
...
+			"properties": {
+				"metrics": {
+					"type": "array",
+					"items": { "type": "string" }
+				},
+				"events": {
+					"type": "array",
+					"items": { "type": "string" }
+				},
+				"Entities": {
```
lmolkova:
```suggestion
				"entities": {
```

schemas/semconv.schema.json
```diff
@@ -45,6 +45,24 @@
 					}
 				]
 			}
+		},
```
lmolkova, line 48: Could you please also update syntax.md file here and add some docs on what import does and how to use it?

lmolkova (approved):

PR #770, opened 2025-06-06
[forge]: attribute_namespace fallback

sh0rez:
Makes the jinja filter `attribute_namespace` fallback to `other` for unqualified ids (e.g. `phase` becomes `other.phase`)

This is in line with the jq filter `semconv_attributes`, which does the same:
https://github.com/open-telemetry/weaver/blob/e5d7e697ae4b716e5aac335f24c8faf7b824f2ce/defaults/jq/semconv.jq#L80

This makes it easier to work with unqualified attributes, which is common for existing Prometheus metrics.


jsuereth:
Would it make sense to just have a different JQ function called `raw_attributes` vs `semconv_attributes`?

I.e. we can have a set of functions geared towards prometheus use cases that don't add the complexities of namespacing at all.

jsuereth (approved):
This looks good!

Apologies, based on CL description I thought this was modifying JQ filters, but instead this looks to be a good cleanup and addition for JINJA templates.

sh0rez:
thanks @jsuereth! is there anything needed from my side for this to be merged?

PR #773, opened 2025-06-08
Adds metric value type

jerbly:
This PR partially addresses: https://github.com/open-telemetry/semantic-conventions/issues/591 

- Adds `value_type` to `metric` group type.
- Produces a warning if `value_type` is not specified.
- `registry emit` uses the `value_type` to produce correct values
- `registry live-check` compares the incoming value type from data-points and exemplars to produces a `value_type_mismatch` violation.


lmolkova (approved):
thank you!

crates/weaver_emit/src/metrics.rs
```diff
@@ -27,39 +28,75 @@ pub(crate) fn emit_metrics_for_registry(registry: &ResolvedRegistry) {
...
+                        InstrumentSpec::Histogram => {
+                            let histogram = meter
+                                .f64_histogram(metric_name)
+                                .with_unit(unit)
+                                .with_description(description)
+                                .build();
+                            histogram.record(1.0, &attributes);
+                        }
+                    },
+                    _ => match instrument {
```
lquerel: Why not explicitly specify the second branch?

I agree, it’s unlikely, but if we ever introduce a third variant (both in OTLP and Semconv), the compiler won’t help us do the right thing.

crates/weaver_live_check/src/advice.rs
```diff
@@ -305,11 +316,66 @@ impl Advisor for TypeAdvisor {
...
+            }
             _ => Ok(Vec::new()),
         }
     }
 }
 
+fn json_to_value_type(value: &Value) -> Option<MetricValueTypeSpec> {
+    match value {
+        Value::Number(num) => {
+            if num.is_i64() {
```
lquerel: what about `is_u64` ?

jsuereth (commented):
Overall this looks good, if we agree on the direction of enforcing types in weaver.

I think in our data model for OTLP, we don't actually limit metrics to remain the same type for a stream.

For example, from https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/data-model.md, it says:

> Within certain data point types (e.g., `Sum` and `Gauge`) there is variation permitted in the numeric point value; in this case, the associated variation (i.e., floating-point vs. integer) is not considered identifying.

This makes me think we should NOT be enforcing types in weaver generally.

jerbly:
@jsuereth - OK. I won't merge this then until we have a decision. Perhaps we have the `value_type` more as a **recommendation** and the out-of-the-box codegen can use this. Live-check advice-level could be changed to `information` instead of `violation` where there's a mismatch.

PR #774, opened 2025-06-08
Fix: Use requirement level for attributes in live-check

jerbly:
When checking the set of attributes provided in a sample metric against the model, the `requirement_level` was ignored.

This PR fixes this oversight.

If an attribute is not present in the sample this is the mapping:

| RequirementLevel | Live-check advice level |
| --- | --- |
| Required | Violation |
| Recommended | Improvement |
| Opt-In | Information |
| Conditionally Required | Information |

It would be nice to do something better with Conditionally Required but this is not possible until the condition is specified in a machine readable way.

### DIFF
--- a/crates/weaver_live_check/src/advice.rs
+++ b/crates/weaver_live_check/src/advice.rs
@@ -17,7 +17,10 @@ use weaver_checker::{
 use weaver_forge::{jq, registry::ResolvedGroup};
 use weaver_resolved_schema::attribute::Attribute;
 use weaver_semconv::{
-    attribute::{AttributeType, PrimitiveOrArrayTypeSpec, TemplateTypeSpec, ValueSpec},
+    attribute::{
+        AttributeType, BasicRequirementLevelSpec, PrimitiveOrArrayTypeSpec, RequirementLevel,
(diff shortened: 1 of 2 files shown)


crates/weaver_live_check/src/advice.rs
```diff
@@ -157,23 +160,48 @@ impl Advisor for StabilityAdvisor {
...
-/// Checks if required attributes from a resolved group are present in a list of attributes
-/// Returns a list of advice for missing attributes
-fn check_required_attributes(
-    required_attributes: &[Attribute],
-    attributes: &[SampleAttribute],
+/// Checks if attributes from a resolved group are present in a list of sample attributes
+/// Returns a list of advice for the attributes based on their RequirementLevel
+fn check_attributes(
+    semconv_attributes: &[Attribute],
+    sample_attributes: &[SampleAttribute],
```
lquerel: It would be nice to have the same summary table you created in the PR description included in this comment, to map requirement levels with advice levels.

lquerel (approved):
I made a minor suggestion.

LGTM. Thanks!

PR #784, opened 2025-06-11
Use `name` or `metric_name` as namespace for `weaver diff` on signal differences.

### LINKED ISSUE #776: weaver registry diff: old_name is a group id and not a metric/event name
Run 

```bash
weaver.exe registry diff -r .\model --baseline-registry https://github.com/open-telemetry/semantic-conventions/archive/refs/tags/v1.34.0.zip[model] --diff-template .\internal\tools\scripts\schema-diff\ --diff-format yaml
```

on semconv main.

Result

```yaml
...
changes:
  events: []
  metrics:
  - name: metric.k8s.container.status.reason
    type: added
  - new_name: system.cpu.time
    note: Replaced by `system.cpu.time`.
    old_name: metric.cpu.time
    type: renamed
  - new_name: system.cpu.utilization
    note: Replaced by `system.cpu.utilization`.
    old_name: metric.cpu.utilization
    type: renamed
...
```

expected
```yaml
...
changes:
  events: []
  metrics:
  - name: k8s.container.status.reason
    type: added
  - new_name: system.cpu.time
    note: Replaced by `system.cpu.time`.
    old_name: cpu.time
    type: renamed
  - new_name: system.cpu.utilization
    note: Replaced by `system.cpu.utilization`.
    old_name: cpu.utilization
    type: renamed
...
```

jsuereth:
Fixes #776. First step for #785.

This PR:

- Updates diff algorithm to pull groups by *name* instead of *id* (except metric groups, which use `metric_name`.
  - Today in weaver, `name` is guaranteed to be unique via validation done on resolved registries. So is `metric_name`.
  - `extends` is wonky on signals, as they cannot actually use the same name, so that use case isn't supported.
- Also fixes some missing Resource => Entity renames I noticed.

Out of Scope for this PR:

- Fixing the "attribute registry" hackery for `weaver diff`.  I have a simple path forward laid out in #785 that I think can be a quick follow up
- Providing actual "refinement" groups that I think we'll need for multi-registry.

Notes:

- `weaver diff` integration tests actually already used `name` on span groups.

### DIFF
--- a/crates/weaver_resolved_schema/src/lib.rs
+++ b/crates/weaver_resolved_schema/src/lib.rs
@@ -123,7 +123,7 @@ impl ResolvedTelemetrySchema {
             prefix: "".to_owned(),
             extends: None,
             stability: None,
-            deprecated: deprecated,
+            deprecated,
             name: Some(group_id.to_owned()),
             lineage: None,
             display_name: None,
@@ -136,7 +136,7 @@ impl ResolvedTelemetrySchema {
             body: None,
             annotations: None,
             entity_associations: vec![],
-            value_type: None,
+            value_type: Some(weaver_semconv::metric::MetricValueTypeSpec::Double),
         });
     }
 


lquerel (approved):
LGTM.

Issue #785, opened 2025-06-11
Path forward for `id` vs `name` and weaver diff


jsuereth:
I'd like to propose the following:

- Move to a single representation of "name" for signals.  
  - Here `group.name` means "metric name", "span type", "event name" and "entity type".
  - This name forms a namespace *per signal*.  So the entity with name `host` means the same thing everywhere
  - This name is something we'll make sure is transmitted via OTLP and enforce-able in live-check.
- Clarify what "id" means for groups.
   -  `group.id` denotes an identifier for understanding a definition in weaver/semconv YAML files.
   - The id forms a namespace *per group definition*.
   - We expect code and documentation generation to interact *at this level* in weaver.
     - e.g. a `java.apache.http.client.duration` group id would refer to a metric of name `http.client.duration`.  This may be a specialized instance of `http.client.duration` used to codegen for the Java Apache HTTP client library, and can be referenced when generating documentation for that library.

## Background

The difference between `id` and actual identity of a signal has been a sore spot in semantic conventions for some time.  

For example, today all semantic convention compatibility policies are [*enforced at the name level*](https://github.com/open-telemetry/semantic-conventions/blob/main/policies/compatibility.rego#L258).  This means there *is no enforcement for spans today*: an attribute may be dropped with no automated check.  Additionally, we have issues in weaver with resolving registries and enforcing "extension" group ids.

We actually already have this issue with attributes.  Today, *only* registries which match semantic conventions usage of `registry.*` attribute groups can use weaver diff.

- See #776
- See https://github.com/open-telemetry/semantic-conventions/issues/2055
- See https://github.com/open-telemetry/weaver/issues/580

## What's the meaning of `id` vs. `name`

We should view `name` and `id` the same way we see a *type* and an *instance* (or term) in a programming language.  For example `String x`, we would have `String` be the name, and `x` be the id.  The identity denotes a string in a more well understood context and may have more limitations that the general `String`.

Within weaver, today, we do not define "type" and "instance" separately.  Instead, e.g. with attributes we've promoted a special usage in Semantic Conventions where we define "registry" groups, and use `ref` to refine an attribute within a specific context.  Additionally, weaver allows `extend` on group to allow *refinement* of a definition or re-use of a set of attributes.

### Refinement

A key aspect of this proposal is that we understand when a group or attribute is the "root source" of the definition vs. when it is a refinement.  I propose adding refinement tracking in weaver with the following rules:

- An attribute that is a `ref` is a refinement of what it refers.
[...]

PR #787, opened 2025-06-12
Remove need for groups to be named registry

### LINKED ISSUE #580: Add the concept of registry attribute in Weaver.
Today, Weaver recognizes the concept of attributes. To list all registry attributes, it relies on the registry. prefix. The concept of registry attributes should be a first-class citizen in Weaver.

See PR comment: https://github.com/open-telemetry/weaver/pull/400/#discussion_r1905398302

jsuereth:
Fixes #580, part 2 of #785

Remove need for groups to be named registry to figure out attribute registry.

- Use lineage to determine if an attribute is a refinement.
- Update `weaver diff` to pull non-refinement attributes instead of looking for `registry.*` groups.
- Update test to ensure non-registry groups are still treated as registry if they own a raw attribute.

### DIFF
--- a/crates/weaver_forge/README.md
+++ b/crates/weaver_forge/README.md
@@ -523,7 +523,8 @@ The following filters are available (the code for all available extension can be
 - `attribute_registry_title`: Converts registry.{namespace}.{other}.{components} to {Namespace} (title case the
   namespace).
 - `attribute_registry_namespace`: Converts metric.{namespace}.{other}.{components} to {namespace}.
-- `attribute_namespace`: Converts {namespace}.{attribute_id} to {namespace}.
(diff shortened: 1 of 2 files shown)


jsuereth:
cc @lmolkova for awareness

jsuereth:
> I like the direction!
> 
> I assume we'll keep requiring `registry` prefix on semconv side - https://github.com/open-telemetry/semantic-conventions/blob/main/policies/registry.rego (at least until we find a more formal way)?

Yes!  I think semantic conventions can remain the way it is for some time.

This just unblocks "simple" weaver usage so they can still get a valid difference on attribute definitions.

lquerel (approved):
LGTM. 

Glad to see the test on "registry." disappear. I was not sure at first that a registry attribute could be defined solely based on the absence of any refinement.

PR #805, opened 2025-06-22
[WIP] New readme focused on Observability by Design

jerbly:
I have tried to change the tone and style of the readme to focus more on the mission of Weaver. This closely follows the presentation from the last CNCF Europe. That presentation was designed to pitch ObD, and Weaver, the tool to provide it.

I've taken out almost all of the detail which I think will be better served through doc pages for each command. As we write them they can be linked from the Readme.

Opening this PR now to get feedback. I know it needs more changes.
For example, I'm not too happy with the examples. I think this should be a thread that flows through the readme but needs to be very very simple.


README.md
```diff
@@ -1,273 +1,247 @@
...
+    stability: development
+    brief: This span represents a simple message.
+    span_kind: client
+    attributes:
+      - ref: example.message
+        requirement_level: required
+      - ref: host.name
+        requirement_level: required
+      - ref: host.arch
+        requirement_level: required
```
lquerel:
```suggestion
      - ref: host.name                 # imported from OTEL semantic conventions
        requirement_level: required.   # requirement level redefined locally
      - ref: host.arch                 # imported from OTEL semantic conventions
        requirement_level: required.   # requirement level redefined locally
```

README.md
```diff
@@ -1,273 +1,247 @@
...
+```
```
lquerel: After this yaml file, we could talk about the custom registry declaration itself and introduce the `registry_manifest.yaml` file.

README.md
```diff
@@ -1,273 +1,247 @@
...
+id"}')
+- Category        : backward_compatibility
+- Type           : semconv_attribute
+- SemConv group  : metric.auction.bid.count
+- SemConv attribute: auction.id
+- Provenance: /home/weaver/source
+
+Total execution time: 5.531342577s
+make: *** [Makefile:89: check] Error 1
 ```
```
lquerel: A slide that I had in the SRECON presentation but didn’t include in the Kubecon presentation for timing reasons could be interesting to share here.

It's a summary of the policies we already have in place. We could organize the following list as a 2 columns table.

- No attributes outside registry.
- Definitions require stability.
- No requirement levels on attributes
- Names must follow format rules
- IDs must match naming patterns
- Attributes fully qualified
- No constant name collisions
- No namespace collisions
- No duplicate attributes in group
- Experimental attributes in stable groups must be opt-in
- No removal of elements
- No stability downgrades
- No type or unit changes
- Attribute sets immutable
- Enum values immutable

I think it will give readers a more concrete idea of what `weaver check` can do without going too much into detail.

lquerel (commented):
Made few suggestions.
I really like the direction this refactoring of the main readme is taking.
I’m glad that the Kubecon slides were able to help you in this process.

README.md
```diff
@@ -1,273 +1,247 @@
 # OpenTelemetry Weaver
 
+<p align="left">
+  <img src="docs/images/weaver-logo.svg" alt="OpenTelemetry Weaver" width="200" height="100" align="left" style="margin-right: 20px;">
+</p>
+
+### Observability by Design
```
jsuereth, line 7: I like Observability by Design, but we probably still need an elevator pitch for what weaver is before this.

Something like:  "Weaver is an automation tool that defines and verifies observability, allowing code and documentation generation.  Weaver enables Observability by Design."

Then it can transition into what "Observability by Design" is.

jsuereth (approved):

PR #807, opened 2025-06-24
Prepare release v0.15.3

lquerel:


jsuereth (commented):
Should there be a patch file committed here?

PR #812, opened 2025-07-01
Annotations for enum members

lmolkova:
We need to be able to exclude specific enum members from code generation to prevent collisions, see https://github.com/open-telemetry/semantic-conventions/issues/2455


crates/weaver_semconv/src/attribute.rs
```diff
@@ -385,6 +385,10 @@ pub struct EnumEntriesSpec {
     pub stability: Option<Stability>,
     /// Deprecation note.
     pub deprecated: Option<String>,
+    /// Annotations for the member.
+    #[serde(default)]
+    #[serde(skip_serializing_if = "Option::is_none")]
```
jerbly: These serde attributes are not present on the other use of `annotations` - should we be consistent?

crates/weaver_emit/src/attributes.rs
```diff
@@ -577,6 +584,7 @@ mod tests {
                         note: None,
                         stability: None,
                         deprecated: None,
+                        annotations: None,
```
jsuereth, line 587: @jerbly / @lquerel After seeing this type of change bloat PRs a thousand times,

How do we feel about using the DEFAULT syntax so this isn't so tedious when safely adding new fields?

e.g.:

```rust
EnumEntriesSpec {
  id: "non-default".to_owned(),
  ...Default::default(),
}
```

Instead of these more verbose updates where it's harder to see the content of what has changed...

defaults/jq/semconv.jq
```diff
@@ -53,6 +53,10 @@ def code_generation_exclude_filter($options):
             or .annotations.code_generation.exclude == null
             or .annotations.code_generation.exclude == false
         ))
+        | map(.type.members? |= map(select(.annotations == null
```
jsuereth, line 56: I'm amazed I can read this.  Also would this work?

```jq
| map(.type.members? |= map(select(.annotations?.code_generation?.exclude? == true | not)
```

jsuereth (approved):
Overall, looks great.  Just nits.

defaults/jq/semconv.jq
lmolkova: it should have, but jaq doesn't support null coalescence (aka optional operator).

don't ask me how much time I spent figuring this one out in the past 😅 

https://github.com/open-telemetry/weaver/blob/3090602ed1984003b64afdc2708392608fa7bd5f/defaults/jq/semconv.jq#L49

crates/weaver_semconv/src/attribute.rs
lmolkova: I wish we didn't serialize null values everywhere, but that's way beyond the scope of this PR.

Updated in https://github.com/open-telemetry/weaver/pull/812/commits/026a28ff0d4d2011cbebf11f49060f3824689f98

crates/weaver_emit/src/attributes.rs
lquerel: I agree

lquerel (approved):
LGTM. Thanks

lquerel:
@lmolkova Ideally, an entry should have been added to the CHANGELOG.md to describe the feature in this PR. But I forgot to check that before the merge of this PR. I’ll add the entry in the PR I’ll create for the release.

PR #816, opened 2025-07-02
Reverts value_type

jerbly:
Following some discussion on `value_type` in the context of https://github.com/open-telemetry/semantic-conventions/pull/2444 , the conclusion is that it's more of a code-generation hint than an actual requirement for over-the-wire data. According to spec changing value type is not breaking.


lquerel (approved):
LGTM

lmolkova (approved):

PR #822, opened 2025-07-03
Don't serialize null properties and empty arrays

lmolkova:


jsuereth (approved):
Can you add this to the changelog under "next"?

Issue #824, opened 2025-07-03
[WIP] Schema v2: proposal


lmolkova:
Part of  https://github.com/open-telemetry/opentelemetry-specification/issues/4427

## Harmonize identity property names

Principles: 
- same as OTLP property when possible/makes sense
- default to `name`

| signal/object | current id semconv | proposal | in proto | 
| --- | --- | --- | --- |
| attributes | `id` (unresolved schema) and `name` (resolved) | key | key |
| metrics | metric_name | name | name | 
| events | name | name | `event_name`, no name for log records |
| spans | - |  type |  [will propose](https://github.com/open-telemetry/semantic-conventions/issues/2055) `type`, spans already have `name`, `kind` and `spanId` |
| entity | name | type | type |
| profiles | - | - | - | 
| scopes | - | - | name |

## Have dedicated types for different objects/signals

```yaml
attributes: # instead of `registry.*` attribute groups
- key: foo.name
  ...
- key: foo.target
metrics:  # instead of group.type: metric
- name: foo.duration
entities:
- type: foo
events:
- name: foo.ended
spans:
- type: foo.client.request # new
  kind: client
  name: {foo.name} {foo.target}  # new
```

### References have different types as well

(referencing is only available in unresolved schema)

```yaml
metric_refs:
- id: my_flavor.foo.duration
  extends: foo.duration # or fully-qualified otel.foo.duration, namespace delimiter tbd
  attributes:
  - ref: azure.resource.id 
    brief: ...
```

each `_ref` instance has to be identifiable. There could be multiple favors of the same metric that have the same name, but different ids. 

### Imports 

no changes, works naturally and looks consistent

```yaml
imports:
- attributes: foo.*
- spans: foo.*
```

## Stages

1. weaver supports new unresolved schema format 
    - unresolved schema is still pretty much contained within semconv, easy to change
    - weaver should probably support old schema for the time being and convert it to the same internal representation
--- we're here on 8/18/25 -- 
3. new resolution format is produced behind feature flag    - resolved schema is extensively used by codegen
    - need to socialize among maintainers
    - jq helpers used by codegen produce something close to the proposal (group by attributes/metrics/etc), so we can probably keep back-compat
3. default to the new format

jsuereth:
@lmolkova (and others) When you get a chance can you look at: https://github.com/jsuereth/weaver/commit/192d539ea1c2fdd91206618782f56014eebf9bdc

[...]

Issue #820, opened 2025-07-02
Define error handling in yaml


thompson-tomo:
**Is your feature request related to a problem? Please describe.**
I want to be able to have my attributes clearly defining what values should be indicating that an error has occured

**Describe the solution you'd like**
On attributes we should be able to define an error status object which can be used for code generation. To facilitate this the object needs a comparison type ie =, <. As well as the comparator ie 0.

**Describe alternatives you've considered**
Using annotations

**Additional context**
N/a

thompson-tomo:
@jsuereth an example would be for https://opentelemetry.io/docs/specs/semconv/rpc/grpc/#client it would have conditions of:

- error.* is present
- rpc.response.status_code != OK

Whereas https://opentelemetry.io/docs/specs/semconv/rpc/grpc/#server would have:

- error.* is present
- rpc.response.status_code == UNKNOWN
- rpc.response.status_code == DEADLINE_EXCEEDED
- rpc.response.status_code == UNIMPLEMENTED
- rpc.response.status_code == INTERNAL
- rpc.response.status_code == UNAVAILABLE
- rpc.response.status_code == DATA_LOSS

That way tools like live-check can check if the status matches the attributes present.

Issue #819, opened 2025-07-02
Define span names in yaml


thompson-tomo:
**Is your feature request related to a problem? Please describe.**
For semconv we are needing to define the span name in the note property as we don't have another field for it.

**Describe the solution you'd like**
The ability to define the following for spans & scopes
- naming.templates - the templates to be use to construct the name in the order they are tried
- naming.default - the name to be used temple can not be used
- ~~Naming.expectation - should vs must~~
- naming.brief
- naming.note

**Describe alternatives you've considered**
Leave it as it is

**Additional context**
N/a

Issue #796, opened 2025-06-18
Metrics semconv should define requirement levels in yaml


lmolkova:
Extracted from https://github.com/open-telemetry/semantic-conventions/issues/591

thompson-tomo:
@lmolkova is this a duplicate of #406?

joaopgrassi:
I'll close it as it is a duplicate of https://github.com/open-telemetry/weaver/issues/406

PR #829, opened 2025-07-04
Initial cut at version two specification

jsuereth:
- Create specific structures for each signal type in semantic conventions
- Add conversion for these back to "Group" Semantics
- Simple tests for backwards conversion, and overall registry resolution.
- Create overall "SemConvSpec" which can resolve unversioned or versioned formats.

A few notes:

- Signals MUST reference attributes, and cannot define them fresh
- Specify attributes creates an arbitrary `attribute_group`, named `registry.{cleaned to file}`.  It has a synthetic brief
  and may "break" or make awkward the semantic conventions registry generation.  This needs some discussion.
- Schemars has some issues with our lint rules: https://github.com/GREsau/schemars/issues/467

See https://github.com/open-telemetry/weaver/issues/824


lquerel:
From what I understand, the approach taken in this PR is not to start from the resolved version of the schema, but to start from the semantic conventions part. Is that correct? I'm saying this because I see that we have AttributeRef with overloading and the changes are in the `weaver_semconv` crate.

I thought we had agreed to start from the resolved part in order to get closer to schema v2 more quickly. This approach also works, but I feel like it will just take a bit more time.

crates/weaver_semconv/src/v2/span.rs
```diff
@@ -0,0 +1,111 @@
...
+};
+
+/// A group defines an attribute group, an entity, or a signal.
+/// Supported group types are: `attribute_group`, `span`, `event`, `metric`, `entity`, `scope`.
+/// Mandatory fields are: `id` and `brief`.
+///
+/// Note: The `resource` type is no longer used and is an alias for `entity`.
+#[derive(Serialize, Deserialize, Debug, Clone, JsonSchema)]
+#[serde(deny_unknown_fields)]
+pub struct SpanGroup {
```
thompson-tomo: Any chance we could add #819

thompson-tomo: Any chance we could add #820

crates/weaver_semconv/src/v2/metric.rs
```diff
@@ -0,0 +1,106 @@
...
+
+use crate::{
+    group::{GroupSpec, InstrumentSpec},
+    v2::{attribute::AttributeRef, CommonFields},
+};
+
+/// A MetricGroup defines a new metric.
+#[derive(Serialize, Deserialize, Debug, Clone, JsonSchema)]
+#[serde(deny_unknown_fields)]
+pub struct MetricGroup {
```
thompson-tomo: Could we address #796/#406

crates/weaver_semconv/src/v2/attribute.rs
```diff
@@ -0,0 +1,170 @@
...
+    stability::Stability,
+    v2::CommonFields,
+    YamlValue,
+};
+
+/// A refinement of an Attribute for a signal.
+#[derive(Serialize, Deserialize, Debug, Clone, JsonSchema)]
+#[serde(deny_unknown_fields)]
+#[serde(rename_all = "snake_case")]
+pub struct AttributeRef {
```
thompson-tomo, line 22: Could #740 be added?

crates/weaver_semconv/src/v2/span.rs
jsuereth: This does have the name now.

jsuereth: Yes - this is just a prototype of what we could do, based on @lmolkova's proposal

crates/weaver_semconv/src/v2/metric.rs
jsuereth: This would need to add support in the old groups section too.  Some of this I'm keeping out of scope for the initial PR, but we can address these going forward, It's just lower on the totem pole of "get it working"

crates/weaver_semconv/src/v2/attribute.rs
jsuereth: Similar response as before.

jsuereth:
> Should we take this opportunity to split custom types (enum) out and simply mention the type on the attribute?

I like this, but suspect we need some more thought here, and need to tie in the other discussion around "custom types" we had about a year ago. @lmolkova what do you think?

thompson-tomo:
No worries @jsuereth The namespace discussion is separate to this one as there is very little if any overlap. I have raised https://github.com/open-telemetry/weaver/issues/857 for introducing bundles and ensuring that we are not adding additional maintance effort etc.

crates/weaver_semconv/src/v2/attribute.rs
```diff
@@ -0,0 +1,170 @@
...
+    /// attributes. Example values must be of the same type of the
+    /// attribute. If only a single example is provided, it can directly
+    /// be reported without encapsulating it into a sequence/dictionary.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    examples: Option<Examples>,
+    /// Associates a tag ("sub-group") to the attribute. It carries no
+    /// particular semantic meaning but can be used e.g. for filtering
+    /// in the markdown generator.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    tag: Option<String>,
```
lmolkova: do we need it in v2 right away? we don't tags in semconv anymore, but we can always add them later if necessary

crates/weaver_semconv/src/v2/attribute.rs
```diff
@@ -0,0 +1,170 @@
...
+    /// present and stability differs from deprecated, this will result in an
+    /// error.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    stability: Option<Stability>,
+    /// Specifies if the attribute is deprecated.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[serde(
+        deserialize_with = "crate::deprecated::deserialize_option_deprecated",
+        default
+    )]
```
lmolkova: do we need this back-compat-with-string case in v2?

crates/weaver_semconv/src/v2/attribute.rs
```diff
@@ -0,0 +1,170 @@
...
+    deprecated: Option<Deprecated>,
+    /// Specifies the prefix of the attribute.
+    /// If this parameter is set, the resolved id of the referenced attribute will
+    /// have group prefix added to it.
+    /// It defaults to false.
+    #[serde(default)]
+    #[serde(skip_serializing_if = "<&bool>::not")]
+    prefix: bool,
```
lmolkova: do we need it at all? we don't do prefixes anymore

crates/weaver_semconv/src/v2/entity.rs
```diff
@@ -0,0 +1,104 @@
...
+    #[serde(default)]
+    #[serde(skip_serializing_if = "Vec::is_empty")]
+    pub description: Vec<AttributeRef>,
+    /// Common fields (like brief, note, attributes).
+    #[serde(flatten)]
+    pub common: CommonFields,
+}
+
+impl EntityGroup {
+    /// Converts a v2 event into a v1 GroupSpec.
```
lmolkova:
```suggestion
    /// Converts a v2 entity into a v1 GroupSpec.
```

crates/weaver_semconv/src/v2/event.rs
```diff
@@ -0,0 +1,103 @@
...
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+
+use crate::{
+    any_value::AnyValueSpec,
+    group::GroupSpec,
+    v2::{attribute::AttributeRef, CommonFields},
+};
+
+/// A MetricGroup defines a new metric.
```
lmolkova:
```suggestion
/// An EventGroup defines a new event.
```

?

crates/weaver_semconv/src/v2/span.rs
```diff
@@ -0,0 +1,111 @@
...
+    /// Which resources this span should be associated with.
+    #[serde(default)]
+    #[serde(skip_serializing_if = "Vec::is_empty")]
+    pub entity_associations: Vec<String>,
+    /// List of strings that specify the ids of event semantic conventions
+    /// associated with this span semantic convention.
+    /// Note: only valid if type is span
+    #[serde(default)]
+    #[serde(skip_serializing_if = "Vec::is_empty")]
+    pub events: Vec<String>,
```
lmolkova: can we nuke it in v2 now that span events are being deprecated?

lmolkova (commented):
Looks great! thanks a lot for putting it together! 

Would it be possible to document schema v2 format? If you want help, I can do it as a separate PR - let's not block this one.

crates/weaver_semconv/src/v2/attribute.rs
```diff
@@ -0,0 +1,170 @@
...
+    /// the attribute is "recommended". When set to
+    /// "conditionally_required", the string provided as `condition` MUST
+    /// specify the conditions under which the attribute is required.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    requirement_level: Option<RequirementLevel>,
+    /// Specifies if the attribute is (especially) relevant for sampling
+    /// and thus should be set at span start. It defaults to false.
+    /// Note: this field is experimental.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    sampling_relevant: Option<bool>,
```
jsuereth: @jsuereth TODO - move this into span-only reference.

crates/weaver_semconv/src/v2/attribute.rs
```diff
@@ -0,0 +1,170 @@
...
+    /// attribute. If only a single example is provided, it can directly
+    /// be reported without encapsulating it into a sequence/dictionary.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub examples: Option<Examples>,
+    /// Associates a tag ("sub-group") to the attribute. It carries no
+    /// particular semantic meaning but can be used e.g. for filtering
+    /// in the markdown generator.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub tag: Option<String>,
+    /// Common fields (like brief, note, attributes).
```
jsuereth, line 126: s/attributes/annotations/

crates/weaver_semconv/src/v2/mod.rs
```diff
@@ -0,0 +1,223 @@
...
+    #[serde(default, skip_serializing_if = "String::is_empty")]
+    pub note: String,
+    /// Specifies the stability of the span.
+    pub stability: Stability,
+    /// Specifies if the semantic convention is deprecated. The string
+    /// provided as description MUST specify why it's deprecated and/or what
+    /// to use instead. See also stability.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[serde(
+        deserialize_with = "crate::deprecated::deserialize_option_deprecated",
```
jsuereth: Remove string-based deprecation options from v1 here.

crates/weaver_semconv/src/v2/mod.rs
```diff
@@ -0,0 +1,223 @@
...
+        default
+    )]
+    pub deprecated: Option<Deprecated>,
+    /// Annotations for the group.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub annotations: Option<BTreeMap<String, YamlValue>>,
```
jsuereth: Move to empty BTree here.

crates/weaver_semconv/src/v2/metric.rs
```diff
@@ -0,0 +1,106 @@
...
+    /// [guidelines](https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/metrics/semantic_conventions#instrument-units).
+    pub unit: String,
+    /// List of attributes that belong to the semantic convention.
+    #[serde(default)]
+    #[serde(skip_serializing_if = "Vec::is_empty")]
+    pub attributes: Vec<AttributeRef>,
+    /// Which resources this metric should be associated with.
+    #[serde(default)]
+    #[serde(skip_serializing_if = "Vec::is_empty")]
+    pub entity_associations: Vec<String>,
```
jsuereth, line 36: TODO: Think about EntityRef

crates/weaver_semconv/src/v2/span.rs
```diff
@@ -0,0 +1,111 @@
...
+#[serde(deny_unknown_fields)]
+pub struct SpanGroup {
+    /// The type of the Span. This denotes the identity
+    /// of the "shape" of this span, and must be unique.
+    pub r#type: String,
+    /// Specifies the kind of the span.
+    /// Note: only valid if type is span
+    pub kind: SpanKindSpec,
+    /// The name pattern for the span.
+    pub name: String,
```
jsuereth: TODO - Consider making this more robust/explicit in how it works.

crates/weaver_semconv/src/v2/span.rs
```diff
@@ -0,0 +1,111 @@
...
+/// Supported group types are: `attribute_group`, `span`, `event`, `metric`, `entity`, `scope`.
+/// Mandatory fields are: `id` and `brief`.
+///
+/// Note: The `resource` type is no longer used and is an alias for `entity`.
+#[derive(Serialize, Deserialize, Debug, Clone, JsonSchema)]
+#[serde(deny_unknown_fields)]
+pub struct SpanGroup {
+    /// The type of the Span. This denotes the identity
+    /// of the "shape" of this span, and must be unique.
+    pub r#type: String,
```
jsuereth: Consider creating a rust type (called `Identity`?) which enforces restrictions on allowable strings.

crates/weaver_semconv/src/v2/event.rs
```diff
@@ -0,0 +1,103 @@
...
+#[derive(Serialize, Deserialize, Debug, Clone, JsonSchema)]
+#[serde(deny_unknown_fields)]
+pub struct EventGroup {
+    /// The name of the event.
+    pub name: String,
+    /// The event body definition
+    pub body: AnyValueSpec,
```
jsuereth: TODO - Remove or change this to be json-schema, like an embeeded one or something.

crates/weaver_semconv/src/v2/span.rs
lmolkova: maybe `SignalIdentity`

crates/weaver_semconv/src/v2/span.rs
thompson-tomo: Wouldn't that be #819? Perhaps start with an object so we can add to the object later.

crates/weaver_semconv/src/semconv.rs
```diff
@@ -13,11 +14,82 @@ use std::path::Path;
...
+};
+
+/// A semantic convention file as defined [here](https://github.com/open-telemetry/build-tools/blob/main/semantic-conventions/syntax.md)
+/// A semconv file is a collection of semantic convention groups (i.e. [`GroupSpec`]).
+#[derive(Serialize, Deserialize, Debug, Clone, JsonSchema, PartialEq)]
 #[serde(deny_unknown_fields)]
-pub struct SemConvSpec {
+pub struct SemConvSpecV1 {
     /// A collection of semantic convention groups or [`GroupSpec`].
+    #[serde(default)] // TODO - Should we have disjoint V1 and V2 instead?
```
lquerel: I don’t understand this todo, since we are in the context of SemConvSpecV1.

crates/weaver_semconv/src/semconv.rs
```diff
@@ -51,7 +123,16 @@ pub struct SemConvSpecWithProvenance {
     pub provenance: Provenance,
 }
 
-impl SemConvSpec {
+/// A wrapper for a [`SemConvSpec`] with its provenance.
+#[derive(Debug, Clone)]
+pub struct SemConvSpecV1WithProvenance {
```
lquerel, line 128: Couldn’t we make `SemConvSpecWithProvenance` generic to accept both v1 and v2 semconv specs? I haven't looked in detail to see if it’s possible. It's more of a comment for consideration.

crates/weaver_semconv/src/v2/attribute.rs
```diff
@@ -0,0 +1,170 @@
...
+    /// attributes. Example values must be of the same type of the
+    /// attribute. If only a single example is provided, it can directly
+    /// be reported without encapsulating it into a sequence/dictionary.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub examples: Option<Examples>,
+    /// Associates a tag ("sub-group") to the attribute. It carries no
+    /// particular semantic meaning but can be used e.g. for filtering
+    /// in the markdown generator.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub tag: Option<String>,
```
lquerel: Based on the previous comment in `AttributeRef`, we also should remove the tag field into the `AttributeDef` struct.

crates/weaver_semconv/src/v2/entity.rs
```diff
@@ -0,0 +1,104 @@
...
+
+use crate::{
+    group::GroupSpec,
+    v2::{attribute::AttributeRef, CommonFields},
+};
+
+/// An EntityGroup defines a new entity.
+#[derive(Serialize, Deserialize, Debug, Clone, JsonSchema)]
+#[serde(deny_unknown_fields)]
+pub struct EntityGroup {
```
lquerel: Do we need to keep this Group suffix (EntityGroup, EventGroup, MetricGroup)?
It would seem clearer to me without this suffix.

crates/weaver_semconv/src/v2/entity.rs
```diff
@@ -0,0 +1,104 @@
...
+    /// The type of the Entity.
+    pub r#type: String,
+    /// The attributes that make the identity of the Entity.
+    pub identity: Vec<AttributeRef>,
+    /// The attributes that make the description of the Entity.
+    #[serde(default)]
+    #[serde(skip_serializing_if = "Vec::is_empty")]
+    pub description: Vec<AttributeRef>,
```
lquerel, line 24: Just a suggestion to clarify the semantic of these fields:
- `identity` -> `identifying_attributes` or `id_attributes`
- `description` -> `descriptive_attributes` or `info_attributes`

crates/weaver_semconv/src/v2/metric.rs
```diff
@@ -0,0 +1,106 @@
...
+    /// The instrument type that should be used to record the metric. Note that
+    /// the semantic conventions must be written using the names of the
+    /// synchronous instrument types (counter, gauge, updowncounter and
+    /// histogram).
+    /// For more details: [Metrics semantic conventions - Instrument types](https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/metrics/semantic_conventions#instrument-types).
+    /// Note: This field is required if type is metric.
+    pub instrument: InstrumentSpec,
+    /// The unit in which the metric is measured, which should adhere to the
+    /// [guidelines](https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/metrics/semantic_conventions#instrument-units).
+    pub unit: String,
```
lquerel, line 28: [comment] If one day, someone has time to make a unit parser, that would be cool!

lquerel (approved):
I made a few minor suggestions. I love the direction we’re taking with schema v2.

crates/weaver_semconv/src/v2/span.rs
lmolkova: left a suggestion in https://github.com/open-telemetry/weaver/issues/819#issuecomment-3137443676

For the scope of this PR, I think we should stick to a free-form note

```yaml
spans:
- type: http.client
  name: # name_template, name_format, name_pattern, etc
    note: >
      This is a free-form text on how humans/AI should write/review instrumentation code. 
      Code-generator for `start_span` method would require span_name to be provided by the caller.
```

as a semconv author, I'd still prefer `name:  "This is a free-form text..."` for the current usage, but it probably has negative effect on error messages and complexity within weaver/codegen once we start evolving it

crates/weaver_semconv/src/semconv.rs
jsuereth: AH, I had fixed this todo but forgot to remove it...

crates/weaver_semconv/src/semconv.rs
jsuereth: I tried this, yes.

I can do a bit more work around this, but effectively we FIRST get SemConvSpecWithProvenance then we erase to SemconvSpecV1WithProvenance.

We need to do this conversion somewhere, and the way error messages are created, I had to do it earlier vs. later.

crates/weaver_semconv/src/v2/attribute.rs
jsuereth: Removed

crates/weaver_semconv/src/v2/attribute.rs
jsuereth: Doh, removed

crates/weaver_semconv/src/v2/attribute.rs
jsuereth: Removed

crates/weaver_semconv/src/v2/span.rs
jsuereth: Removed!

crates/weaver_semconv/src/v2/attribute.rs
jsuereth: removed

crates/weaver_semconv/src/v2/entity.rs
jsuereth: Removed!

crates/weaver_semconv/src/v2/attribute.rs
jsuereth: Fixed

crates/weaver_semconv/src/v2/span.rs
jsuereth: I called it `SignalId` (can switch to Identity if desired).  Does not have any string restrictions yet, but we have a location to add them.

crates/weaver_semconv/src/v2/event.rs
jsuereth: removed

crates/weaver_semconv/src/v2/attribute.rs
```diff
@@ -0,0 +1,148 @@
...
+    pub brief: Option<String>,
+    /// Sequence of example values for the attribute or single example
+    /// value. They are required only for string and string array
+    /// attributes. Example values must be of the same type of the
+    /// attribute. If only a single example is provided, it can directly
+    /// be reported without encapsulating it into a sequence/dictionary.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub examples: Option<Examples>,
+    /// Specifies if the attribute is mandatory. Can be "required",
+    /// "conditionally_required", "recommended" or "opt_in". When omitted,
```
jerbly, line 37: > When omitted, the attribute is "recommended".

Is this true? I think it inherits the requirement level.

crates/weaver_semconv/src/v2/attribute.rs
```diff
@@ -0,0 +1,148 @@
...
+    /// the attribute is "recommended". When set to
+    /// "conditionally_required", the string provided as `condition` MUST
+    /// specify the conditions under which the attribute is required.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub requirement_level: Option<RequirementLevel>,
+    /// A more elaborate description of the attribute.
+    /// It defaults to an empty string.
+    #[serde(default)]
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub note: Option<String>,
```
jerbly: default and Option?

crates/weaver_semconv/src/v2/attribute.rs
```diff
@@ -0,0 +1,148 @@
...
+    /// Note that, if stability is missing but deprecated is present, it will
+    /// automatically set the stability to deprecated. If deprecated is
+    /// present and stability differs from deprecated, this will result in an
+    /// error.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub stability: Option<Stability>,
+    /// Specifies if the attribute is deprecated.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[serde(default)]
+    pub deprecated: Option<Deprecated>,
```
jerbly: default and Option?

crates/weaver_semconv/src/v2/attribute.rs
```diff
@@ -0,0 +1,148 @@
...
+    /// Annotations for the attribute.
+    pub annotations: Option<BTreeMap<String, YamlValue>>,
```
jerbly:
```suggestion
    #[serde(skip_serializing_if = "Option::is_none")]
    pub annotations: Option<BTreeMap<String, YamlValue>>,
```

crates/weaver_semconv/src/v2/attribute.rs
```diff
@@ -0,0 +1,148 @@
...
+    }
+}
+
+/// The definition of an Attribute.
+#[derive(Serialize, Deserialize, Debug, Clone, JsonSchema)]
+#[serde(deny_unknown_fields)]
+#[serde(rename_all = "snake_case")]
+pub struct AttributeDef {
+    /// String that uniquely identifies the attribute.
+    pub key: String,
```
jerbly, line 114: Maybe I missed a discussion on this. But why `key` and not `name`?

crates/weaver_semconv/src/v2/metric.rs
jerbly: I was looking for a good UCUM lib for live-check so samples not found in the registry could have an `improvement` advisory if the unit was not UCUM. But, I couldn't find one.

crates/weaver_semconv/src/v2/mod.rs
```diff
@@ -0,0 +1,209 @@
...
+    #[serde(default, skip_serializing_if = "String::is_empty")]
+    pub note: String,
+    /// Specifies the stability of the span.
+    pub stability: Stability,
+    /// Specifies if the semantic convention is deprecated. The string
+    /// provided as description MUST specify why it's deprecated and/or what
+    /// to use instead. See also stability.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[serde(default)]
+    pub deprecated: Option<Deprecated>,
```
jerbly: default and Option?

crates/weaver_semconv/src/v2/mod.rs
```diff
@@ -0,0 +1,209 @@
...
+    /// Annotations for the group.
+    #[serde(default, skip_serializing_if = "BTreeMap::is_empty")]
+    pub annotations: BTreeMap<String, YamlValue>,
```
jerbly, line 45: Elsewhere annotations are defined as Option. We should be consistent.

crates/weaver_semconv/src/v2/span.rs
jerbly: Is there an opportunity to be consistent across attributes, spans, metrics, ... We've used `key` on attribute - should `key` be this identity thing?

crates/weaver_semconv/src/v2/mod.rs
jsuereth: They're define as Option because they can be an override.  For the purpose of provenance, should we just assume an empty tree is the same as not defining anything?  Should override always be *additive* as opposed to replacing the previous value?

crates/weaver_semconv/src/v2/attribute.rs
jsuereth: Attributes are key-value pairs.  The OTLP proto uses `key`: https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/common/v1/common.proto#L65

crates/weaver_semconv/src/v2/mod.rs
jsuereth: Updated

crates/weaver_semconv/src/v2/mod.rs
jsuereth: Can you clarify what you mean by this comment?

crates/weaver_semconv/src/v2/attribute.rs
jsuereth: Fixed

crates/weaver_semconv/src/v2/attribute.rs
jsuereth: Can you clarify? I added skip serializing if...

crates/weaver_semconv/src/v2/attribute.rs
jsuereth: Updated all the docs around refinement to denote they are refinements.

crates/weaver_semconv/src/v2/attribute.rs
```diff
@@ -23,42 +23,45 @@ pub struct AttributeRef {
     /// Reference an existing attribute by key.
     pub r#ref: String,
 
-    // TODO - Simplify the options below for "override" / "refine" focus.
-    /// A brief description of the attribute.
+    /// Refines the brief description of the attribute.
+    #[serde(default)]
```
jerbly: #[serde(default)] is not needed for `Option` fields. They deserialize to `None` by default anyway. There are a lot of occurrences of this, I won't comment on them all.

crates/weaver_semconv/src/v2/mod.rs
jerbly: `#[serde(default)]` is not needed for Option fields. They deserialize to None by default anyway.

crates/weaver_semconv/src/v2/attribute.rs
jerbly: OK. So we're aligning with OTLP, great!

crates/weaver_semconv/src/v2/attribute.rs
jerbly: #[serde(default)] is not needed for Option fields. They deserialize to None by default anyway.

crates/weaver_semconv/src/v2/attribute.rs
jerbly: #[serde(default)] is not needed for Option fields. They deserialize to None by default anyway.

crates/weaver_semconv/src/v2/span.rs
jsuereth: I think the opportunity is to align the ids with what's in OTLP / rest of the opentelemetry specification.  At least that's the approach we took.  Trying to have something be "key" in semconv, but "name", "type" or "key" in OTLP causes confusion.  Semconv shouldn't be an island.

crates/weaver_semconv/src/v2/attribute.rs
jsuereth: Fixed

crates/weaver_semconv/src/v2/attribute.rs
jsuereth: Fixed

crates/weaver_semconv/src/v2/span.rs
jsuereth: Updated to @lmolkova's suggestion.

crates/weaver_semconv/src/v2/event.rs
```diff
@@ -0,0 +1,96 @@
...
+    #[serde(skip_serializing_if = "Vec::is_empty")]
+    pub attributes: Vec<AttributeRef>,
+    /// Which resources this event should be associated with.
+    #[serde(default)]
+    #[serde(skip_serializing_if = "Vec::is_empty")]
+    pub entity_associations: Vec<String>,
+    /// Common fields (like brief, note, annotations).
+    #[serde(flatten)]
+    pub common: CommonFields,
+}
```
thompson-tomo, line 30: @jsuereth it Looks like we are missing body?

lmolkova (approved):

crates/weaver_semconv/src/v2/event.rs
jsuereth: It was purposefully removed, cc @lmolkova with details.

crates/weaver_semconv/src/v2/entity.rs
```diff
@@ -7,15 +7,15 @@ use serde::{Deserialize, Serialize};
...
+    v2::{attribute::AttributeRef, signal_id::SignalId, CommonFields},
 };
 
 /// Defines a new entity.
 #[derive(Serialize, Deserialize, Debug, Clone, JsonSchema)]
 #[serde(deny_unknown_fields)]
 pub struct Entity {
     /// The type of the Entity.
-    pub r#type: String,
+    pub r#type: SignalId,
```
jerbly, line 18: Is an entity a signal?

jerbly (approved):
This looks great.

crates/weaver_semconv/src/v2/event.rs
thompson-tomo: That is a shame as I don't feel that we should be removing the ability to define how that property of the protocol is used especially as it is currently used in semconv. Yes have complex attributes on events would decrease the need for them but not eliminate it. Scenarios are for things like raw request & response's ie values which are likely to only relevant to this single event as a summary is available via other properties ie http.response.status_code

crates/weaver_semconv/src/v2/entity.rs
jsuereth: That's the (eventual) plan, yes. It's like a meta-signal now.

crates/weaver_semconv/src/v2/event.rs
jsuereth: We also don't define Summary metrics that OTLP supports.  Semconv has always been a subset of OTLP - the subset we think is the predominate way moving forward for generating and sharing data.

I'll let @lmolkova respond more directly to event plans.

PR #851, opened 2025-07-18
docs: Usage section

tombrk:
I wrote the usage section I hoped would have been there when I started using Weaver.

I focused on describing the core concepts and functionality involved when using Weaver to generate code, which imo is the most likely thing someone will attempt first in their schematized telemetry journey.

The new Usage section of the README consolidates the previous documentation section with added docs in such a way that introduced concepts build onto each other.

I did not aim for completeness. Existing reference documentation is just fine, but is now hopefully more easily approachable for newcomers


README.md
```diff
@@ -79,15 +97,6 @@ cargo build --release
 | [weaver registry emit](docs/usage.md#registry-emit)                       | Emits a semantic convention registry as example signals to your OTLP receiver |
 | [weaver completion](docs/usage.md#completion)                             | Generate shell completions |
 
-## Documentation
```
jsuereth, line 82: I realize you didn't aim for completeness, but until we can ensure we have links to all these things, I think we should at least keep this section somewhere below your new sections.

Generally like the improvements and flow, but would like to keep these links until we determine they're no longer needed.

docs/validate.md
```diff
@@ -0,0 +1,16 @@
...
+- Parsing the loaded semantic convention specifications.
+- Resolving references and extends clauses within the specifications.
+- Checking compliance with specified Rego policies, if provided.
+
+
+## Custom rules (Rego)
+
+Specific validation rules can be expressed using the Rego policy language (https://www.openpolicyagent.org/docs/policy-language).
+
+Please see [Weaver Forge](https://github.com/open-telemetry/weaver/blob/main/crates/weaver_forge/README.md) for details.
```
jsuereth: Is this the right link?

jsuereth (approved):
Really love this PR!

Definitely love the documentation around how template generation works.   I would like to preserve some of the existing references, but I think this is overall another big improvement like the previous one from @jerbly.

README.md
lquerel: @jsuereth @tombrk 

I made an inventory of the links removed from the main README and the relocation of these links into the new documents introduced in this PR. Here’s the result:
- Links moved to `docs/codegen.md`:
  - `docs/architecture.md`
  - `docs/weaver-config.md`
  - `crates/weaver_forge/README.md`
- Links moved elsewhere in the main README:
  - `crates/weaver_live_check/README.md`
  - `docs/schema-changes.md`
  - application telemetry schema OTEP

The only link that I think has disappeared is `crates/weaver_checker/README.md`. It should be in `docs/validate.md` instead of the link to `crates/weaver_forge/README.md`, which was probably introduced by mistake.

I will add a comment in the `docs/validate.md` file to more explicitly point out this minor issue.

docs/codegen.md
```diff
@@ -0,0 +1,110 @@
...
+>[!Note]
+>Weaver enables this using code-generation, powered by [minijinja](https://github.com/mitsuhiko/minijinja) and [jaq](https://github.com/01mf02/jaq).  
+>Full documentation is available at [Weaver Forge](/crates/weaver_forge/README.md)
+
+## High-level overview
+
+>[!Tip]
+>For a full tutorial, consider the [Step-by-Step Guide](/crates/weaver_forge/README.md#step-by-step-guide)
+
+Let's say we want to generate Markdown documentation for the metrics of out application or convention.
```
lquerel:
```suggestion
Let's say we want to generate Markdown documentation for the metrics of our application or convention.
```

docs/codegen.md
```diff
@@ -0,0 +1,110 @@
...
+in1 --> res
+in2 --> res
+
+res --> jq --> jinja
+
+jinja --> file1
+jinja --> file2
+```
+
+Weaver resolves the entire registry (http and grpc in this case) into a single document and. This is passed to the [JQ filter](/crates/weaver_forge/README.md#jq-filters) `semconv_grouped_metrics`. This groups individual metrics by their root namespcae (`http` or `grpc`)
```
lquerel:
```suggestion
Weaver resolves the entire registry (http and grpc in this case) into a single document. This is passed to the [JQ filter](/crates/weaver_forge/README.md#jq-filters) `semconv_grouped_metrics`. This groups individual metrics by their root namespace (`http` or `grpc`)
```

docs/validate.md
lquerel: This is where we should replace the link on weaver_forge by the weaver_checker read me link.

lquerel (approved):
Thank you very much for these improvements to the documentation!

In my opinion, there are a few minor changes to be made that I have noted in my comments. Apart from these points to address, I think this PR is ready to go.

tombrk:
@jsuereth @lquerel fixed!

PR #853, opened 2025-07-21
#850 Scope don't need attributes

### LINKED ISSUE #850: Scope shouldn't require attributes
It should be possible to define a scope which has no attributes as per the proto definition

When a scope is defined in yaml it should not either directly or indirectly be required to associate attributes with the scope.

N/a

**Additional context**
Model validation currently fails

thompson-tomo:
Closes #850

Scope's are no longer required to have attributes as per the specification


thompson-tomo:
Actually I am not adding support for scopes as that is pre-existing functionality. But removing the requirement of scopes to have attributes as per spec.

lmolkova:
we don't support scope in the [v2](https://github.com/open-telemetry/weaver/blob/main/schemas/semconv.schema.v2.json) or even [v1](https://github.com/open-telemetry/weaver/blob/main/schemas/semconv.schema.json) syntax, and `GroupType::Scope` is de-facto not used.

I'm going to close this PR given its age, but feel free to send a PR to remove `GroupType::Scope` since it's effectively dead code that would need a full set of v2 models to be supported in the future.

PR #855, opened 2025-07-24
(chore): Clean up parsing logic

jsuereth:
This is a refactoring to help out the V2 Specification work.  It includes the following:

1. Start cleaning up all the public methods in `weaver_semconv`.  We should get to a *minimal* set of methods, where it's hard to resolve semantic conventions in tests or other parts of weaver without going through some consistency checks.
2. Start using `SemConvSpecWithProvenance` instead of `(SemconvSpec, Provenance)`.  We were of two minds about this in the code, this should bring everyone to using the named tuple.
3. Update unit tests that we only validating code paths NOT used in general resolution (at least some).
4. Remove data files that were not used in tests.
5. Improve flexibility in JSON Schema validation due to an error that lead to completely useless error messages about not being able to parse the JSON schema itself.


lquerel (approved):
Really nice cleanup, it was needed!

PR #856, opened 2025-07-27
Three registry dependency chain

jerbly:
Add support for a three registry dependency chain, a->b->c. This pattern is useful when making narrow application registries that depend on a corporate registry based on the OpenTelemetry semantic conventions. Max depth 10.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ All notable changes to this project will be documented in this file.
   ([#823](https://github.com/open-telemetry/weaver/pull/823) by @lmolkova)
 - Don't serialize default values and empty arrays when resolving semantic conventions.
   ([#822](https://github.com/open-telemetry/weaver/pull/822) by @lmolkova)
+- Add support for a three registry dependency chain, a->b->c. This pattern is useful when making narrow application registries that depend on a corporate registry based on the OpenTelemetry semantic conventions.
+  ([#???](https://github.com/open-telemetry/weaver/pull/???) by @jerbly)
 
 # [0.16.1] - 2025-07-04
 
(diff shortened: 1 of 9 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ All notable changes to this project will be documented in this file.
   ([#823](https://github.com/open-telemetry/weaver/pull/823) by @lmolkova)
 - Don't serialize default values and empty arrays when resolving semantic conventions.
   ([#822](https://github.com/open-telemetry/weaver/pull/822) by @lmolkova)
+- Add support for a three registry dependency chain, a->b->c. This pattern is useful when making narrow application registries that depend on a corporate registry based on the OpenTelemetry semantic conventions.
```
lmolkova: Is there a difference between supporting a>b->c and supporting arbitrary non-circular chain?

jerbly: Not really. I added in a max-depth constant after which it will throw an error as some protection. I set this to 3 for now but could make it much larger. The circular reference check doesn't rely on the limit. Three was my use-case,  should we make it larger?

lquerel: I had the same question :-)

Having a maximum limit seems like a good idea, but I think we can go up to 10 levels to cover the most extreme cases.

crates/weaver_resolver/data/multi-registry/app_registry/app_registry.yaml
```diff
@@ -0,0 +1,12 @@
...
+    type: attribute_group
+    brief: Example attributes for the application.
+    attributes:
+      - id: app.name
+        type: string
+        brief: Name of the application.
+        stability: stable
+        requirement_level: required
+        examples: "MyApp"
+      - ref: error.type
```
lquerel, line 12: I think it would be good to test several levels of import. We have one in `data/multi-registry/custom_registry`, we could just add one in `data/multi-registry/app_registry/app_registry.yaml`.

lquerel (approved):
I just made a suggestion about the maximum depth for dependencies between registries, but otherwise it LGTM.

CHANGELOG.md
jerbly: Set this to 10

crates/weaver_resolver/data/multi-registry/app_registry/app_registry.yaml
jerbly: Added an attribute ref to the auction.name which is one level down and an import for example which is two levels down.

PR #849, opened 2025-07-16
Introduce the ability to define namespaces #802

thompson-tomo:
Closes: #802 

With the introduction of namespaces to the semconv schema we gain the ability to define in Yaml the details of the namespace needed to generate the documentation.

This would help to enable https://github.com/open-telemetry/semantic-conventions/issues/2327 to be progressed as now that general information about the namespaces can be placed on the namespace and used for documentation. At the same time we are progressing towards auto generated migration documents

This also further supports the introduction of the event registry as currently some of the event pages contains meta data about the topic which would now be able to reside as the note for the namespace.

Issue #802, opened 2025-06-22
Define Namespace items


thompson-tomo:
**Is your feature request related to a problem? Please describe.**
I want to be able to define my namespaces within yaml and have the corresponding markdown generated rather than needing to define it within readme.md files within the docs folder.

**Describe the solution you'd like**

To be able to write a yaml file similiar to the below describes my namespace and at build time generates a markdown file for me
````
groups:
  - id: namespace. Workflow
    type: namespace
    name: Workflow
    stability: development
    brief: 
    note:
    items:
        - ref: zyz
```

It would be helpful if the build process could automatically add references to any groups defined in the same folder so that we would have insight into what is available and could render it.

**Describe alternatives you've considered**
Manually maintain the linkage in markdown and remember to update it.

**Additional context**
Likely successor to #801

thompson-tomo:
A few corrections:

- Headers & footers should be used by code generation for all items in the namespace this is certainly the case for dotnet.
- Yes you can create the grouping of the signals currently but you have no way to add additional information, This is one of the things which you wanted addressed when attempting to add events to a namespace registry and is also needed to remove the resources document.

lmolkova:
please keep in mind that [code generation](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/non-normative/code-generation.md) != markdown generation. Code generation has no use for MD headers and footers for the document. 

I do not believe that it's beneficial to write free-form markdown like headers and footers in yaml with the sole goal to render them in the markdown documents.

PR #867, opened 2025-08-03
Proposal: Define namespaces in yaml

thompson-tomo:
Progresses #802 and unblocks #849 

This provides as a proposal to introducing namespace definitions alongside metrics, attributes etc,

Namespaces are a way in which definitions can be grouped together as well as a method to define & eventually generate additional useful data about that namespace. Such as examples would be revision history & migration guides.


docs/proposals/namespaces.md
```diff
@@ -0,0 +1,206 @@
...
+
+This approach can be also used for subnamespaces for instance the `gen_ai.openai.request.service_tier` attribute could be
+automatically associated with the `gen_ai.openai` namespace, 
+if the namespace definition existed. 
+
+#### Fallback linking
+
+If neither of the 2 previous methods are satisifed,
+a namespace is created by weaver which corresponds to the first path segment.
+This namespace will use the default values for a namespace.
```
jerbly, line 39: I like the idea of defining namespaces explicitly. Within Weaver right now there is no concept of namespace whatsoever **by design**. Hence your proposal.

(I took a wrong turn when developing `live-check`, I had namespace checking with the `.` separator built into rust code. This was removed and rewritten in rego policies.)

So, 
1. What if I don't want namespaces at all?
2. What if I want namespaces but a different separator? Or, just `CamelCase` (I've seen examples of this in the wild)

docs/proposals/namespaces.md
```diff
@@ -0,0 +1,206 @@
...
+
+Currently in v1 schema this can be achieved to a point via attribute groups however these do not exist in v2 schema.
+Hence this proposal becomes a useful way to migrate the information which is contained in the attribute groups to v2
+and in the process increase the capabilities of this grouping to include other signals.
+
+#### Explicit linking
+
+This is where the author explicitly states the namespace which this telemetry signal should be associated with.
+This is to help with the when the namespace of an attribute changes. 
+For instance `az` has become `azure`.
```
jerbly, line 22: We used to have `prefix` - I wonder what the reasoning was behind its removal?

docs/proposals/namespaces.md
thompson-tomo: So my thought would be that there should be a default namespace "default" which items are assigned to if the namespace can not be determined via either of the methods. I think that would handle the scenario of not wanting to use namespaces.

In an effort to keep it simple I would stick with the dot convention being the namespace seperator however if an author is using another convention  they should use the explicit option. We could always extend later.

thompson-tomo: Should we have a configuration option to use the folder as fall back source?

docs/proposals/namespaces.md
thompson-tomo: My understanding was prefix was added to the attribute keys etc whereas this explicit namespace can be thought of as aliasing.

PR #873, opened 2025-08-07
Documentation/Repo cleanup

lquerel:
Updated the documentation to increase the registry level limit from 2 to 10.

Moved outdated or obsolete docs and examples into a directory clearly identifying them as such, to reduce the risk of users reading content that is not implemented or not up to date.


nicolastakashi (approved):

docs/old-stuff/component-telemetry-schema-proposal.yaml
lmolkova, line 1: do we still need these files? can we remove them and rely on the git history ?

lquerel: Yes, I’d like to keep them for now. By moving them into a directory explicitly marked as old stuff, I figured it would be clear enough to prevent people from referring to them. There’s content in those docs I’d like to update, but I don’t have the time to tackle it at the moment.

jsuereth (approved):

PR #880, opened 2025-08-14
Fix error messages for unversioned files.  `version: ...` syntax will need fixed up later

jsuereth:
Fixes unversioned file error messages, versioned files still get horrible results.

Example new (well old, but returned) error message:

```
Diagnostic report:

invalid_semconv_group

  × Missing required property: "brief".
   ╭─[/home/joshuasuereth/test/bad.yaml:3:5]
 1 │ my_bad_yaml: test
 2 │ groups:
 3 │   - id: test
   ·     ▲
   ·     ╰── somewhere in this block
 4 │     
   ╰────
  help: A group defines an attribute group, an entity, or a signal. Supported group types are: `attribute_group`, `span`, `event`, `metric`, `entity`, `scope`.
        Mandatory fields are: `id` and `brief`.
        
        Note: The `resource` type is no longer used and is an alias for `entity`.


invalid_semconv_group

  × Object contains unexpected properties: my_bad_yaml. These properties are not defined in the schema.
  help: A semantic convention file as defined [here](https://github.com/open-telemetry/build-tools/blob/main/semantic-conventions/syntax.md) A semconv file is a
        collection of semantic convention groups (i.e. [`GroupSpec`]).
```

However, once you add the version flag, you're back to bad error messages:

```
$ ../src/open-telemetry/weaver/target/debug/weaver registry check -r $(pwd)
Weaver Registry Check
Checking registry `/home/joshuasuereth/test`
ℹ No registry manifest found: /home/joshuasuereth/test/registry_manifest.yaml

Diagnostic report:

invalid_semconv_group

  × Value {"groups":[{"id":"test"}],"my_bad_yaml":"test","version":"1"} does not match any schema in a 'oneOf' group; it must match exactly one.
  help: A versioned semantic convention file.
```

or even worse:

```
$ ../src/open-telemetry/weaver/target/debug/weaver registry check -r $(pwd)
Weaver Registry Check
Checking registry `/home/joshuasuereth/test`
ℹ No registry manifest found: /home/joshuasuereth/test/registry_manifest.yaml

Diagnostic report:

  × data did not match any variant of untagged enum SemConvSpec
```

Will fix this up as part of V2 schema efforts.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to this project will be documented in this file.
 
+# Unreleased
+
+- Fix error messages to ignore new version variants ([#880](https://github.com/open-telemetry/weaver/pull/880) by @jsuereth)
+
 # [0.17.0] - 2025-08-08
 
 - Filter based on deprecation, stability, and annotations in signal JQ helpers
--- a/crates/weaver_resolver/src/lib.rs
+++ b/crates/weaver_resolver/src/lib.rs
@@ -322,7 +322,8 @@ impl SchemaResolver {
 
         let local_path = registry_repo.path().to_path_buf();
         let registry_path_repr = registry_repo.registry_path_repr();
-        let validator = JsonSchemaValidator::new();
+        let versioned_validator = JsonSchemaValidator::new_versioned();
+        let unversioned_validator = JsonSchemaValidator::new_unversioned();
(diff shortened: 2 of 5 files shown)


jerbly (approved):
This makes sense to bring back the prior behaviour. Needs a changelog entry.

PR #894, opened 2025-08-21
Fail when JQ filter fails

lmolkova:
We're only failing when filter is invalid, but ignore execution errors. 

Fixed it, added test, and debug logs

### DIFF
--- a/crates/weaver_forge/src/jq.rs
+++ b/crates/weaver_forge/src/jq.rs
@@ -36,6 +36,9 @@ pub fn execute_jq(
     // Note: This will be exposed with `${key}` as the variable name.
     params: &BTreeMap<String, serde_json::Value>,
 ) -> Result<serde_json::Value, Error> {
+    // TODO: save input into temp file for debugging if debug is enabled
+    log::debug!("Executing JQ filter: {} ...", filter_expr);
+
     let loader = Loader::new(
         // ToDo: Allow custom preludes?
         jaq_std::defs()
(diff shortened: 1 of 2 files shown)


crates/weaver_forge/src/jq.rs
```diff
@@ -75,15 +78,19 @@ pub fn execute_jq(
     let ctx = Ctx::new(values, &inputs);
 
     // Bundle Results
-    let mut errs = Vec::new();
```
lmolkova: me every time rust developer ignores errors

image

crates/weaver_forge/src/jq.rs
```diff
@@ -75,15 +78,19 @@ pub fn execute_jq(
...
-    for r in filter_result {
-        match r {
-            Ok(v) => values.push(serde_json::Value::from(v)),
-            Err(e) => errs.push(e),
-        }
-    }
+    let mut values = filter
+        .run((ctx, Val::from(input.clone())))
+        .collect::<Result<Vec<_>, _>>()
+        .map_err(|e| Error::FilterError {
```
jsuereth: So this returns first error right now (via `?`) vs. all of them.  I think this is a fine improvement, but we may want to consider collecting all errors, returning them ALL immediately across all filters, but this is fine as an initial fix.

jsuereth (approved):

crates/weaver_forge/src/jq.rs
lmolkova: I see, that's why we had this logic to collect errors! let me see if I can fix it

defaults/jq/semconv.jq
```diff
@@ -84,7 +84,7 @@ def deprecated_filter($options):
 def semconv_attributes($options):
     .groups
     | map(select(.type == "attribute_group" and (.id | startswith("registry."))))
-    | map(.attributes) | add
+    | map(.attributes) | add // []
```
lmolkova, line 87: hours_spent_fighting_jq += ~3

docs/usage.md
```diff
@@ -12,7 +12,7 @@ Commands:
   help        Print this message or the help of the given subcommand(s)
 
 Options:
-      --debug...  Turn debugging information on
+      --debug...  Turn debugging information on. Use twice (--debug --debug) for trace-level logs.
```
jerbly, line 15: Seems weird. Why not have --trace?

lmolkova: yeah, I was following existing format:

https://github.com/open-telemetry/weaver/blob/7beff49c71cf2d17a04071b72386138fcfabd2b6/src/cli.rs#L21-L22

and was wondering if there is a convention. I realized it's similar to what curl does with `-v`, `-vv`, `-vvv`. Curious how @lquerel envisioned it.

crates/weaver_forge/src/jq.rs
lmolkova: fixed now

docs/usage.md
lquerel: The intention was to be able to define the verbosity level. It’s common practice to have `-v`, `-vv`, `-vvv`, but I didn’t specify that I wanted the short `-v` and not `-d`. Sorry for this oversight. If you prefer to use `--trace`, I have no objection.

crates/weaver_forge/src/jq.rs
lquerel: The errors were carefully collected in a Vec with, I assume, the intention of reporting them. From my point of view, it looks more like an unfortunate oversight in not doing something with these errors than negligence. If I’m responsible, sorry.

lquerel (approved):

PR #899, opened 2025-08-27
Support automatic deprecation inheritance for enum members

### LINKED ISSUE #878: Deprecation of enum members when attribute is deprecated

Currently when we deprecate an attribute we are needing to also explicitly deprecate all of the enum members. But if we don't the members will still show with their original stability.

If an attribute is deprecated, the members should automatically be deprecated.

Only update the stability in rendered docs.

**Additional context**

rpotluri-f5:
When an attribute with enum type is deprecated enum members now automatically inherit the deprecated status. This eliminates the need to manually deprecate each enum member individually, reducing maintenance overhead and preventing inconsistencies.

Fixes #878

### DIFF
--- a/crates/weaver_semconv/src/attribute.rs
+++ b/crates/weaver_semconv/src/attribute.rs
@@ -204,6 +204,34 @@ impl AttributeSpec {
             AttributeSpec::Id { tag, .. } => tag.clone(),
         }
     }
+
+    /// Apply stability inheritance to enum members when parent attribute is deprecated.
+    /// This mutates the attribute in-place to propagate deprecation status.
+    pub fn apply_enum_stability_inheritance(&mut self) {
+        if let AttributeSpec::Id {
+            r#type: Enum { members },
+            deprecated,
+            stability,
+            ..
+        } = self
+        {
+            // Only apply inheritance if parent attribute is deprecated
+            if deprecated.is_some() {
+                for member in members {
+                    // Only inherit stability if member doesn't have explicit stability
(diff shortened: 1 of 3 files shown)


crates/weaver_semconv/src/attribute.rs
```diff
@@ -204,6 +204,34 @@ impl AttributeSpec {
...
+                for member in members {
+                    // Only inherit stability if member doesn't have explicit stability
+                    if member.stability.is_none() {
+                        // Inherit the parent's stability (if any), or default to experimental
+                        member.stability = stability.clone().or(Some(Stability::Development));
+                    }
+                    // Only inherit deprecation if member doesn't have explicit deprecation
+                    if member.deprecated.is_none() {
+                        // Inherit the parent's deprecation status
+                        member.deprecated = deprecated.clone();
```
thompson-tomo, line 229: @rpotluri-f5 i think we need to extend this logic to be such that if the attribute deprecation reason is obsoleted, then it is cloned otherwise the enum member deprecation reason should be uncategorised, note/brief should be copied across and nothing else. This is to handle the fact that an attribute rename is different to a member rename as the renamed to have different targets.

rpotluri-f5:
> Thanks for the contribution, @rpotluri-f5! Unfortunately, I don't think we should have this feature. I've left a comment on the issue - [#878 (comment)](https://github.com/open-telemetry/weaver/issues/878#issuecomment-3239397982) to explain why.

Your explanation makes sense. I will pick another open item to contribute to.

PR #906, opened 2025-08-31
Bumps prost and saphyr 

jerbly:
Bumps and associated fixes for prost and saphyr.

Clears #797 and #786.


lquerel (approved):
LGTM. 

The fix around spans in Saphyr has the potential to improve a bit the integration with miette and the error reporting.

jsuereth (approved):

PR #921, opened 2025-09-05
chore: add cross arch build identifier for darwin/arm64

codeboten:
Add `darwin/arm64` in the list of recognized `TARGETPLATFORM` to make getting started a little bit easier on MacOS.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -79,12 +79,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "683d7910e743518b0e34f1186f92494becacb047c7b6bf616c96772180fef923"
 
 [[package]]
-name = "android-tzdata"
-version = "0.1.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e999941b234f3131b00bc13c22d06e8c5ff726d1b6318ac7eb276997bbb4fef0"
-
-[[package]]
 name = "android_system_properties"
 version = "0.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


cross-arch-build.sh
```diff
@@ -10,6 +10,8 @@ elif [ "${TARGETPLATFORM}" = "linux/arm64" ]; then
   if [ "${TARGETPLATFORM}" != "${BUILDPLATFORM}" ]; then
     apt-get update && apt-get install -y gcc-aarch64-linux-gnu
   fi
+elif [ "${TARGETPLATFORM}" = "darwin/arm64" ]; then
+  RUST_TARGET=aarch64-apple-darwin
```
jsuereth, line 14: Does this need any gcc packages to run in our build runners?

Also, are you calling this script as part of your developer workflow or planning to fully add a darwin/arm64 docker image?

codeboten: i didn't need any additional packages as i was using this locally, haven't tried to put this in a container just yet, but can do so if it's useful

jsuereth: Is this needed to build locally?  This script is designed for building our containers.  I don't think we'd recommend using this script (with `apt-get` calls) in a dev scenario.

I'd love to better understand the friction on developing locally.  If it's not just a simple `cargo <cmd>` we probably need to do something to fix that, or provide a "setup-dev-env" script you would run once.

jerbly: I always use zigbuild for local builds for other targets. I use a Mac so:
`cargo zigbuild --release --target x86_64-unknown-linux-gnu`

codeboten: this is not needed for local builds, happy to close this.

> I'd love to better understand the friction on developing locally. If it's not just a simple cargo <cmd> we probably need to do something to fix that, or provide a "setup-dev-env" script you would run once.

i ran the steps in the contributing doc but it wasn't immediately clear how to build the binary locally (total rust noob here), so i ended up looking at the CI files to figure out how to get a binary. i will open a separate PR adding a note in the contributing doc to clarify

codeboten: PR opened here: https://github.com/open-telemetry/weaver/pull/940

PR #922, opened 2025-09-07
Live-check stats: don't serialize 0 when reporting seen attributes and metrics

lmolkova:
a tiny improvement for the report.

It currently contains ~1000 of extra lines that look like 

```json
...
    "seen_registry_attributes": {
      ...
      "vcs.ref.type": 0,
      "vcs.repository.name": 0,
      "vcs.repository.url.full": 0,
      "vcs.revision_delta.direction": 0,
      "webengine.description": 0,
      "webengine.name": 0,
      "webengine.version": 0,
      "zos.smf.id": 0,
      "zos.sysplex.name": 0
    },
    "seen_registry_metrics": {
      "aspnetcore.authentication.authenticate.duration": 0,
      "aspnetcore.authentication.challenges": 0,
      "aspnetcore.authentication.forbids": 0,
      "aspnetcore.authentication.sign_ins": 0,
      "aspnetcore.authentication.sign_outs": 0,
      "aspnetcore.authorization.attempts": 0,
      "aspnetcore.diagnostics.exceptions": 0,
      ...
    }
```

this pr customizes serialization to not print `seen_registry*` keys with 0 value.

### DIFF
--- a/crates/weaver_live_check/src/lib.rs
+++ b/crates/weaver_live_check/src/lib.rs
@@ -20,6 +20,18 @@ use weaver_common::diagnostic::{DiagnosticMessage, DiagnosticMessages};
 use weaver_forge::registry::{ResolvedGroup, ResolvedRegistry};
 use weaver_semconv::group::GroupType;
 
+/// Helper function to serialize HashMap<string, usize> excluding zero values
+fn serialize_map_non_zero_values<S>(map: &HashMap<String, usize>, serializer: S) -> Result<S::Ok, S::Error>
+where
+    S: serde::Serializer,
+{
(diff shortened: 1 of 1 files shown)


jerbly:
One of the functions of the live-check report is to give you coverage. So, if you see a coverage < 100% it's useful to see which attributes or metrics you did not cover.

I agree though that this doesn't make sense if you're using the full semconv registry to check against.

Perhaps we should have `coverage` as an option which could then switch the serialization?

lmolkova:
> Perhaps we should have coverage as an option which could then switch the serialization?

makes sense, let me try to repurpose this PR for this

PR #923, opened 2025-09-07
Live-check: add min-advice-level option to control advice verbosity

lmolkova:


crates/weaver_live_check/README.md
```diff
@@ -107,9 +107,13 @@ As mentioned, a list of `Advice` is returned in the report for each sample entit
 }
 ```
 
-> **Note**  
+> **Note**
 > The `live_check_result` object augments the sample entity at the pertinent level in the structure. If the structure is `metric`->`[number_data_point]`->`[attribute]`, advice should be give at the `number_data_point` level for, say, required attributes that have not been supplied. Whereas, attribute advice, like `missing_attribute` in the JSON above, is given at the attribute level.
 
+#### Customize advise level
+
+By default, live-check shows `information`, `improvement` and `violation` advices. You can provide `--min-advice-level` argument to only show advices of this and higher levels.
```
jerbly, line 115:
```suggestion
By default, live-check shows `information`, `improvement` and `violation` advice. You can provide `--min-advice-level` argument to only show advices of this and higher levels.
```

lmolkova:
Discussed at the weaver call, the conclusion:

The full report contains a lot of details which are useful too. 
Let's focus on a compact version like a list of advices grouped by source (attribute / signal), violation type and level with number of them.

We could print this summary at the bottom of ansi report and have the section in the json report. We could also print it to stderr if `-output` is used.

PR #928, opened 2025-09-10
Fail when template doesn't match any file

lmolkova:


crates/weaver_codegen_test/templates/registry/rust/weaver.yaml
```diff
@@ -22,12 +22,6 @@ comment_formats:
 default_comment_format: rust
 
 templates:
-  - template: README.md
-    filter: .
-    application_mode: single
-  - template: lib.rs
-    filter: .
-    application_mode: single
```
lmolkova, line 30: templates don't exist

crates/weaver_forge/templates/whitespace_control/weaver.yaml
```diff
@@ -25,45 +25,3 @@ templates:
   - template: "**/registry.md"
     filter: "."
     application_mode: single
-  - template: "**/attribute_group.md"
```
lmolkova, line 28: these templates don't exist

src/main.rs
```diff
@@ -102,7 +102,7 @@ fn main() {
             .format(|buf, record| writeln!(buf, "{}", record.args()))
             .init();
 
-        log_error(format!("Debug is set to {}", cli.debug));
+        log::debug!("Debug is set to {}", cli.debug);
```
lmolkova, line 105: unrelated, this sneaked in by mistake from some prev PR

crates/weaver_forge/src/lib.rs
```diff
@@ -198,7 +198,7 @@ impl TemplateEngine {
         mut config: WeaverConfig,
         loader: impl FileLoader + Send + Sync + 'static,
         params: Params,
-    ) -> Self {
+    ) -> Result<Self, Error> {
```
lquerel, line 201: nit: In Rust, by convention, a fallible constructor is often called `try_new`.

lmolkova: cool, updated

PR #933, opened 2025-09-12
Attribute groups in schema v2

lmolkova:
Implements https://github.com/open-telemetry/weaver/issues/824#issuecomment-3194585303

TODO:
- [x] changelog
- [x] tests
- [x] JSON schema and doc

Semconv PR to update http and DB: https://github.com/open-telemetry/semantic-conventions/compare/main...lmolkova:semantic-conventions:schema-v2-http-db

Example

```yaml
  - id: attributes.db.client.authority
    visibility: internal
    attributes:
      - ref: server.address
        brief: Name of the database host.
      - ref: server.port
        requirement_level:
          conditionally_required: If using a port other than the default port for this DBMS and if `server.address` is set.
```

```yaml
metrics:
  - name: db.client.operation.duration
    ...
    attributes:
      - ref_group: attributes.db.client.query
      - ref_group: attributes.db.client.response
      - ref_group: attributes.db.client.authority
      - ref_group: attributes.db.client.network
      - ref_group: attributes.db.client.operation
      - ref: db.system.name
        requirement_level: required
      - ref: db.stored_procedure.name
        requirement_level:
          recommended: If operation applies to a specific stored procedure.
...
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 # Unreleased
 
 - Add support for attribute_group (public and internal) in schema v2 ([#933](https://github.com/open-telemetry/weaver/pull/933) by @lmolkova)
+- Live-check report changes ([#943](https://github.com/open-telemetry/weaver/pull/943) by @lmolkova)
+  - 💥 BREAKING CHANGE 💥 `value` property in `Advice` is renamed to `advice_context`
+  - Advice now contains `signal_type` and `signal_name` properties to simplify post-processing
+  - Message format is changed to include all dynamic details about the advice
+- The target and the `registry` sub-directory are now optional for the
+  `weaver registry generate` command. ([#962](https://github.com/open-telemetry/weaver/pull/962) by @lquerel)
 
 # [0.18.0] - 2025-09-17
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -403,12 +403,6 @@ checksum = "d71b6127be86fdcfddb610f7182ac57211d4b18a3e9c82eb2d17662f2227ad6a"
 
 [[package]]
 name = "bytesize"
-version = "1.3.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2e93abca9e28e0a1b9877922aacb20576e05d4679ffa78c3d6dc22a26a216659"
-
-[[package]]
-name = "bytesize"
 version = "2.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a3c8f83209414aacf0eeae3cf730b18d6981697fba62f200fcfb92b9f082acba"
@@ -669,9 +663,9 @@ checksum = "19d374276b40fb8bbdee95aef7c7fa6b5316ec764510eb64b8dd0e2ed0d7e7f5"
 
 [[package]]
 name = "crc32fast"
-version = "1.4.2"
+version = "1.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a97769d94ddab943e4510d138150169a2758b5ef3eb191a9ee688de3e23ef7b3"
+checksum = "9481c1c90cbf2ac953f07c8d4a58aa3945c425b7185c9154d67a65e4230da511"
 dependencies = [
  "cfg-if",
 ]
@@ -976,15 +970,6 @@ dependencies = [
 
 [[package]]
 name = "faster-hex"
-version = "0.9.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a2a2b11eda1d40935b26cf18f6833c526845ae8c41e58d09af6adeb6f0269183"
-dependencies = [
- "serde",
-]
-
-[[package]]
-name = "faster-hex"
 version = "0.10.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7223ae2d2f179b803433d9c830478527e92b8117eab39460edae7f1614d9fb73"
@@ -1019,9 +1004,9 @@ checksum = "1d674e81391d1e1ab681a28d99df07927c6d4aa5b027d7da16ba32d1d21ecd99"
 
 [[package]]
 name = "flate2"
-version = "1.1.2"
+version = "1.1.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4a3d7db9596fecd151c5f638c0ee5d5bd487b6e0ea232e5dc96d5250f6f94b1d"
+checksum = "dc5a4e564e38c699f2880d3fda590bedc2e69f3f84cd48b457bd892ce61d0aa9"
 dependencies = [
  "crc32fast",
  "libz-rs-sys",
@@ -1194,9 +1179,9 @@ checksum = "07e28edb80900c19c28f1072f2e8aeca7fa06b23cd4169cefe1af5aa3260783f"
 
 [[package]]
 name = "gix"
-version = "0.71.0"
+version = "0.73.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a61e71ec6817fc3c9f12f812682cfe51ee6ea0d2e27e02fc3849c35524617435"
+checksum = "514c29cc879bdc0286b0cbc205585a49b252809eb86c69df4ce4f855ee75f635"
 dependencies = [
  "gix-actor",
  "gix-attributes",
@@ -1207,11 +1192,11 @@ dependencies = [
  "gix-date",
  "gix-diff",
  "gix-discover",
- "gix-features 0.41.1",
+ "gix-features",
  "gix-filter",
- "gix-fs 0.14.0",
+ "gix-fs",
  "gix-glob",
- "gix-hash 0.17.0",
+ "gix-hash",
  "gix-hashtable",
  "gix-ignore",
  "gix-index",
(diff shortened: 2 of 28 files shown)


crates/weaver_semconv/src/v2/attribute_group.rs
```diff
@@ -0,0 +1,184 @@
...
+/// A group defines an attribute group, an entity, or a signal.
+/// Mandatory fields is: `id`. Groups are expected to have `attributes`,
+/// `include_groups` or both
+#[derive(Serialize, Deserialize, Debug, Clone, JsonSchema)]
+#[serde(tag = "visibility")]
+#[serde(rename_all = "snake_case")]
+#[serde(deny_unknown_fields)]
+pub enum AttributeGroup {
+    /// An internal attribute group
+    Internal {
```
jsuereth: As discussed in the meeting, can we make this be:

```
pub enum AttributeGroup {
  Internal(InternalAttributeGroup),
  Public(PublicAttributeGroup),
}
pub struct InternalAttributeGroup { ... }
pub struct PublicAttributeGroup { ... }
```

?

crates/weaver_resolver/src/registry.rs
```diff
@@ -543,30 +583,77 @@ fn resolve_extends_references(ureg: &mut UnresolvedRegistry) -> Result<(), Error
...
+                        // check if any of the attr in attrs is already in the all_attrs
+                        // and fail, otherwise add them to all_attrs
+                        for attr in attrs {
+                            if attr_ids.contains_key(&attr.spec.id()) {
+                                log::debug!(
+                                    "Found duplicate attribute id {} in include_groups for group {}",
+                                    attr.spec.id(),
+                                    unresolved_group.group.id
+                                );
+                                // TODO: accumulate real errors and return all of them
```
jsuereth: I believe you can actually just:

```
errors.push(Error::DuplicateAttributeId { ... });
all_resolved = false;
```

And let this continue to the error reporting below.

crates/weaver_resolved_schema/src/registry.rs
```diff
@@ -70,6 +71,11 @@ pub struct Group {
     /// convention.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub extends: Option<String>,
+    /// Attribute groups to include - this parameter must not be provided
+    /// in yaml, it's only used to convert v2 schema into v1
+    #[serde(default)]
+    #[serde(skip_serializing)]
+    pub include_groups: Vec<String>,
```
jsuereth: Do we need this in the resolved schema?

crates/weaver_resolver/src/registry.rs
```diff
@@ -603,16 +689,25 @@ fn resolve_inheritance_attrs(
...
-            AttrWithLineage {
-                spec: parent_attr.spec.clone(),
-                lineage: AttributeLineage::inherit_from(parent_group_id, &parent_attr.spec),
-            },
-        );
+    // Inherit the attributes from all parent groups.
+    for (parent_group_id, included_group) in include_groups {
+        for parent_attr in included_group.iter() {
+            let attr_id = parent_attr.spec.id();
+            let lineage = AttributeLineage::inherit_from(parent_group_id, &parent_attr.spec);
```
jsuereth, line 699: We don't really want internal attribute groups to show up in lineage here.  The constructed view should be against the groups which "stay" in the resolved schema.

I believe what we want to do is "flatten" the lineage when there's an internal group in the middle.

I actually need to seriously revamp lineage for V2 schema output, so it's possible we limp along with things in an odd state for internal groups until that work is done.

lmolkova: this method (resolve_inheritance_attrs_unified) updates lineage on the go and eventually resolves everything to `registry` - groups in the middle never show up in the result regardless of their visibility.

I think we need to revisit lineage:
- decide what we do with registry - there is no group anymore - we should probably replace source_group with file path
- do we need to include any intermediary public groups? if so, how and what for? we haven't done it so far.

I think this code could be significantly simplified as a result.

crates/weaver_semconv/src/v2/attribute_group.rs
lmolkova: we can, but Rust is actively pushing back on it. It considers them unnecessarily qualified (since they are defined here and never used on their own)

```
error: unnecessary qualification
  --> crates/weaver_semconv/src/v2/attribute_group.rs:54:14
   |
54 |     Internal(InternalAttributeGroup),
   |              ^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: requested on the command line with `-D unused-qualifications`
```

Note, I cannot even suppress it with `#[allow(unused_qualifications)]` - according to chatgpt 

https://github.com/open-telemetry/weaver/blob/66f0dcd0ac439563f55273d21c4818661e81f7df/Cargo.toml#L144

has ultimate power over local suppression. 

----------------------

I'm going to  change `unused_qualifications` to `warn` in `Cargo.toml` to make it possible.

crates/weaver_resolver/data/registry-test-v2-2-multifile/expected-registry.json
```diff
@@ -1,17 +1,6 @@
 {
   "registry_url": "https://127.0.0.1",
   "groups": [
-    {
-      "id": "registry.data.registry-test-v2-2-multifile.registry.http",
```
lmolkova, line 5: no attributes defined, no need to create a group

crates/weaver_semconv/src/v2/attribute_group.rs
lmolkova: Clippy also doesn't like unused_qualifications and I had to allow them. Do we really want to go against Rust here?

.github/workflows/ci.yml
```diff
@@ -177,7 +177,7 @@ jobs:
         # [workaround] removed --all-features due to an issue in one of the dependency in Tantity (zstd-safe)
         # [ToDo LQ] Re-enable --all-features once the issue is resolved
         run: >
-          cargo clippy --workspace --all-targets --message-format=json -- -D warnings --allow deprecated
+          cargo clippy --workspace --all-targets --message-format=json -- -D warnings --allow deprecated --allow unused_qualifications
```
lquerel: We shouldn't need to allow this, let alone globally. I need to play with the code locally on my laptop to understand the root cause of the problem, because the PR is too large for me to identify this from a simple review.

lmolkova: This is the problem - https://github.com/open-telemetry/weaver/pull/933#discussion_r2358967305

crates/weaver_semconv/src/v2/attribute_group.rs
lquerel: I have to admit I don't understand why the compiler emits a warning in this context. It seems to be related to a combination of code generated by derive macros and this nesting at the level of the enum variants.

I'm also not clear on the reasons for moving from the struct-variant approach to a variant that references an external struct. If the rationale behind this refactor isn’t driven by a strong need, I'd be more inclined to return to the initial approach than to disable the `unused_qualifications` lint.

.github/workflows/ci.yml
lquerel: Thanks. See my answer in the other comment.

crates/weaver_resolver/src/registry.rs
lmolkova: discussed on weaver call, will work on lineage separately.

crates/weaver_semconv/src/v2/attribute_group.rs
lmolkova: Updated to use code generated by macro, removed unused_qualifications suppression.

schemas/semconv-syntax.v2.md
```diff
@@ -394,9 +455,68 @@ metrics:
...
+
+#### Public Attribute Groups
+
+Public attribute groups are included in generated documentation and resolved schema. A group of attributes is meaningful and
+describes some complex object (such as exception, thread, CloudEvent, etc.) - these attributes may be included in
+different signals as a group.
+
+A public attribute group definition consists of the following properties:
+
+- `id` - Required. Uniquely identifies the attribute group.
```
thompson-tomo, line 493: Should we not call this key and then generate an id for the group?

This key should also be included in the resolved schema as a property of the attribute on a signal when it is a public group otherwise empty. This is to help with documentation.

lmolkova: I don't understand the proposal. This is a group id, key is used for attributes themselves, `id` matches today's structure and would be more intuitive than key.

Public group ids would eventually be available on via lineage in the resolved schema after we figure out how lineage should work in v2 (https://github.com/open-telemetry/weaver/pull/933#discussion_r2358996060)

lmolkova:
> I also might have missed it but are we emitting public attribute groups as a part of the resolved schema and can they be distinguished from internal?

yes, please take a look at the syntax doc.

lquerel (approved):
LGTM

schemas/semconv-syntax.v2.md
thompson-tomo: I see it as the same as attributes as you define a key which you then use to reference it from within an attributes array. Whereas an id should be unique across signals which is why id is usually prefixed with the signal type. 

If it is more intuitive then it should be the same for attributes given they aim for the same thing which is adding 1 or more attributes to a signal.

I agree lineage could be made to work but it currently doesn't work. we would need to ensure that it can be distinguished from simply coming from the registry. This is so we can link from a span to the public group definition which can provide additional details about sourcing the values for instance.

lmolkova: I don't agree that attributes are the same as attribute groups. Also I don't see how renaming `id` to `key` makes anything better.

lmolkova:
> > > I also might have missed it but are we emitting public attribute groups as a part of the resolved schema and can they be distinguished from internal?
> > > yes, please take a look at the syntax doc.
> 
> Isn't the syntax document describing what a user writes (un-resolved), whereas the resolved is what is emitted by Weaver? In effect shouldn't we have the visibility field included in https://github.com/open-telemetry/weaver/blob/561ff28a4e3e81f9175457cda7bf63199b59fd5b/crates/weaver_resolved_schema/src/registry.rs
> 
> Also doesn't your comment in [#933 (comment)](https://github.com/open-telemetry/weaver/pull/933#discussion_r2370328821) confirm that groups are no longer present?

this PR does not implement resolved schema v2 - it does not exist yet, @jsuereth is working on it. After this PR v2 internal groups are do not appear in the resolve schema. public groups do. All v1 groups appear. It's still an intermediate state

schemas/semconv-syntax.v2.md
thompson-tomo: For me It is all about how it would be used.

Let's think about some of the use cases where an identifying label for a group could/would be used:

- rendering a title of the group in docs. Currently  string manipulation is done
- Rendered in attribute tables on signals so link to group can be implemented.
- Added on signal definitions to add the attributes
- Code generating methods to add the groups to signals

Hence considering the above and that id's are implicitly used for internal linking, we should either add a name alongside the id with both being user defined or internally generate the id based on the key and type.

jsuereth (approved):

PR #943, opened 2025-09-17
Live-check reporting improvements

lmolkova:
This change is an update of #923 and simplifies reporting with weaver.

The goal is to print short violation-only reports like (which I can now do using custom weaver.yaml and templates)

```
Violations:
- [required_attribute_not_present] Required attribute `server.port` is not present. (3 occurrence(s)
 on metric `http.client.request.duration`)
- [missing_attribute] Attribute `asgi.event.type` does not exist in the registry. (2 occurrence(s))
- [deprecated] Attribute `db.name` is deprecated; reason = renamed, note = Replaced by `db.namespace`. (1 occurrence(s))
- [deprecated] Attribute `db.statement` is deprecated; reason = renamed, note = Replaced by `db.query.text`. (1 occurrence(s))
- [deprecated] Attribute `db.system` is deprecated; reason = renamed, note = Replaced by `db.system.name`. (1 occurrence(s))
- [deprecated] Attribute `db.user` is deprecated; reason = obsoleted, note = Removed, no replacement at this time. (1 occurrence(s))
- [deprecated] Attribute `net.peer.name` is deprecated; reason = uncategorized, note = Replaced by `server.address` on client spans and `client.address` on server spans. (1 occurrence(s))
- [deprecated] Attribute `net.peer.port` is deprecated; reason = uncategorized, note = Replaced by `server.port` on client spans and `client.port` on server spans. (1 occurrence(s))
- [deprecated] Attribute `net.transport` is deprecated; reason = renamed, note = Replaced by `network.transport`. (1 occurrence(s))

Seen: 12 metric(s), 5 span(s), 0 log(s), 4 resource(s)
```

Changes:
1. Makes advices self-contained: advices now include signal name and type, attribute name (when known / applicable). It simplifies JQ A LOT if I want to include context into the report. E.g. I can do `.. | objects | select(has("live_check_result"))` and it'd return all advices regardless of their nesting with full details
2. Advice message is plain english that describes violation. Then consumers don't need to custom-format or interpret the advice
3. Advice values are structured. E.g. `attribute_name = test.me` is better than `test.me` with meaning that depends on the context.

Default report update:

New:
   image

Old:
   image

### DIFF
--- a/crates/weaver_checker/src/lib.rs
+++ b/crates/weaver_checker/src/lib.rs
@@ -106,7 +106,7 @@ pub enum Error {
         /// The provenance of the violation (URL or path).
         provenance: String,
         /// The violation.
-        violation: Violation,
+        violation: Box<Violation>,
     },
 
     /// A container for multiple errors.
--- a/crates/weaver_checker/src/violation.rs
+++ b/crates/weaver_checker/src/violation.rs
@@ -93,8 +93,9 @@ pub struct Advice {
     /// machine-readable string that categorizes the advice.
     pub advice_type: String,
 
-    /// The context associated with the advice e.g. { "attribute_name": "foo.bar" }
-    /// Context values may be used with custom templates and filters to query, summarize, and format advice.
+    /// The context associated with the advice e.g. { "attribute_name": "foo.bar", "attribute_value": "bar" }
+    /// The value should contain all dynamic parts of the message in a structured way.
+    /// Context values may be used with custom templates and filters to customize reports.
     pub value: Value,
 
     /// The human-readable message of the advice e.g. "This attribute 'foo.bar' is deprecated, reason: 'use foo.baz'"
--- a/crates/weaver_checker/src/violation.rs
+++ b/crates/weaver_checker/src/violation.rs
@@ -88,7 +88,7 @@ pub enum AdviceLevel {
 
 /// Represents a live check advice
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize, JsonSchema)]
-pub struct         Advice {
+pub struct Advice {
     /// The type of advice e.g. "is_deprecated". This should be a short,
     /// machine-readable string that categorizes the advice.
     pub advice_type: String,
--- a/crates/weaver_checker/src/violation.rs
+++ b/crates/weaver_checker/src/violation.rs
@@ -43,7 +43,7 @@ impl Display for Violation {
             }
             Violation::Advice(Advice {
                 advice_type: r#type,
-                value,
+                advice_context,
                 message,
                 advice_level,
                 signal_type,
@@ -51,7 +51,7 @@ impl Display for Violation {
             }) => {
                 write!(
                     f,
-                    "type={type}, value={value}, message={message}, advice_level={advice_level:?}, signal_type={signal_type:?}, signal_name={signal_name:?}"
+                    "type={type}, context={advice_context}, message={message}, advice_level={advice_level:?}, signal_type={signal_type:?}, signal_name={signal_name:?}"
                 )
             }
         }
@@ -88,20 +88,20 @@ pub enum AdviceLevel {
 
 /// Represents a live check advice
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize, JsonSchema)]
-pub struct Advice {
+pub struct         Advice {
     /// The type of advice e.g. "is_deprecated". This should be a short,
     /// machine-readable string that categorizes the advice.
     pub advice_type: String,
 
     /// The context associated with the advice e.g. { "attribute_name": "foo.bar", "attribute_value": "bar" }
-    /// The value should contain all dynamic parts of the message in a structured way.
+    /// The context should contain all dynamic parts of the message
     /// Context values may be used with custom templates and filters to customize reports.
-    pub value: Value,
+    pub advice_context: Value,
 
     /// The human-readable message of the advice e.g. "This attribute 'foo.bar' is deprecated, reason: 'use foo.baz'"
     /// The message, along with signal_name and signal_type, should contain enough information to understand the advice and
     /// identify the issue and how to fix it.
-    /// Some of the values used in the message may be also present in the `value` field to support report customization.
+    /// Some of the values used in the message may be also present in the `advice_context` field to support report customization.
     pub message: String,
 
     /// The level of the advice e.g. "violation"
--- a/crates/weaver_checker/src/violation.rs
+++ b/crates/weaver_checker/src/violation.rs
@@ -46,10 +46,12 @@ impl Display for Violation {
                 value,
                 message,
                 advice_level,
+                signal_type,
+                signal_name,
             }) => {
                 write!(
                     f,
-                    "type={type}, value={value}, message={message}, advice_level={advice_level:?}"
+                    "type={type}, value={value}, message={message}, advice_level={advice_level:?}, signal_type={signal_type:?}, signal_name={signal_name:?}"
                 )
             }
         }
@@ -95,4 +97,9 @@ pub struct Advice {
     pub message: String,
     /// The level of the advice e.g. "violation"
     pub advice_level: AdviceLevel,
+
+    /// The signal type the advice applies to e.g. "span", "metric", "resource"
+    pub signal_type: Option<String>,
+    /// The signal name the advice applies to e.g. "http.server.duration"
+    pub signal_name: Option<String>,
 }
--- a/crates/weaver_checker/src/violation.rs
+++ b/crates/weaver_checker/src/violation.rs
@@ -89,17 +89,26 @@ pub enum AdviceLevel {
 /// Represents a live check advice
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize, JsonSchema)]
 pub struct Advice {
-    /// The type of advice e.g. "is_deprecated"
+    /// The type of advice e.g. "is_deprecated". This should be a short,
+    /// machine-readable string that categorizes the advice.
     pub advice_type: String,
-    /// The value of the advice e.g. "true"
+
+    /// The context associated with the advice e.g. { "attribute_name": "foo.bar" }
+    /// Context values may be used with custom templates and filters to query, summarize, and format advice.
     pub value: Value,
-    /// The message of the advice e.g. "This attribute is deprecated"
+
+    /// The human-readable message of the advice e.g. "This attribute 'foo.bar' is deprecated, reason: 'use foo.baz'"
+    /// The message, along with signal_name and signal_type, should contain enough information to understand the advice and
+    /// identify the issue and how to fix it.
+    /// Some of the values used in the message may be also present in the `value` field to support report customization.
     pub message: String,
+
     /// The level of the advice e.g. "violation"
     pub advice_level: AdviceLevel,
 
-    /// The signal type the advice applies to e.g. "span", "metric", "resource"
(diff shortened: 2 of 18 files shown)


crates/weaver_live_check/src/sample_attribute.rs
```diff
@@ -182,31 +182,40 @@ impl LiveCheckRunner for SampleAttribute {
...
 
+        let signal_type = parent_group.as_ref().map(|g| format!("{:?}", g.r#type));
+        let signal_name = parent_group.as_ref().map(|g| g.id.clone());
         if semconv_attribute.is_none() {
             result.add_advice(Advice {
                 advice_type: MISSING_ATTRIBUTE_ADVICE_TYPE.to_owned(),
-                value: Value::String(self.name.clone()),
-                message: "Does not exist in the registry".to_owned(),
+                value: json!({ "attribute_name": self.name.clone() }),
+                message: format!("Attribute `{}` does not exist in the registry.", self.name),
```
lmolkova: let's check what default template would do

lmolkova:
> But, maybe also for this specific case you could enhance the stats object to include all the detail needed for this summary. > Then you'll need close to zero JQ hopefully. Since we're doing this aggregation already for the stats we could have an struct there instead of just a count.

I'm less interested in a specific case, but more in the ability to generate whatever reports and aggregate in whatever way. I hope eventually we can find the golden path, but for now I don't mind a bit of not complicated jq AI can write on the first try. I tried writing JQ over current advices and me and AI both failed miserably.

PS: I updated default ansi template and attached some screenshots in the description. curious what you think @jerbly

jerbly:
Another thought. What if we leave the value and message back as it was but add a new verbose_message where you can embed the context in the message text?

lmolkova:
@jerbly 

here you go

```
Resource
    service.name = my_service

Span test `client`
    http.response.status_code = foo
        - [violation] Attribute `http.response.status_code` has type `string`. Type should be `int`.
    aws.s3.bucket = value
        - [improvement] Attribute `aws.s3.bucket` is not stable; stability = development.
    aws.s3.bucket.name = value
        - [violation] Attribute `aws.s3.bucket.name` does not exist in the registry.
        - [information] Extends existing namespace
        - [violation] Namespace matches existing attribute
    task.id = value
        - [violation] Attribute `task.id` does not exist in the registry.
    TaskId = value
        - [violation] Attribute `TaskId` does not exist in the registry.
        - [improvement] Does not have a namespace
        - [violation] Does not match name formatting rules
    aws.s3.extension.name = foo
        - [violation] Attribute `aws.s3.extension.name` does not exist in the registry.
        - [information] Extends existing namespace
    http.request.method = GET
    Span event test_event
        hello = world
            - [violation] Attribute `hello` does not exist in the registry.
            - [improvement] Does not have a namespace
    Span link
        hello = world
            - [violation] Attribute `hello` does not exist in the registry.
            - [improvement] Does not have a namespace

Samples
  - total: 14  - by type:
    - attribute: 10
    - resource: 1
    - span: 1
    - span_event: 1
    - span_link: 1
  - by highest advice level:
    - no advice: 6
    - improvement: 1
    - violation: 7

Advisories given
  - total: 15  - advice level:
    - improvement: 4
    - information: 2
    - violation: 9
  - advice type:
    - extends_namespace: 2
    - illegal_namespace: 1
    - invalid_format: 1
    - missing_attribute: 6
    - missing_namespace: 3
    - not_stable: 1
    - type_mismatch: 1

Registry coverage
  - entities seen: 0.36%

✔ Performed live check for registry `https://github.com/open-telemetry/semantic-conventions.git[model]`

Total execution time: 1.924507458s
```

crates/weaver_checker/src/lib.rs
```diff
@@ -106,7 +106,7 @@ pub enum Error {
         /// The provenance of the violation (URL or path).
         provenance: String,
         /// The violation.
-        violation: Violation,
+        violation: Box<Violation>,
```
jerbly, line 109: Why does this need to be Boxed now?

crates/weaver_checker/src/violation.rs
```diff
@@ -87,12 +89,27 @@ pub enum AdviceLevel {
...
-    /// The type of advice e.g. "is_deprecated"
+    /// The type of advice e.g. "is_deprecated". This should be a short,
+    /// machine-readable string that categorizes the advice.
     pub advice_type: String,
-    /// The value of the advice e.g. "true"
+
+    /// The context associated with the advice e.g. { "attribute_name": "foo.bar", "attribute_value": "bar" }
+    /// The value should contain all dynamic parts of the message in a structured way.
+    /// Context values may be used with custom templates and filters to customize reports.
     pub value: Value,
```
jerbly: This was originally called Value to represent a single value but with any type. Now it can be a complex type representing the context of the advice. So should we rename it to `advice_context`?

crates/weaver_live_check/src/advice.rs
```diff
@@ -68,33 +69,54 @@ impl Advisor for DeprecatedAdvisor {
...
                 if let Some(attribute) = registry_attribute {
                     if let Some(deprecated) = &attribute.deprecated {
                         advices.push(Advice {
                             advice_type: "deprecated".to_owned(),
-                            value: deprecated_to_value(deprecated),
-                            message: deprecated.to_string(),
+                            value: json!({
+                                "attribute_name": sample_attribute.name.clone(),
+                                "deprecation_reason": deprecated_to_reason(deprecated),
+                                "deprecation_note": deprecated.to_string(),
```
jerbly: These keys should be constants so we can ensure consistency. (This comment is to cover all occurrences through this module for value construction).

crates/weaver_live_check/src/advice.rs
```diff
@@ -68,33 +69,54 @@ impl Advisor for DeprecatedAdvisor {
...
                         advices.push(Advice {
                             advice_type: "deprecated".to_owned(),
-                            value: deprecated_to_value(deprecated),
-                            message: deprecated.to_string(),
+                            value: json!({
+                                "deprecation_reason": deprecated_to_reason(deprecated),
+                                "deprecation_note": deprecated,
+                            }),
+                            message: format!(
+                                "Metric is deprecated; reason = {}, note = {}",
```
jerbly, line 118: For Attributes we're repeating the name in the message and context: `"Attribute '{}' is deprecated; reason = '{}', note = '{}'.",` should we do the same for Metric names, or remove the repeat for Attribute?

crates/weaver_live_check/src/live_checker.rs
```diff
@@ -915,9 +998,10 @@ mod tests {
 
         let mut stats = LiveCheckStatistics::new(&live_checker.registry);
         for sample in &mut samples {
-            // This should fail with: "error: use of undefined variable `attribu1te_name` is unsafe"
+            // This should fail with: "error: use of undefined variable `attribute_name` is unsafe"
```
jerbly: this misspelling of attribute_name is intentional - it's the purpose of the test. See `bad.rego` file.

src/util.rs
```diff
@@ -121,7 +121,7 @@ pub(crate) fn check_policy_stage<T: Serialize, U: Serialize>(
                 for violation in violations {
                     errors.push(PolicyViolation {
                         provenance: policy_file.to_owned(),
-                        violation,
+                        violation: Box::new(violation),
```
jerbly, line 124: Why is this boxed?

tests/registry_emit.rs
```diff
@@ -83,7 +83,7 @@ fn test_emit_with_live_check() {
     assert_eq!(no_advice_count, 37);
     assert_eq!(total_advisories, 13);
     assert_eq!(total_entities, 50);
-    assert!(registry_coverage > 0.7);
+    assert!(registry_coverage > 0.5);
```
jerbly: Why is the coverage reduced in the test? The model hasn't changed in `crates/weaver_emit/data` in this PR - so does this indicate a bug?

crates/weaver_live_check/src/advice.rs
lmolkova: we have universal signal_name and signal_type as top-level properties.

message identifies error within the signal and provides human readable explanation. The full message could look like  `{signal_type} with name {signal_name} has advice with level {advice_level} and message {message}`

think about it as a log record. No need to mention file name in log message body, only mention what's specific to this record.

crates/weaver_checker/src/lib.rs
lmolkova: because Advice got new members, Violation got bigger, and PolicyViolation is used in error and apparently there is a clippy check for error values.

```log

error: the `Err`-variant returned from this function is very large
   --> crates/weaver_checker/src/lib.rs:417:52
    |
105 | /     PolicyViolation {
106 | |         /// The provenance of the violation (URL or path).
107 | |         provenance: String,
108 | |         /// The violation.
109 | |         violation: Violation,
110 | |     },
    | |_____- the largest variant contains at least 160 bytes
...
417 |       pub fn check(&mut self, stage: PolicyStage) -> Result<Vec<Violation>, Error> {
    |                                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = help: try reducing the size of `Error`, for example by boxing large elements or replacing it with `Box<Error>`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#result_large_err
```

Please let me know if there is a better way to fix it.

crates/weaver_live_check/src/advice.rs
lmolkova: done

crates/weaver_checker/src/violation.rs
lmolkova: renamed to advice_context

src/util.rs
lmolkova: same as https://github.com/open-telemetry/weaver/pull/943#discussion_r2380285736, let's continue the discussion there

jerbly:
Looks good. I think we just need to update the crate README: https://github.com/open-telemetry/weaver/tree/main/crates/weaver_live_check#readme

And, we need a breaking change notice in the changelog.

lmolkova:
thanks @jerbly, I updated readme and added changelog.

crates/weaver_live_check/README.md
```diff
@@ -85,8 +85,12 @@ As mentioned, a list of `Advice` is returned in the report for each sample entit
 
 - `advice_level`: _string_ - one of `violation`, `improvement` or `information` with that order of precedence. Weaver will return with a non-zero exit-code if there is any `violation` in the report.
 - `advice_type`: _string_ - a simple machine readable string to represent the advice type
-- `message`: _string_ - a verbose string describing the advice
-- `value`: _any_ - a pertinent entity associated with the advice
+- `signal_type`: _string_ - a type of the signal advice is reported for: `metirc`, `span`, or `resource`
```
jerbly:
```suggestion
- `signal_type`: _string_ - a type of the signal advice is reported for: `metric`, `span`, or `resource`
```

crates/weaver_live_check/README.md
```diff
@@ -85,8 +85,12 @@ As mentioned, a list of `Advice` is returned in the report for each sample entit
...
+- `signal_name`: _string_ - a name of the signal addvice is reported for: metric name or span name
```
jerbly:
```suggestion
- `signal_name`: _string_ - a name of the signal advice is reported for: metric name or span name
```

crates/weaver_live_check/README.md
jerbly: This rego policy needs updating for `advice_context`

jerbly (approved):
Looks good. Excited to give it a try! Thanks!

Issue #689, opened 2025-04-09
Make the intermediate registry directory optional within templates


lquerel:
This issue proposes making the intermediate registry directory, currently required inside the templates folder pointed to by the `--templates` parameter, optional or removable.

Currently, we have:

```
my_templates/   # Directory provided to --templates parameter
  registry/     # Intermediate directory we propose making optional or removable
    go/
    rust/
    ...
```

Initially, the `registry` directory matched the first-level Weaver command. For example:

`weaver registry generate rust --templates my_templates`

This command searches for templates in `my_templates/registry/rust`. While logical, after several months, this intermediate directory has proven confusing and unnecessary for users.

Instead, users should have flexibility over their template structure. Thus, rather than:

`weaver registry generate rust --templates my_templates`

Users could simply write:

`weaver registry generate rust --templates my_templates/registry`

Or, even simpler, eliminate this intermediate level entirely:

`weaver registry generate rust --templates my_templates`

Another simpler option would be eliminating the target (rust) entirely from the command, allowing the `--templates` parameter to directly specify the template directory for Rust code generation. However, this simpler option conflicts slightly with another usability enhancement we have in mind: bundling the most common templates directly within Weaver’s binary distribution. Ideally, we would like users to be able to type:

`weaver registry generate rust`

without specifying `--templates`, allowing Weaver to use built-in templates by default. Of course, specifying custom targets and template directories would remain possible for advanced users.

In my opinion, the decision boils down to one of the following options:
- Do nothing (keep current structure)
- Make the intermediate directory optional
- Completely remove the intermediate directory and handle migration for users (breaking change)

There might be other approaches worth considering that aren’t listed here. Feel free to suggest additional alternatives in the comments.

jerbly:
> However, this simpler option conflicts slightly with another usability enhancement we have in mind: bundling the most common templates directly within Weaver’s binary distribution.

This UX concept extends beyond templates to all other builtins. In the live-check PR we're discussing how to select builtin rego advisors versus user-defined ones. We need a consistent approach to selecting a builtin versus providing your own.

Perhaps this could fit neatly with the `VirtualDirectory` pattern that you've just introduced? Could there be some syntax that specifies a builtin. Off the top of my head:

`weaver registry generate --templates builtin::rust`

PR #962, opened 2025-09-26
Make optional the target and the registry sub-directory for the weaver registry generate command.

lquerel:
Closes https://github.com/open-telemetry/weaver/issues/689

This PR aims to simplify the user experience of Weaver. Initially, Weaver required following a strict structure for templates (i.e. <template-dir>/registry/<target>/). With this PR, the registry subdirectory and the target are now optional.

For example, if you have your templates and the `weaver.yaml` file in the `my-templates` directory, the following command now allows you to generate the templates:

```
weaver registry generate --templates my-templates
```

In the same way, suppose your templates are in the `my-templates/registry/rust` directory (legacy style), the following 2 commands are equivalent.

```
weaver registry generate rust --templates my-templates
weaver registry generate --templates my-templates/registry/rust
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 All notable changes to this project will be documented in this file.
 
+# [TBD] - TDB
+
+- The target and the `registry` sub-directory are now optional for the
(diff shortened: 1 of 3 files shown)


jerbly (approved):
LGTM

Issue #430, opened 2024-10-24
Support passing github credentials if remote repository requires authentication


awangc:
**Is your feature request related to a problem? Please describe.**
I'm trying to use weaver's `-r` option to point to a repository that is hosted on github.com but via an enterprise account requiring authentication. When generating artifacts it prompts for username and password, which prevents the tool from being used in automation pipeline directly (needs to wrap it around something like `tmux`) 

**Describe the solution you'd like**
a way to pass credentials (username and password) so repositories can be accessed: specific environment variables or CLI args would be good

**Describe alternatives you've considered**
Wrapping the weaver command around `tmux` works. 
```
tmux new-session -d -s "weaver" "weaver registry generate -r https://github.com/<path-to-repository> ..."
tmux send-key -t weaver: $GHEC_USER "Enter"
tmux send-key -t weaver: $GHEC_PASS "Enter"
```
But this feels an extra step that could be removed.

PR #871, opened 2025-08-07
Proposal: Introduce bundles

thompson-tomo:
Progresses #857

This proposal seeks to introduce the concept of bundles which can be used when authoring definitions and provide extensability for other standards.

PR #978, opened 2025-10-10
feat: authenticating when using remote archive

matoous:
> [!NOTE]  
> Work in progress PR, opening early to collect feedback.

Add `auth-token` option to the CLI that allows configuring authentication when fetching registry from a remote archive.

## Notes

`auth-token` is passed and handled separately from the virtual directory path string as the usual `@` character separating userinfo portion of the URL from the host is reserved for the `refspec`. Taking the token from the userinfo portion of the URL would be an alternative to the approach in this commit but would make the parsing a little bit more cumbersome. I am open to both approach and happy to adjust the code to whatever the OpenTelemetry team deems better.

## Open for discussion

- I am not sure what the real-world distribution of hosts of remote registries will be and if we can safely assume `Bearer` scheme in the `Authorization` header. For example, most guides using GitHub PATs use `token` scheme (even though `Bearer` works too). Using `Bearer` might be an OK compromise for the initial version.
- Registries can rely on further dependencies which in turn might again be private. In such scenario option for an approach with the token being part of the URL might not work as one wouldn't want to put their access token into the registry manifest. In such the proposed approach in this PR would work as long as all dependencies are hosted by the same remote and can be access with the same token. This might become irrelevant with e.g. https://github.com/open-telemetry/weaver/pull/871 or similar bundling approach.

---

Fixes: https://github.com/open-telemetry/weaver/issues/430

### DIFF
--- a/crates/weaver_common/src/test.rs
+++ b/crates/weaver_common/src/test.rs
@@ -5,7 +5,7 @@
 use paris::error;
 use std::path::PathBuf;
 
-use rouille::{match_assets, Server};
+use rouille::{match_assets, Response, Server};
 use std::sync::mpsc::Sender;
 
 /// An error that can occur while starting the HTTP server.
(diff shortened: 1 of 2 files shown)


jsuereth:
Apologies for taking a while to get to this.  My thoughts are:

- Whatever auth we support, ideally, we'd keep it similar to `curl` from a command line perspective.
- If we wanted deeper configuration, that should be done via `~/weaver.yml` or `~/.weaver/weaver.yml`.
- We can evolve our support over time based on need.  Priority wise I think:
  - Supporting Github resolution should be top priority
  - General purpose (but possibly awkward bearer-token) should ALSO exist for fallback / other cases.
  - Further auth models we can evaluate on an as-needed basis.

@lquerel is best to review any configuration options / auth model here for multi-registry support.

PR #980, opened 2025-10-10
V2 Resolved Schema

jsuereth:
Ready for review

This modifies the resolved telemetry schema (and template values) to look like the V2 definition schema.

With this PR you can resolve into the new schema via a new flag to check how it looks, e.g.

```
weaver registry resolve --v2 > test.yaml
```

Here's a sample of the new layout:

```yaml
attributes:
  - key: android.app.state
    type:
      ...
signals:
  metrics:
  - name: aspnetcore.authentication.authenticate.duration
    instrument: histogram
    unit: s
    ...
  spans:
  - type: aws.client
    ...
  events:
  - name: app.jank
    ...
  entities:
  - type: android
    ...
refinements:
  metrics:
  - id: ...refinement id...
    ...full metric definition...
  spans:
    - id: ...refinement id...
    ...full span definition...
  events:
    - id: ...refinement id...
    ...full event definition...
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,12 +67,12 @@ Weaver provides a _set of tools_ for working with **schematized telemetry**.
 
 Further reading:
 - [Weaver Architecture](docs/architecture.md): A document detailing the architecture of the project.
-- [Application Telemetry Schema OTEP](https://github.com/open-telemetry/oteps/blob/main/text/0243-app-telemetry-schema-vision-roadmap.md): A vision and roadmap for the concept of Application Telemetry Schema.
+- [Application Telemetry Schema OTEP](https://github.com/open-telemetry/opentelemetry-specification/blob/main/oteps/0243-app-telemetry-schema-vision-roadmap.md): A vision and roadmap for the concept of Application Telemetry Schema.
 
 ## Examples and How-Tos
 
+- [Weaver Examples](https://github.com/open-telemetry/opentelemetry-weaver-examples) - Official OTel project showcasing weaver use cases through runnable examples.
 - [O11y by design](https://github.com/jsuereth/o11y-by-design/) - from the CNCF 2025 presentation
-- [Weaver Example](https://github.com/jerbly/weaver-example) - doc-gen, code-gen, emit, live-check in CI/CD
 - [Define your own telemetry schema](docs/define-your-own-telemetry-schema.md) - A guide on how to define your own
   telemetry schema using semantic conventions.
 


crates/weaver_forge/src/lib.rs
```diff
@@ -956,6 +957,19 @@ mod tests {
             .expect("Failed to generate registry assets");
 
         assert!(diff_dir("expected_output/test", "observed_output/test").unwrap());
+
+        // TODO - Remove this.
```
lmolkova: is it ok to remove now?

crates/weaver_forge/src/v2/metric.rs
```diff
@@ -0,0 +1,81 @@
...
+    #[serde(default)]
+    #[serde(skip_serializing_if = "Vec::is_empty")]
+    pub entity_associations: Vec<String>,
+
+    /// Common fields (like brief, note, annotations).
+    #[serde(flatten)]
+    pub common: CommonFields,
+}
+
+/// A special type of reference to attributes that remembers metric-specicific information.
```
lmolkova:
```suggestion
/// A special type of reference to attributes that remembers metric-specific information.
```

crates/weaver_forge/src/v2/registry.rs
```diff
@@ -0,0 +1,380 @@
...
+/// of codegeneration for MySQL.
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, JsonSchema)]
+#[serde(deny_unknown_fields)]
+pub struct Refinements {
+    /// The metric refinements defined.
+    pub metrics: Vec<MetricRefinement>,
+    /// The span refinements defined.
+    pub spans: Vec<SpanRefinement>,
+    /// The event refinements defined.
+    pub events: Vec<EventRefinement>,
```
lmolkova, line 67: future: we should probably have entity refinement at some point (linux vs windows process, etc)

crates/weaver_forge/src/v2/registry.rs
```diff
@@ -0,0 +1,380 @@
...
+    },
+};
+
+/// A resolved semantic convention registry used in the context of the template and policy
+/// engines.
+///
+/// This includes all registrys fully fleshed out and ready for codegen.
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, JsonSchema)]
+#[serde(deny_unknown_fields)]
+pub struct ResolvedRegistry {
```
lmolkova: we should have some place for public attribute groups here

crates/weaver_forge/src/v2/registry.rs
```diff
@@ -0,0 +1,380 @@
...
+                    }
+                    attr
+                })
+                .collect();
+            metrics.push(Metric {
+                name: metric.name,
+                instrument: metric.instrument,
+                unit: metric.unit,
+                attributes,
+                entity_associations: metric.entity_associations,
```
lmolkova, line 130: not blocking, we can follow-up: we're checking that attributes exist, but not checking it for entities, it seems inconsistent.

crates/weaver_forge/src/v2/registry.rs
```diff
@@ -0,0 +1,380 @@
...
+                            key: a.key.clone(),
+                            r#type: a.r#type.clone(),
+                            examples: a.examples.clone(),
+                            common: a.common.clone(),
+                        },
+                        requirement_level: ar.requirement_level.clone(),
+                    });
+                    if attr.is_none() {
+                        errors.push(Error::AttributeNotFound {
+                            group_id: format!("event.{}", &event.id),
```
lmolkova, line 302: shouldn't we use refinement id here (and in other places where we deal with refinement)?

```suggestion
                            group_id: event.id,
```

crates/weaver_forge/src/v2/span.rs
```diff
@@ -0,0 +1,81 @@
...
+    ///
+    /// Note: For attributes that are "recommended" or "opt-in" - not all metric source will
+    /// create timeseries with these attributes, but for any given timeseries instance, the attributes that *were* present
+    /// should *remain* present. That is - a metric timeseries cannot drop attributes during its lifetime.
+    pub requirement_level: RequirementLevel,
+
+    /// Specifies if the attribute is (especially) relevant for sampling
+    /// and thus should be set at span start. It defaults to false.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub sampling_relevant: Option<bool>,
```
lmolkova, line 57: there were some discussions on 'sampling-relevant' being non-intuitive name. Maybe we should rename it in v2 to something else? (We can also follow-up since this would go beyond the scope of this PR).

Copy-pasting from past meeting notes:

> Rename to something like ‘head-sampling’  / ‘provided - at span start time’ 
"start span operation" <- specification term.

E.g. `record_on_start`, `provided_on_start`, `available_at_start_time`, etc

crates/weaver_resolved_schema/src/lineage.rs
```diff
@@ -38,6 +38,11 @@ pub struct GroupLineage {
     /// The provenance of the source file where the group is defined.
     provenance: Provenance,
 
+    /// The group that this group extended, if available.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[serde(default)]
+    pub extends_group: Option<String>,
```
lmolkova, line 44: do you envision it being used after we migrate to v2? I assume it's a temporary measure until we redesign lineage.

crates/weaver_resolved_schema/src/v2/event.rs
```diff
@@ -0,0 +1,69 @@
...
+    /// The name of the event.
+    pub name: SignalId,
+
+    /// List of attributes that belong to this event.
+    #[serde(default)]
+    #[serde(skip_serializing_if = "Vec::is_empty")]
+    pub attributes: Vec<EventAttributeRef>,
+
+    // TODO - Should Entity Associations be "strong" links?
+    /// Which resources this span should be associated with.
```
lmolkova: (and for other signals)

```suggestion
    /// Which entities this event should be associated with.
```

crates/weaver_resolved_schema/src/v2/event.rs
```diff
@@ -0,0 +1,69 @@
...
+pub struct Event {
+    /// The name of the event.
+    pub name: SignalId,
+
+    /// List of attributes that belong to this event.
+    #[serde(default)]
+    #[serde(skip_serializing_if = "Vec::is_empty")]
+    pub attributes: Vec<EventAttributeRef>,
+
+    // TODO - Should Entity Associations be "strong" links?
```
lmolkova, line 24: what does `strong` mean here? I think it'd be consistent with attributes if we had something like `EntityAssociation` and resolved entities similar to attributes. It can be a follow-up though

crates/weaver_resolved_schema/src/v2/event.rs
```diff
@@ -0,0 +1,69 @@
...
+    /// Which resources this span should be associated with.
+    ///
+    /// This list is an "any of" list, where a metric may be associated with one or more entities, but should
+    /// be associated with at least one in this list.
```
lmolkova: I thought about entity associations as all-of rather than any-of - perhaps we should document whatever it is somewhere in the semconv repo and come back here to update docs?

crates/weaver_resolved_schema/src/v2/event.rs
```diff
@@ -0,0 +1,69 @@
...
+/// e.g. for HTTP events, there may be a refinement that provides only the necessary information for dealing with Java's HTTP
+/// client library, and drops optional or extraneous information from the underlying http event.
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, JsonSchema)]
+pub struct EventRefinement {
+    /// The identity of the refinement
+    pub id: SignalId,
+
+    // TODO - This is a lazy way of doing this.  We use `name` to refer
+    // to the underlying span defintiion, but override all fields here.
+    // We probably should copy-paste all the "span" attributes here
```
lmolkova:
```suggestion
    // We probably should copy-paste all the "event" attributes here
```

src/registry/resolve.rs
```diff
@@ -39,6 +39,11 @@ pub struct RegistryResolveArgs {
     #[arg(short, long, default_value = "yaml")]
     format: Format,
 
+    // TODO - Figure out long term plan for verisons here.
```
lmolkova: we can prob always emit v2 with `--future` flag but allow to downgrade to v1. I.e. 
- `resolve` would do v1
- `resolve --future` would do v2
- `resolve --future --v1` or `resolve --future --v2=false` would do v1 - we might need it to simplify semconv onboarding, but it does not seem essential

lmolkova (commented):
left some comments, but looks great overall!

crates/weaver_resolved_schema/src/v2/event.rs
```diff
@@ -0,0 +1,69 @@
...
+    pub attributes: Vec<EventAttributeRef>,
+
+    // TODO - Should Entity Associations be "strong" links?
+    /// Which resources this span should be associated with.
+    ///
+    /// This list is an "any of" list, where a metric may be associated with one or more entities, but should
+    /// be associated with at least one in this list.
+    #[serde(default)]
+    #[serde(skip_serializing_if = "Vec::is_empty")]
+    pub entity_associations: Vec<String>,
```
thompson-tomo, line 31: Could we either instead or in addition, add the attributes from the entities to the attribute list along with 2 additional properties:

- `capture.scope` with a value of `resource`
- `capture.group` with the name of the entity as the value.

This way in documentation we have a complete list of the attributes and reader knows what resource detectors are suggested and they gain. This would also flow to instrumentation scope attributes when we can define/associate them to signals. This same approach also works nicely for public attribute groups. For the later 2 scenarios, code gen could generate methods for setting the attributes.

This would apply to spans & events. 

For metrics perhaps we could either include just the identifying attributes from the entity or leave them off for the time being.

crates/weaver_resolved_schema/src/v2/registry.rs
```diff
@@ -0,0 +1,57 @@
...
+///   no storage location in OTLP.
+///
+/// Note: Refinements will always include a "base" refinement for every signal definition.
+///       For example, if a Metric signal named `my_metric` is defined, there will be
+///       a metric refinement named `my_metric` as well.
+///       This allows codegen to *only* interact with refinements, if desired, to
+///       provide optimised methods for generating telemetry signals.
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, JsonSchema)]
+#[serde(deny_unknown_fields)]
+pub struct Registry {
```
thompson-tomo, line 23: It would be good to see public attribute groups included here.

crates/weaver_resolved_schema/src/v2/registry.rs
```diff
@@ -0,0 +1,57 @@
...
+    pub entities: Vec<Entity>,
+
+    /// A  list of span refinements.
+    pub span_refinements: Vec<SpanRefinement>,
+
+    /// A  list of metric refinements.
+    pub metric_refinements: Vec<MetricRefinement>,
+
+    /// A  list of event refinements.
+    pub event_refinements: Vec<EventRefinement>,
```
thompson-tomo: I think it might be beneficial to group the definitions based on the base namespace ie db then at the top level we just have an array of namespaces. This concept and the use case for namespaces is described in more detail in https://github.com/open-telemetry/weaver/pull/867/files but I would start off with just the name + signals being the only properties of the namespace as the rest could be added via non-breaking changes.

crates/weaver_forge/src/lib.rs
jsuereth: I'm still using this to test/demo.  I was going to remove after we lock-in on the details of the schema.

crates/weaver_forge/src/v2/registry.rs
jsuereth: I haven't sync'd to master since starting this. I'll add those in next.

crates/weaver_forge/src/v2/registry.rs
jsuereth: I believe this is the refinement id?

Do you mean the string should be different?

crates/weaver_forge/src/v2/span.rs
jsuereth: Yeah - I think we need to make some decisions here.  I was looking in the past, and "sampling relevant" was still marked "experimental" on our old schema, but we somehow dropped that notion as we evolved into weaver.

crates/weaver_resolved_schema/src/lineage.rs
jsuereth: 100% temporary.  I haven't (yet) created Lineage for V2.

crates/weaver_resolved_schema/src/v2/event.rs
jsuereth: I was thinking of moving `entities` into the Catalog (next to attributes) and have this be a "ref" instead of a pure string.

I was actually thinking about doing that across the board:

- Catalog:  Pure lists of `Attribute`, `Event`, `Metric`, `Entity`, `Span`, `AttributeGroup`.
- Registry: All refinements of the above.

crates/weaver_resolved_schema/src/v2/event.rs
jsuereth: It's absolutely documented that way now - both in the modelling guide and in the JSON schema / weaver comments.

crates/weaver_resolved_schema/src/v2/event.rs
lquerel: @jsuereth I think that's indeed a good approach.

crates/weaver_forge/src/v2/entity.rs
```diff
@@ -0,0 +1,51 @@
+//! Event related definitions structs.
```
lquerel, line 1:
```suggestion
//! Entity related definitions structs.
```

lquerel, line 1:
```suggestion
//! Entity related definitions structs.
```

crates/weaver_forge/src/v2/entity.rs
```diff
@@ -0,0 +1,51 @@
...
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+use weaver_semconv::{
+    attribute::RequirementLevel,
+    v2::{signal_id::SignalId, CommonFields},
+};
+
+use crate::v2::attribute::Attribute;
+
+/// The definition of an event signal.
```
lquerel:
```suggestion
/// The definition of an entity.
```

crates/weaver_forge/src/v2/entity.rs
```diff
@@ -0,0 +1,51 @@
...
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, JsonSchema)]
+#[serde(deny_unknown_fields)]
+pub struct Entity {
+    /// The type of the entity.
+    pub r#type: SignalId,
+
+    /// List of attributes that belong to this event.
```
lquerel: The comment would benefit from being updated to match the field below :-)

crates/weaver_forge/src/v2/entity.rs
```diff
@@ -0,0 +1,51 @@
...
+    #[serde(default)]
+    #[serde(skip_serializing_if = "Vec::is_empty")]
+    pub identity: Vec<EntityAttribute>,
+
+    /// List of attributes that belong to this event.
```
lquerel: The comment would benefit from being updated to match the field below :-)

crates/weaver_forge/src/v2/entity.rs
```diff
@@ -0,0 +1,51 @@
...
+    #[serde(default)]
+    #[serde(skip_serializing_if = "Vec::is_empty")]
+    pub description: Vec<EntityAttribute>,
+
+    /// Common fields (like brief, note, annotations).
+    #[serde(flatten)]
+    pub common: CommonFields,
+}
+
+/// A special type of reference to attributes that remembers event-specicific information.
```
lquerel:
```suggestion
/// A special type of reference to attributes that remembers entity-specicific information.
```

crates/weaver_forge/src/v2/event.rs
```diff
@@ -0,0 +1,73 @@
...
+pub struct Event {
+    /// The name of the event.
+    pub name: SignalId,
+
+    /// List of attributes that belong to this event.
+    #[serde(default)]
+    #[serde(skip_serializing_if = "Vec::is_empty")]
+    pub attributes: Vec<EventAttribute>,
+
+    /// Which resources this span should be associated with.
```
lquerel:
```suggestion
    /// Which resources this event should be associated with.
```

crates/weaver_forge/src/v2/event.rs
```diff
@@ -0,0 +1,73 @@
...
+    ///
+    /// This list is an "any of" list, where a metric may be associated with one or more entities, but should
```
lquerel: metric ?

crates/weaver_forge/src/v2/event.rs
```diff
@@ -0,0 +1,73 @@
...
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, JsonSchema)]
+pub struct EventRefinement {
+    /// The identity of the refinement.
+    pub id: SignalId,
+
+    // TODO - This is a lazy way of doing this.  We use `type` to refer
+    // to the underlying span defintiion, but override all fields here.
+    // We probably should copy-paste all the "span" attributes here
+    // including the `ty`
+    /// The definition of the metric refinement.
```
lquerel: This comment needs some massage :-)

crates/weaver_forge/src/v2/metric.rs
```diff
@@ -0,0 +1,81 @@
...
+    pub instrument: InstrumentSpec,
+    /// The unit in which the metric is measured, which should adhere to the
+    /// [guidelines](https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/metrics/semantic_conventions#instrument-units).
+    pub unit: String,
+    /// List of attributes that should be included on this metric.
+    #[serde(default)]
+    #[serde(skip_serializing_if = "Vec::is_empty")]
+    pub attributes: Vec<MetricAttribute>,
+    // TODO - Should Entity Associations be "strong" links?
+    /// Which resources this span should be associated with.
```
lquerel:
```suggestion
    /// Which resources this metric should be associated with.
```

crates/weaver_forge/src/v2/metric.rs
```diff
@@ -0,0 +1,81 @@
...
+    /// Common fields (like brief, note, annotations).
+    #[serde(flatten)]
+    pub common: CommonFields,
+}
+
+/// A special type of reference to attributes that remembers metric-specicific information.
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, JsonSchema)]
+#[serde(deny_unknown_fields)]
+pub struct MetricAttribute {
+    /// Base metric definitions.
```
lquerel:
```suggestion
    /// Base attribute definitions.
```

crates/weaver_forge/src/v2/metric.rs
```diff
@@ -0,0 +1,81 @@
...
+/// A refinement represents a "view" of a Metric that is highly optimised for a particular implementation.
+/// e.g. for HTTP metrics, there may be a refinement that provides only the necessary information for dealing with Java's HTTP
+/// client library, and drops optional or extraneous information from the underlying http metric.
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, JsonSchema)]
+pub struct MetricRefinement {
+    /// The identity of the refinement.
+    pub id: SignalId,
+
+    // TODO - This is a lazy way of doing this.  We use `type` to refer
+    // to the underlying span defintiion, but override all fields here.
```
lquerel:
```suggestion
    // to the underlying metric definition, but override all fields here.
```

crates/weaver_forge/src/v2/metric.rs
```diff
@@ -0,0 +1,81 @@
...
+    // We probably should copy-paste all the "span" attributes here
+    // including the `ty`
```
lquerel: I'm confused with this comment

crates/weaver_forge/src/v2/registry.rs
```diff
@@ -0,0 +1,380 @@
...
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, JsonSchema)]
+#[serde(deny_unknown_fields)]
+pub struct Signals {
+    /// The metric signals defined.
+    pub metrics: Vec<Metric>,
+    /// The span signals defined.
+    pub spans: Vec<Span>,
+    /// The event signals defined.
+    pub events: Vec<Event>,
+    /// The entity signals defined.
```
lquerel, line 50: In the OTel model, are entities considered signals or things attached to signals? I thought it was the latter, but I probably haven't kept up with the latest iterations on this topic. If it's not a signal, then I think we should place them at the higher level.

crates/weaver_forge/src/v2/span.rs
```diff
@@ -0,0 +1,81 @@
...
+/// A refinement represents a "view" of a Span that is highly optimised for a particular implementation.
+/// e.g. for HTTP spans, there may be a refinement that provides only the necessary information for dealing with Java's HTTP
+/// client library, and drops optional or extraneous information from the underlying http span.
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, JsonSchema)]
+pub struct SpanRefinement {
+    /// The identity of the refinement.
+    pub id: SignalId,
+
+    // TODO - This is a lazy way of doing this.  We use `type` to refer
+    // to the underlying span defintiion, but override all fields here.
```
lquerel:
```suggestion
    // to the underlying span definition, but override all fields here.
```

crates/weaver_resolved_schema/src/v2/catalog.rs
```diff
@@ -0,0 +1,69 @@
...
+            && a.r#type == attribute.r#type
+            && a.examples == attribute.examples
+            && a.common.brief == attribute.brief
+            && a.common.note == attribute.note
+            && a.common.deprecated == attribute.deprecated
+                }, // && a.common.stability == attribute.stability
+                   // && a.common.annotations == attribute.annotations
+            )
+            .map(|idx| AttributeRef(idx as u32))
+    }
```
lquerel, line 75: I think this choice is justified for the kind of volume we have to handle for now. Perhaps one day (I hope so) we'll have catalogs with hundreds of thousands of entries, at which point we'll need to switch from this O(N) approach to a hashmap, for example, at the cost of increased memory usage.
A short comment on the rationale could be helpful for future reviewer/coder.

crates/weaver_resolved_schema/src/v2/entity.rs
```diff
@@ -0,0 +1,43 @@
+//! Event related definition structs.
```
lquerel:
```suggestion
//! Entity related definition structs.
```

crates/weaver_resolved_schema/src/v2/entity.rs
```diff
@@ -0,0 +1,43 @@
...
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+use weaver_semconv::{
+    attribute::RequirementLevel,
+    v2::{signal_id::SignalId, CommonFields},
+};
+
+use crate::v2::attribute::AttributeRef;
+
+/// The definition of an Entity signal.
```
lquerel, line 12: Signal?

crates/weaver_resolved_schema/src/v2/catalog.rs
```diff
@@ -0,0 +1,69 @@
...
+                |a| {
+                    a.key == attribute.name
+            // TODO check everything
+            && a.r#type == attribute.r#type
+            && a.examples == attribute.examples
+            && a.common.brief == attribute.brief
+            && a.common.note == attribute.note
+            && a.common.deprecated == attribute.deprecated
+                }, // && a.common.stability == attribute.stability
+                   // && a.common.annotations == attribute.annotations
```
lquerel: What's the plan regarding these two fields? Annotations are probably something to consider for this comparison. I think at a minimum there should be a TODO here.

crates/weaver_resolved_schema/src/v2/mod.rs
```diff
@@ -0,0 +1,604 @@
...
+    fix_group_id("span.", group_id)
+}
+
+/// Converts a V1 registry + catalog to V2.
+pub fn convert_v1_to_v2(
+    c: crate::catalog::Catalog,
+    r: crate::registry::Registry,
+) -> Result<(Catalog, Registry), crate::error::Error> {
+    // When pulling attributes, as we collapse things, we need to filter
+    // to just unique.
```
lquerel, line 88: I don't understand this comment unless we consider that equality between two attributes doesn't follow the same rules between v1 and v2. Is that the case? And if so, what's the rationale?

crates/weaver_resolved_schema/src/v2/mod.rs
```diff
@@ -0,0 +1,604 @@
...
+                for attr in g.attributes.iter().filter_map(|a| c.attribute(a)) {
+                    if let Some(a) = v2_catalog.convert_ref(attr) {
+                        span_attributes.push(span::SpanAttributeRef {
+                            base: a,
+                            requirement_level: attr.requirement_level.clone(),
+                            sampling_relevant: attr.sampling_relevant.clone(),
+                        });
+                    } else {
+                        // TODO logic error!
+                    }
```
lquerel, line 151: Is this dance, attribute-ref -> attribute -> attribute-ref, necessary because the comparison rules for attributes differ between v1 and v2?
That brings us back to my previous question about the rationale behind this difference. In any case, once we've fully migrated to v2, we shouldn't have this kind of thing anymore.

src/registry/resolve.rs
lquerel: To do that, we'll need to coordinate closely with the changes required for the policies and templates in the semconv.

crates/weaver_resolved_schema/src/v2/event.rs
```diff
@@ -0,0 +1,69 @@
...
+#[serde(deny_unknown_fields)]
+pub struct EventAttributeRef {
+    /// Reference, by index, to the attribute catalog.
+    pub base: AttributeRef,
+    /// Specifies if the attribute is mandatory. Can be "required",
+    /// "conditionally_required", "recommended" or "opt_in". When omitted,
+    /// the attribute is "recommended". When set to
+    /// "conditionally_required", the string provided as <condition> MUST
+    /// specify the conditions under which the attribute is required.
+    pub requirement_level: RequirementLevel,
```
lquerel, line 49: I admit I'm not entirely clear on why `requirement_level` gets special treatment here. In what way is `requirement_level` more event-specific than `note`, `examples`, ..., in the context of a refinement?

lquerel (commented):
Wow, it took me quite some time to go through all of this. Thanks for the massive refactoring of the semconv schema and everything that comes with it!

Sorry for the number of comments. A lot of them are just copy/paste issues or typos. There are probably only a handful that really concern the approach or the decisions made. Apologies in advance if some of these points or answers have already been discussed in the SIG meetings; I haven't been very present in those lately.

crates/weaver_resolved_schema/src/v2/event.rs
jsuereth: I'm thinking of this via the "attribute registry" we have in otel.

Requirement level isn't an independent concept for any individual attribute, while examples, note *is*.
Requirement level *only* makes sense in the context of a specific signal (e.g. this metric *requires* this attribute).  Additionally, requirement level is almost guaranteed to diverge between Metrics<->Spans/Events in some critical ways, due to cardinality constraints between the signals.

If you look at the existing semconv registries (e.g. [attributes](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/registry/attributes/README.md)), you'll see we don't include requirement_level.

I feel pretty strongly we should NOT include requirement_level in either the attribute catalog/registry *or* outside of a signal context.

crates/weaver_resolved_schema/src/v2/mod.rs
jsuereth: Yes - this actually leads to many less attributes in the catalog, but the catalog now matches almost exactly what we get in Semconv attribute registry.

A lot of this dance can be fully removed when we're directly doing resolve on V2.

crates/weaver_resolved_schema/src/v2/mod.rs
jsuereth: The removal of `requirement_level` actually leads to a decent "compression" or removal of duplicate attribute definitions that diverge only in requirement level.

In V2, this is critical, as we want *attribute key* to be a namespace for attributes.  In V1 - a pure 'attribute key' is actually ambiguous in the catalog.

crates/weaver_resolved_schema/src/v2/catalog.rs
jsuereth: There's a TODO above this.

Yes we need to check everything, I just didn't create a comparison for annotations yet.

crates/weaver_resolved_schema/src/v2/catalog.rs
jsuereth: I agree we'll need to evolve this over time.  I actually think we need to evolve the output file as well.  Hopefully we can talk more wed in person.

For now, things to consider:

Should there be *single* YAML output, or should we output multiple files?

I'm thinking we want at least *two* outputs (similar to what you had in catalog + registry, but subtle difference):

- Definition Registry (attributes, metrics, entities, events, etc.): These are the "pure" definitions and we use these in livecheck / validation.
- Refinement Registry: These are the "codegen" appropriate refinements of "pure" definition, we would use these in code generation.

Making this explicit in our data model could simplify a couple key use cases.

crates/weaver_forge/src/v2/entity.rs
jsuereth: Apologies, I was rushing for the demo, I'll clean all these up - Didn't want this to be a full review, more of "check the direction" review.

I think the discussion on `Attribute`/`AttributeRef` is the thing to focus on in this review :)

crates/weaver_forge/src/v2/registry.rs
jsuereth: It *will be both*.

Today, Signals are attached to producing entities.  In Phase 2 of Entity SIG - we plan to create a signal that can report entities as a signal (both existence + relationships).  Think CMDB monitoring.

crates/weaver_forge/src/v2/registry.rs
jsuereth: I debated this, I agree it's a future task.

Given entities may be "associated with signals" or a signal in themselves, I think we may need to bifurcate refinements as "resource detection refinements" and "entity-as-a-signal" refinements.  I'd like to defer those decisions for now, as I don't think limiting refinements is onerous for entities today.

crates/weaver_resolved_schema/src/v2/registry.rs
jsuereth: Yes - Once I have a chance to resolve merge conflicts, I'll be adding those in here.

crates/weaver_resolved_schema/src/v2/registry.rs
jsuereth: I'm still not convinced.  I see the registry itself as a namespace going forward, but this is something that needs more discussion going forward.

e.g. in instance where we want a "bundle of things" would it be easier to create a registry to describe that bundle?  That registry would come with some nice properties, like versioning, isolated codegen, docs, etc.

I think I'd like to move to point where, e.g. an implementation specific registry exists in OTEL for any implementation.  I think this may give us some nice properties:

- Implementation-specific concerns can live near the instrumentation
- Implementations can document *exactly* what they provide, while guaranteeing conformance to semconv
- Implementations would document semconv version explicitly via their dependency to semconv repo.  We could allow these to "hold back" while we do major evolutions (e.g. current RPC efforts).

There's obviously downsides too, but it's something I'd like us to consider / think through.

src/registry/resolve.rs
jsuereth: Yes.

For now - to avoid breaking changes - I was planning the following:

- There would be multiple "v2" flags
  - Input format is in the file itself, via `version: '2'`
  - A flag for `policy` engine to declare it needs the V2 data model. (Note this would impact live-check too)
  - A flag for `template` engine to declare it needs the V2 data model.  This would impact JQ + Jinja.
- The resolution engine would be updated to produce both V1 and V2 if any component needs V2 output.
- We reserve `--future` to flag features in V1 that will no longer be supported in V2, but it does NOT alter models used.

crates/weaver_resolved_schema/src/v2/registry.rs
thompson-tomo: I like what you are thinking with regards to implementations being their own registry, but I am torn about the core sem conv registry being split. I can see it helping but also making things harder. Let's discuss implementations as standalone registry in a seperate thread.

If we look at how the namespace value would be used in the  core registry, it would enable a single entry point to be provided for the namespace ie db which lists everything in that namespace rather than splitting it into a seperate entry point per signal type like we do now.

crates/weaver_resolved_schema/src/v2/registry.rs
```diff
@@ -0,0 +1,57 @@
...
+    /// A  list of span signal definitions.
+    pub spans: Vec<Span>,
+
+    /// A  list of metric signal definitions.
+    pub metrics: Vec<Metric>,
+
+    /// A  list of event signal definitions.
+    pub events: Vec<Event>,
+
+    /// A  list of entity signal definitions.
```
thompson-tomo, line 44: Moving across @jsuereth comment https://github.com/open-telemetry/weaver/pull/980/files#r2440979400 to decouple it from discussion about organising definitions into namespaces.

> I'm still not convinced. I see the registry itself as a namespace going forward, but this is something that needs more discussion going forward.
> 
> e.g. in instance where we want a "bundle of things" would it be easier to create a registry to describe that bundle? That registry would come with some nice properties, like versioning, isolated codegen, docs, etc.
> 
> I think I'd like to move to point where, e.g. an implementation specific registry exists in OTEL for any implementation. I think this may give us some nice properties:
> 
> - Implementation-specific concerns can live near the instrumentation
> - Implementations can document exactly what they provide, while guaranteeing conformance to semconv
> - Implementations would document semconv version explicitly via their dependency to semconv repo. We could allow these to "hold back" while we do major evolutions (e.g. current RPC efforts).
> 
> There's obviously downsides too, but it's something I'd like us to consider / think through.

With the above in mind, wouldn't these properties be redundant if we were to split instrumentations into individual registries. With that in mind could we use the standard and indicate via lineage if it is a refinement?

crates/weaver_forge/src/v2/registry.rs
lmolkova: Oh I forgot event refinement id is only unique within events, never mind

crates/weaver_resolved_schema/src/v2/registry.rs
jsuereth: > With that in mind could we use the standard and indicate via lineage if it is a refinement?

The issue is identity / uniqueness.  We are using uniqueness by signal name in these registries now, so we need a "new" identity for refinement, which is why we have that feature.

Also, outside of semantic conventions, I don't see refinements as a "happy path" feature.  Ideally folks won't need to engage with it at all, and complexity doesn't raise that high

crates/weaver_resolved_schema/src/v2/event.rs
jsuereth: I'm going to hold of on this suggestion.  OpenTelemetry goes to lengths to create the Resource->Signal hierarchy.  I think it's still too early to force-flatten here.  We could do so in templates without forcing this to happen all the time.

crates/weaver_resolved_schema/src/v2/registry.rs
jsuereth: added

crates/weaver_resolved_schema/src/v2/event.rs
jsuereth: Fixed

crates/weaver_resolved_schema/src/v2/event.rs
jsuereth: Fixed

crates/weaver_forge/src/v2/entity.rs
jsuereth: Fixed

crates/weaver_forge/src/v2/entity.rs
jsuereth: Fixed

crates/weaver_forge/src/v2/entity.rs
jsuereth: Fixed

crates/weaver_forge/src/v2/entity.rs
jsuereth: Fixed

crates/weaver_forge/src/v2/entity.rs
jsuereth: Fixed

crates/weaver_forge/src/v2/event.rs
jsuereth: Fixed

crates/weaver_forge/src/v2/event.rs
jsuereth: Yeah - I need to figure out that todo...

crates/weaver_forge/src/v2/metric.rs
jsuereth: Fixed

crates/weaver_forge/src/v2/metric.rs
jsuereth: Fixed

crates/weaver_forge/src/v2/metric.rs
jsuereth: Fixed

crates/weaver_forge/src/v2/span.rs
jsuereth: Fixed

crates/weaver_resolved_schema/src/v2/catalog.rs
jsuereth: I've updated and optimised this portion

crates/weaver_resolved_schema/src/v2/entity.rs
jsuereth: yeah, entities will eventually be signals that can be sent on their own.

crates/weaver_resolved_schema/src/v2/entity.rs
jsuereth: Fixed

crates/weaver_resolved_schema/src/v2/catalog.rs
jsuereth: Fixed this.

crates/weaver_forge/src/v2/metric.rs
jsuereth: Fixed

lmolkova:
I don't see any attribute-refs under any of the signals  in the resolved schema when doing `resolve --v2`.

E.g. 
```yaml
entities:
  - type: android
    brief: |
      The Android platform on which the Android application is running.
    stability: development
  - type: app
    brief: |
      An app used directly by end users — like mobile, web, or desktop.
    stability: development
  - type: aws.ecs
    brief: |
      Entities used by AWS Elastic Container Service (ECS).
    stability: development
```

crates/weaver_forge/src/v2/entity.rs
```diff
@@ -0,0 +1,47 @@
...
+    #[serde(default)]
+    #[serde(skip_serializing_if = "Vec::is_empty")]
+    pub description: Vec<EntityAttribute>,
+
+    /// Common fields (like brief, note, annotations).
+    #[serde(flatten)]
+    pub common: CommonFields,
+}
+
+/// A special type of reference to attributes that remembers entity-specicific information.
```
jerbly:
```suggestion
/// A special type of reference to attributes that remembers entity-specific information.
```

crates/weaver_resolved_schema/src/v2/event.rs
jsuereth: Marking resolved

crates/weaver_resolved_schema/src/v2/mod.rs
jsuereth: Marking resolved

crates/weaver_resolved_schema/src/v2/registry.rs
jsuereth: I don't think we're going to resolve this discussion on this PR.

Also - given the GC direction of more Federation in semantic conventions, it's likely we WILL be splitting apart semconv in a reasonable way as tooling allows

crates/weaver_forge/src/v2/metric.rs
jsuereth: Updated

crates/weaver_forge/src/v2/registry.rs
jsuereth: marking resolved

lquerel (approved):
LGTM. Thanks!

Issue #986, opened 2025-10-24
[entities] Disallow `requirement_level` field in the `identity` section


dmitryax:
**Is your feature request related to a problem? Please describe.**

Per discussion with @jsuereth, all identifying attributes should be required. Therefore, having a `requirement_level` field in the `identity` section of the entity definition in [semconv-syntax.v2.md#entities-definition](https://github.com/open-telemetry/weaver/blob/main/schemas/semconv-syntax.v2.md#entities-definition) creates confusion.

**Describe the solution you'd like**

Disallow the `requirement_level` field in the `identity` section of the entity definition in the v2 syntax and assume all identifying attributes are required.

<sub>**Tip**: [React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

PR #987, opened 2025-10-24
[chore] Update entity example in the language v2 doc

dmitryax:
- `role` field shouldn't be allowed since it's defined by the group role
- `requirement_level` shouldn't be used in the identity group since all of them should be required

Related to https://github.com/open-telemetry/weaver/issues/986

### DIFF
--- a/schemas/semconv-syntax.v2.md
+++ b/schemas/semconv-syntax.v2.md
@@ -374,12 +374,11 @@ entities:
     stability: stable
     identity:
       - ref: service.name
-        requirement_level: required
       - ref: service.namespace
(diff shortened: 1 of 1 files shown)


lmolkova (approved):

schemas/semconv-syntax.v2.md
```diff
@@ -374,12 +374,11 @@ entities:
     stability: stable
     identity:
       - ref: service.name
-        requirement_level: required
```
jsuereth, line 377: Note: We still need to change the underlying schema model to disallow this.

PR #992, opened 2025-11-03
fix: allow loading registries and policies from hidden directories

### LINKED ISSUE #958: Weaver cannot load registry directory beginning with `.`
**Describe the bug**
this command fails because the registry directory begins with `.` character
```
weaver registry generate \
	    --registry .semconv \
	    --templates templates \
	    --config /templates/registry/markdown/weaver.yaml
```

this command succeeds
```
weaver registry generate \
	    --registry semconv \
	    --templates templates \
	    --config /templates/registry/markdown/weaver.yaml
```

**To Reproduce**
see above

**Expected behavior**
I expect hidden directories to work when trying to render output using `weaver generate`

**Additional context**
v0.18.0, I also tested back to v0.16.0 and the problem persists. This problem also occurs if I use an absolute path instead of a relative path.

kuklyy:
## Summary
Removes unnecessary `is_hidden()` filter that prevented loading from directories starting with `.` (dot). The existing file type filters are sufficient to exclude problematic paths.

## Changes
- Removed `is_hidden()` function and filter from `weaver_resolver/src/lib.rs` (registry loading)
- Removed `is_hidden()` function and filter from `weaver_checker/src/lib.rs` (policy loading)
- Added integration test to verify hidden registry directories work correctly

## Test plan
- [x] Added test `test_hidden_directory` that verifies loading from `.hidden_reg/` directory
- [x] All existing tests pass
- [x] CI quality checks pass (fmt, clippy, doc)

Fixes #958

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/crates/weaver_checker/src/lib.rs
+++ b/crates/weaver_checker/src/lib.rs
@@ -312,14 +312,6 @@ impl Engine {
         policy_dir: P,
         policy_glob_pattern: &str,
     ) -> Result<usize, Error> {
-        fn is_hidden(entry: &DirEntry) -> bool {
-            entry
-                .file_name()
-                .to_str()
-                .map(|s| s.starts_with('.'))
-                .unwrap_or(false)
-        }
-
         let mut errors = Vec::new();
         let mut added_policy_count = 0;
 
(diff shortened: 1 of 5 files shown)


lquerel (changes requested):
I'd prefer an opt-in parameter (e.g. --include-hidden) that allows exploring hidden files/directories, to avoid surprising other users who would not find it normal for a CLI to inspect files in locations such as .git, .svn, .idea, .vscode, .cache, .local, .venv, .aws/credentials, .kube/config, ...

lmolkova:
@kuklyy thanks for the contribution! It seems it's been a while and there are unresolved discussions on the PR. I'm going to close it, but you're welcome to come back to it if you want.

PR #1016, opened 2025-11-19
Update code so weaver check now supports v2 schema.

jsuereth:
- [x] Adds a new flag which will force all policies to use *new* v2 schema
 - [x] Tests after_resolution policy against v2 schema
 - [x] Tests comparison with baseline policy with v2 schema
 - [x] Adds warning to before_resolution for v2 schema and policies. (We will not support before_resolution in v2, for now)


src/registry/mod.rs
```diff
@@ -172,6 +172,10 @@ pub struct PolicyArgs {
     /// Display the policy coverage report (useful for debugging).
     #[arg(long, default_value = "false")]
     pub display_policy_coverage: bool,
+
+    /// Whether or not the policy engine should use the V2 schema.
+    #[arg(long, default_value = "false")]
+    pub policy_use_v2: bool,
```
lmolkova: could we use the same flag as on generate --v2?

lmolkova (approved):
Tried it out in semconv, works like a charm!

src/registry/mod.rs
jerbly: FYI I moved the `--v2` switch into `RegistryArgs` in my emit PR since this is needed everywhere we use a registry. https://github.com/open-telemetry/weaver/pull/1019/files#diff-eaeed62b438e3eea9b8de29790b3a95e17d3eddd4eed6d54b7a12666dd7f0a81R157

jsuereth: I purposely had the separate, so you could migrate them individually.

I *can* make it all one flag, is that the transition period we want?  All of it at once?

jerbly: IMO it's already confusing enough. We should go all-in.

jsuereth: done! PTAL

jerbly (approved):
LGTM

Issue #694, opened 2025-04-09
Document exact datamodel for values sent to REGO policies


jsuereth:
As part of a "getting started" guide, we should provide a tabular-type view of which fields (and their nested structure) are available in Rego policies, by namespace.

Related to #583

PR #1022, opened 2025-11-23
Schema v2 live-check

jerbly:
Convert live-check to work with both v1 and v2 resolved registries. This has been achieved with enums and generics to limit the amount of code duplication since we'll have dual support for some time.

This approach also means that the interface to rego policies is very similar and will rarely require any changes. In fact all the rego policies for live-check in the tests are dual version compatible and therefore unchanged in this PR.

- [x] Initially working
- [x] Refactor duplications
- [x] v2 tests

### DIFF
--- a/tests/registry_emit.rs
+++ b/tests/registry_emit.rs
@@ -23,7 +23,10 @@ fn test_emit_with_live_check_v2() {
 fn run_emit_with_live_check_test(use_v2: bool) {
     // Create a temporary directory
     let temp_dir = tempdir().expect("Failed to create temporary directory");
-    let temp_dir_path = temp_dir.path().to_str().unwrap();
+    let temp_dir_path = temp_dir
+        .path()
+        .to_str()
+        .expect("Failed to convert temp directory path to string");
 
     // Build the arguments for live check command
     let mut live_check_args = vec![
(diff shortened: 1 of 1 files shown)


crates/weaver_live_check/src/advice.rs
```diff
@@ -201,6 +202,32 @@ impl Advisor for StabilityAdvisor {
 /// An advisor that checks if an attribute has the correct type
 pub struct TypeAdvisor;
 
+/// Trait to abstract over different attribute types for checking
+trait CheckableAttribute {
```
jsuereth, line 206: Nice!  I like this idea

crates/weaver_live_check/src/live_checker.rs
```diff
@@ -7,53 +7,76 @@ use std::collections::HashMap;
...
-            for attribute in &group.attributes {
-                let attribute_rc = Rc::new(attribute.clone());
-                match attribute.r#type {
-                    AttributeType::Template(_) => {
-                        templates_by_length.push((attribute.name.clone(), attribute_rc.clone()));
-                        let _ = semconv_templates.insert(attribute.name.clone(), attribute_rc);
-                    }
-                    _ => {
-                        let _ = semconv_attributes.insert(attribute.name.clone(), attribute_rc);
+            VersionedRegistry::V2(registry) => {
```
jsuereth, line 64: I like how it's more clear what signals are supported here.

crates/weaver_live_check/src/live_checker.rs
```diff
@@ -477,222 +500,438 @@ mod tests {
...
+                        body: None,
+                        annotations: None,
+                    },
+                ],
+            })
         }
     }
 
     #[test]
     fn test_custom_rego() {
```
jsuereth: Does custom rego policy work with V2?  Does this need a test?

jsuereth (commented):
Overall looks great! Wanted to check on V2 + custom rego policies

crates/weaver_live_check/src/live_checker.rs
jerbly: Yes, custom rego does work with V2. The test policies here were unchanged, and the builtin otel policies also. This is due to the way I provide the signal as input to the policy "untagged" by serde. If you happen to need one of the fields that has moved position in the JSON then you would need to fix your policy. We should fix this with documentation: #694 

There are three more examples in these tests that exercise v1 and v2 with custom rego. This is the only one I didn't do that with because it doesn't look at the semconv attribute, only the sample so it's moot. However, I should make it a dual test for consistency. I'll do that now.

jsuereth: ah, cool, can you open a task on the tracking bug to update the docs for this then?

lquerel:
Great!

PR #1038, opened 2025-11-25
Moves away from strict Violation enum to open Violation based on Advice

jsuereth:
## Changes

- Remove `Advice` and `SemconvAttribute` as types, only use `PolicyFinding` now.
- Hide sub-mod in `weaver_checker` crate, so everything is top-level now.
- Create a custom deserializer to support existing Rego policies for both `Advice` and `SemconvAttribute` in `PolicyFinding`
- Update tests to use Violation
- Update `weaver_live_check` to use Violation instead of `Advice`.
- Update docs

This should give us all the same goodness as live check going forward, in terms of flexible policies.

### DIFF
--- a/crates/weaver_checker/README.md
+++ b/crates/weaver_checker/README.md
@@ -261,24 +261,36 @@ groups:
 [
   {
     "type": "semconv_attribute",
-    "id": "attr_stability_deprecated",
-    "category": "attribute",
-    "group": "registry.network1",
-    "attr": "protocol.name"
+    "context": {
+      "id": "attr_stability_deprecated",
+      "category": "attribute",
+      "group": "registry.network1",
+      "attr": "protocol.name",
+    },
+    "message": "...",
+    "level": "violation",
   },
   {
-    "type": "semconv_attribute",
-    "id": "attr_removed",
-    "category": "schema_evolution",
-    "group": "registry.network1",
-    "attr": "protocol.name.3"
+    "id": "semconv_attribute",
+    "context": {
+      "id": "attr_stability_deprecated",
+      "category": "attribute",
+      "group": "registry.network1",
+      "attr": "protocol.name",
+    },
+    "message": "...",
+    "level": "violation",
   },
   {
-    "type": "semconv_attribute",
-    "id": "registry_with_ref_attr",
-    "category": "attribute_registry",
-    "group": "registry.network1",
-    "attr": "protocol.port"
+    "id": "semconv_attribute",
+    "context": {
+      "id": "attr_stability_deprecated",
+      "category": "attribute",
+      "group": "registry.network1",
+      "attr": "protocol.name",
+    },
+    "message": "...",
+    "level": "violation",
   }
 ]
 ```


jerbly:
I guess it doesn't matter too much as it's internal, but I'm curious what was behind the decision to rename Advice to Violation, rather than Violation to Advice? Since now we can have Violations that are not Violations whereas, before one type of Advice was a Violation.

crates/weaver_checker/src/finding.rs
```diff
@@ -7,53 +7,187 @@ use serde::{Deserialize, Serialize};
...
+    /// machine-readable string that categorizes the violation.
+    pub id: String,
+
+    /// The context associated with the violation e.g. { "attribute_name": "foo.bar", "attribute_value": "bar" }
+    /// The context should contain all dynamic parts of the message
+    /// Context values may be used with custom templates and filters to customize reports.
+    pub context: Value,
+
+    /// The human-readable message of the violation e.g. "This attribute 'foo.bar' is deprecated, reason: 'use foo.baz'"
+    /// The message, along with signal_name and signal_type, should contain enough information to understand the advice and
```
lmolkova, line 27: nit

```suggestion
    /// The message, along with signal_name and signal_type, should contain enough information to understand the violation and
```

jsuereth:
> I guess it doesn't matter too much as it's internal, but I'm curious what was behind the decision to rename Advice to Violation, rather than Violation to Advice? Since now we can have Violations that are not Violations whereas, before one type of Advice was a Violation.

@jerbly I actually don't like the name `Violation` or `Advice` here. I think Advice is a bit too tied to live-check (and doesn't have any connotation it can be used to fail things), and `Violation` makes it awkward to make informational/warning messages.

I asked gemini and it suggested `Report` or `PolicyFinding` here, where inside `Report` you can have `informational`/`advice`, `warning` and `violation`.  WDYT of these names?

lmolkova (approved):
Nice!


Could you please update docs:
-  https://github.com/jsuereth/weaver/blob/1bfbf330e04d258d31b13ab9017374ac092d1937/crates/weaver_live_check/README.md#advice and the changelog
- https://github.com/jsuereth/weaver/blob/1bfbf330e04d258d31b13ab9017374ac092d1937/crates/weaver_checker/README.md#policy-examples
?

(it can be a follow up)

and add a changelog record

lquerel:
> I asked gemini and it suggested `Report` or `PolicyFinding` here, where inside `Report` you can have `informational`/`advice`, `warning` and `violation`.  WDYT of these names?

@jsuereth personally I prefer PolicyFinding because it's more informative.

jsuereth:
I'll updated to `PolicyFinding` before merging

crates/weaver_checker/src/finding.rs
jsuereth: Fixed

jsuereth: Fixed

Issue #1018, opened 2025-11-20
Support for logs in emit


wsmoak:
**Is your feature request related to a problem? Please describe.**

The example for emitting events explains how to do spans and metrics, but not logs.
https://github.com/open-telemetry/opentelemetry-weaver-examples/tree/main/basic#emit-example-telemetry

@jerbly confirmed in Slack that logs are not supported yet.

**Describe the solution you'd like**

Running `weaver registry emit -r model` should be able to emit logs (events) as well as spans and traces

<sub>**Tip**: [React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

PR #1042, opened 2025-11-28
Emit and live-check event support

### LINKED ISSUE #1017: Support logs & events in live-check
We should provide means to validate log records with live check. Events (logs with EventName) can be validated against the definition. For both, events and logs, it should be possible to run attribute advises.

jerbly:
Resolves #1017 and #1018

- Emit will now send events as defined in the model
- Live-check now provides advice for events

image

### DIFF
--- a/crates/weaver_live_check/src/live_checker.rs
+++ b/crates/weaver_live_check/src/live_checker.rs
@@ -1644,7 +1644,7 @@ mod tests {
 
         // Check the statistics
         assert_eq!(stats.total_entities, 8);
-        assert_eq!(stats.total_entities_by_type.get("event"), Some(&4));
+        assert_eq!(stats.total_entities_by_type.get("log"), Some(&4));
         assert_eq!(stats.total_entities_by_type.get("attribute"), Some(&4));
 
         // Check advisor advice types


src/registry/otlp/otlp_ingester.rs
```diff
@@ -47,9 +47,43 @@ impl OtlpIterator {
...
+                                    sample_attribute_from_key_value(&attribute),
+                                ));
+                            }
+                        }
+
+                        for log_record in scope_log.log_records {
+                            // Only process log records that have event_name set.
+                            // Per the OTLP spec: "Presence of event_name on the log record identifies this record as an event."
+                            if log_record.event_name.is_empty() {
+                                continue;
```
lmolkova: I think it makes sense to live-check logs in custom policies. E.g. one can check that all logs are structured and/or certain logs that have specific attributes also have other attributes (e.g. if you have `exception.message` you must also have `exception.type`).

There won't be a reference definition to validate against (similarly to spans today).

crates/weaver_live_check/src/lib.rs
```diff
@@ -291,7 +302,8 @@ pub enum SampleRef<'a> {
     ExponentialHistogramDataPoint(&'a SampleExponentialHistogramDataPoint),
     /// A sample exemplar
     Exemplar(&'a SampleExemplar),
-    // TODO: add logs
+    /// A sample event
+    Event(&'a SampleEvent),
```
lmolkova: see my other comment - I think we should eventually have a broader log advice.

src/registry/otlp/otlp_ingester.rs
jerbly: OK, I'll allow SampleEvents without `event_name` - these will not be able to be Type checked against `event` definitions, but will have their attributes checked and custom rego like all Samples.

jerbly: Fixed in this commit: https://github.com/open-telemetry/weaver/pull/1042/commits/882b7dff7a2ccd5bd6ab5d881be50f2b319bab20

crates/weaver_live_check/src/lib.rs
jerbly: Fixed in: https://github.com/open-telemetry/weaver/pull/1042/commits/882b7dff7a2ccd5bd6ab5d881be50f2b319bab20

jsuereth: @jerbly I'm not sure I see where that's fixed.

I think what Liudmilla means is that while semconv only allows you define events, *weaver* may want to allow you to have custom policies / advice on logs that do not have an event_name attribute.  If I'm reading the diff you list there correctly, that's still not an option here.

jerbly: Oh... prior to this commit I was deliberately excluding samples that did not have `event_name` set. I removed that restriction which means attributes are assessed and you should be able to have a custom rego policy... I'll add a rego policy to the test suite to show this...

jerbly: To prove that you can run custom policies on events without `event_name` I added a couple of policies and a test to the latest commit. One of these specifically checks for an empty `event_name` and returns a violation if the `body` is also empty.

crates/weaver_live_check/src/advice.rs
```diff
@@ -125,6 +131,31 @@ impl Advisor for DeprecatedAdvisor {
...
+                                DEPRECATION_NOTE_ADVICE_CONTEXT_KEY: deprecated,
+                            }),
+                            message: format!(
+                                "Event '{}' is deprecated; reason = {}, note = {}",
+                                sample_event.event_name.clone(),
+                                deprecated_to_reason(deprecated),
+                                deprecated
+                            ),
+                            level: FindingLevel::Violation,
+                            signal_type: Some("event".to_owned()),
```
lmolkova: I think we should use `log` as a signal (that's the spec terminology). Events are subtype of logs

lmolkova (approved):
I think we should rename `event` to `log` in signal_type  and other public-facing places to align with the spec language and be future proof

crates/weaver_live_check/src/advice.rs
jerbly: I've changed everything to Log. In the standard output I've made a small distinction between Log and Log Event see the screenshot.
image

PR #1045, opened 2025-12-01
Policy Findings emitted as OTLP log records as they arise

jerbly:
Whenever a PolicyFinding is created in live-check a log_record is emitted to your configured OTLP endpoint. Here's an example from a Collector in debug mode:
```
Flags: 0
LogRecord #21
ObservedTimestamp: 2025-12-01 00:04:58.49238 +0000 UTC
Timestamp: 1970-01-01 00:00:00 +0000 UTC
SeverityText: INFO
SeverityNumber: Info(9)
EventName: weaver.live_check.finding
Body: Str(Conditionally required attribute 'server.port' is not present.)
Attributes:
     -> weaver.finding.id: Str(conditionally_required_attribute_not_present)
     -> weaver.finding.level: Str(information)
     -> weaver.finding.context.attribute_name: Str(server.port)
     -> weaver.sample.type: Str(histogram_data_point)
     -> weaver.sample.signal_type: Str(metric)
     -> weaver.sample.signal_name: Str(rpc.client.duration)
Trace ID: 
Span ID: 
```
...and a screenshot from a back-end:
image

### DIFF
--- a/crates/weaver_live_check/README.md
+++ b/crates/weaver_live_check/README.md
@@ -264,7 +264,7 @@ Each policy finding is emitted as an OTLP log record with the following structur
 **Event Name**: `weaver.live_check.finding`
 
 **Resource Attributes**:
-- `service.name`: "weaver"
+- `service.name`: set by OTEL_SERVICE_NAME or OTEL_RESOURCE_ATTRIBUTES environment variables, defaulting to `weaver`
 
 **Log Attributes**:
 - `weaver.finding.id`: Finding type identifier (e.g., "required_attribute_not_present")


martinjt:
The way it works in .net is that there is a `FormattedMessage` attribute that you can opt in to having populated.

jerbly:
I'm wondering if this is a specific thing that dot net does? The [docs in the proto file](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/logs/v1/logs.proto#L169-L172), for example, say this:
```
  // A value containing the body of the log record. Can be for example a human-readable
  // string message (including multi-line) describing the event in a free form or it can
  // be a structured data composed of arrays and maps of other values. [Optional].
  opentelemetry.proto.common.v1.AnyValue body = 5;
```
I have not used logs much myself so looking for more opinions here before going one way or the other: @open-telemetry/weaver-approvers

jsuereth:
> I'm wondering if this is a specific thing that dot net does? The [docs in the proto file](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/logs/v1/logs.proto#L169-L172), for example, say this:
> 
> ```
>   // A value containing the body of the log record. Can be for example a human-readable
>   // string message (including multi-line) describing the event in a free form or it can
>   // be a structured data composed of arrays and maps of other values. [Optional].
>   opentelemetry.proto.common.v1.AnyValue body = 5;
> ```
> 
> I have not used logs much myself so looking for more opinions here before going one way or the other: @open-telemetry/weaver-approvers

I don't believe Java does this, not sure for other languages, but I don't think it's the "norm" or at least, it's not specified logging should do this.

crates/weaver_live_check/src/advice/deprecated_advisor.rs
```diff
@@ -0,0 +1,123 @@
+// SPDX-License-Identifier: Apache-2.0
+
+//! Deprecation detection advisor
+
+use serde_json::json;
+use std::rc::Rc;
+use weaver_checker::{FindingLevel, PolicyFinding};
+use weaver_semconv::deprecated::Deprecated;
+
+use super::{Advisor, FindingBuilder};
```
lmolkova, line 10: I'm curious what's the difference between `advise` and `finding`. Could we unify them and use one term to describe both? This applies to APIs and attributes/event naming

crates/weaver_live_check/README.md
```diff
@@ -226,6 +226,54 @@ These should be self-explanatory, but:
...
+
+**Severity**:
+- `Error` (17) for `violation` level findings
+- `Warn` (13) for `improvement` level findings
+- `Info` (9) for `information` level findings
+
+**Event Name**: `weaver.live_check.finding`
+
+**Resource Attributes**:
+- `service.name`: "weaver"
```
lmolkova: should it be hardcoded? we can default to weaver, but should eventually support customizing  resource attributes. This is definitely a future concern

crates/weaver_live_check/README.md
```diff
@@ -226,6 +226,54 @@ These should be self-explanatory, but:
...
+
+**Log Attributes**:
+- `weaver.finding.id`: Finding type identifier (e.g., "required_attribute_not_present")
+- `weaver.finding.level`: Finding level as string ("violation", "improvement", "information")
+- `weaver.finding.context.*`: Flattened JSON context with dot notation
```
lmolkova: by flattened, do you mean one-level (there could totally be more than one)? then maybe 

```suggestion
- `weaver.finding.context.<key>`: Key-value pairs provided in the context. Each  pair is recorded as a single attribute.
```

crates/weaver_live_check/README.md
```diff
@@ -226,6 +226,54 @@ These should be self-explanatory, but:
...
+- `weaver.sample.type`: Sample type (e.g., "attribute", "span", "metric")
+- `weaver.sample.signal_type`: Signal type (e.g., "span", "metric")
+- `weaver.sample.signal_name`: Signal name (e.g., span name or metric name)
```
lmolkova: nit: span name makes a bad signal name, afaik we don't even set signal_name it for spans

crates/weaver_live_check/src/advice/deprecated_advisor.rs
jerbly: Until very recently we had:
- `Advisors` who `advise` and produce `Advice`

Now we have:
- `Advisors` who `advise` and produce `PolicyFindings`

I'm not sure what I would rename `Advisor` and the verb `advise` to?

crates/weaver_live_check/README.md
jerbly: Say the JSON is:
```json
{
    "foo" : {
        "bar" : "baz"
    }
}
```
It would be "flattened" to `weaver.finding.context.foo.bar`. Do you think this should be `weaver.finding.context.foo_bar` to make it more obviously a dictionary of single key-value pairs?

crates/weaver_live_check/README.md
jerbly: You're right, that's a copy paste error. I'm also wondering if these last few should be:
- `weaver.finding.sample_type`
- `weaver.finding.signal_type`
- `weaver.finding.signal_name`

WDYT?

crates/weaver_live_check/src/advice/deprecated_advisor.rs
lmolkova: Could `Advisors` produce `PolicyAdvices` ?

crates/weaver_live_check/README.md
lmolkova: I like `weaver.finding.sample_type`!

crates/weaver_live_check/src/advice/deprecated_advisor.rs
jsuereth: I think we should have Advisors produce Findings.  The Findings can be `advice` `violation` or `warning`.

Previously `Advice` as a `violation` was a bit awkward.  So I don't think we're any "worse" here if Advisors produce a Finding - where  finding could be one of these three things.

crates/weaver_live_check/src/otlp_logger.rs
```diff
@@ -0,0 +1,497 @@
...
+            .with_endpoint(endpoint)
+            .build()
+            .map_err(|e| Error::OutputError {
+                error: format!("Failed to create OTLP log exporter: {}", e),
+            })?;
+
+        let provider = SdkLoggerProvider::builder()
+            .with_resource(
+                Resource::builder()
+                    .with_service_name(WEAVER_SERVICE_NAME)
```
jsuereth: We should optionally use the OTEL_SERVICE_NAME env variable if it exists.

src/registry/live_check.rs
```diff
@@ -122,6 +122,18 @@ pub struct RegistryLiveCheckArgs {
     #[clap(long, default_value = "4317")]
     otlp_grpc_port: u16,
 
+    /// Enable OTLP log emission for live check policy findings
+    #[arg(long, default_value = "false")]
+    emit_otlp_logs: bool,
```
jsuereth, line 127: Nit: Should we bundle these into a struct?

jsuereth (approved):

crates/weaver_live_check/src/otlp_logger.rs
jerbly: fixed in [cf761cc](https://github.com/open-telemetry/weaver/pull/1045/commits/cf761cc6f17b3aeb84bd6e0b289f7113e772c18b)

crates/weaver_live_check/README.md
jerbly: fixed in [1b10784](https://github.com/open-telemetry/weaver/pull/1045/commits/1b107840e15c04831b670a9518723c31b4417fd3)

PR #1051, opened 2025-12-03
Prom name function

zeitlinger:
- prometheus_metric_names: returns all possible prometheus metric names using https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk_exporters/prometheus.md#configuration
- prometheus_unit_name: translates prometheus unit name

### DIFF
--- a/crates/weaver_forge/src/extensions/otel.rs
+++ b/crates/weaver_forge/src/extensions/otel.rs
@@ -2024,8 +2024,9 @@ mod tests {
             instrument: "histogram".to_owned(),
         };
 
+        let ctx = json!({"metric": metric});
         assert_eq!(
-            env.render_str("{{ metric | prometheus_unit_name }}", &metric)
+            env.render_str("{{ metric | prometheus_unit_name }}", &ctx)
                 .unwrap(),
             "milliseconds"
         );
@@ -2037,8 +2038,9 @@ mod tests {
             instrument: "histogram".to_owned(),
         };
 
+        let ctx = json!({"metric": metric});
         assert_eq!(
-            env.render_str("{{ metric | prometheus_unit_name }}", &metric)
+            env.render_str("{{ metric | prometheus_unit_name }}", &ctx)
                 .unwrap(),
             "unknown_unit"
         );
@@ -2050,11 +2052,12 @@ mod tests {
             instrument: "histogram".to_owned(),
         };
 
+        let ctx = json!({"metric": metric});
         // NoTranslation strategy
         assert!(env
             .render_str(
                 "{{ metric | prometheus_metric_name(translation_strategy='NoTranslation') }}",
-                &metric
+                &ctx
             )
             .is_ok());
 
@@ -2062,7 +2065,7 @@ mod tests {
         assert!(env
             .render_str(
                 "{{ metric | prometheus_metric_name(translation_strategy='UnderscoreEscapingWithoutSuffixes') }}",
-                &metric
+                &ctx
             )
             .is_ok());
 
@@ -2070,7 +2073,7 @@ mod tests {
         assert!(env
             .render_str(
                 "{{ metric | prometheus_metric_name(translation_strategy='NoUTF8EscapingWithSuffixes') }}",
-                &metric
+                &ctx
             )
             .is_ok());
 
@@ -2078,20 +2081,20 @@ mod tests {
         assert!(env
             .render_str(
                 "{{ metric | prometheus_metric_name(translation_strategy='UnderscoreEscapingWithSuffixes') }}",
-                &metric
+                &ctx
             )
             .is_ok());
 
         // Test error when translation_strategy is missing
         assert!(env
-            .render_str("{{ metric | prometheus_metric_name }}", &metric)
+            .render_str("{{ metric | prometheus_metric_name }}", &ctx)
             .is_err());
 
         // Test error with invalid translation_strategy
         assert!(env
             .render_str(
                 "{{ metric | prometheus_metric_name(translation_strategy='InvalidStrategy') }}",
-                &metric
+                &ctx
             )
             .is_err());
 
@@ -2099,7 +2102,7 @@ mod tests {
         assert!(env
             .render_str(
                 "{% for name in metric | prometheus_metric_names(translation_strategy='NoTranslation') %}{{ name }}{% endfor %}",
-                &metric
+                &ctx
             )
             .is_ok());
 
@@ -2110,10 +2113,11 @@ mod tests {
             instrument: "histogram".to_owned(),
         };
 
+        let ctx = json!({"metric": metric});
         let result = env
             .render_str(
                 "{% for name in metric | prometheus_metric_names(expand_summary_and_histogram=true) %}{{ name }}|{% endfor %}",
-                &metric
+                &ctx
             )
             .unwrap();
         assert!(!result.is_empty());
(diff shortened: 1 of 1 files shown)


crates/weaver_forge/src/extensions/otel.rs
```diff
@@ -33,6 +34,8 @@ pub(crate) fn add_filters(env: &mut minijinja::Environment<'_>) {
     env.add_filter("camel_case_const", camel_case_const);
     env.add_filter("snake_case_const", snake_case_const);
     env.add_filter("screaming_snake_case_const", screaming_snake_case_const);
+    env.add_filter("prom_name", prom_name);
```
lmolkova:
```suggestion
    env.add_filter("prometheus_metric_name", prometheus_metric_name);
```

crates/weaver_forge/src/extensions/prom.rs
```diff
@@ -0,0 +1,197 @@
...
+
+pub(crate) fn get_prom_units(unit: &str) -> Option<&'static str> {
+    match unit {
+        // Time
+        "d" => Some("days"),
+        "h" => Some("hours"),
+        "min" => Some("minutes"),
+        "s" => Some("seconds"),
+        "ms" => Some("milliseconds"),
+        "us" => Some("microseconds"),
```
martincostello, line 167: Is `μs` something that could/should be catered for too?

zeitlinger: I just copied that from the rust SDK - and I think that comes from the  go reference implementation

jsuereth: We use UCCM standard, not sure that's emojii-ified yet.

martincostello: I don't think the Greek letter mu is an emoji 😄

jsuereth: Good point - I should have mentioned "any non-US character set" - https://ucum.org/ucum <- They explicitly say they only use US-ASCII here.

Interesting, *by specification* when printing `us`, you're supposed to use Greek "mu", not `u`.

crates/weaver_forge/src/extensions/prom.rs
```diff
@@ -0,0 +1,380 @@
+// copied from https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-prometheus/src/utils.rs
+// with minor modifications
+// - using String instead of Cow
```
jsuereth: Can you modify this comment to be about `str` instead of Cow?  `str` and `String` are very different things.

zeitlinger: most functions return `String` instead of `Cow<'static, str>`, 
e.g. `pub(crate) fn get_unit_suffixes(unit: &str) -> Option<String> {`

jsuereth: Ah, if that's the case, we should try to return `Cow` in the future, but we can make that cleanup later if this is beyond your rust tolerance.

zeitlinger: no, I can certainly add it - I just wasn't sure what the conventions in this repo are.

jsuereth: We should make that more explicit.  Regarding `Cow` vs. `String` - I try to stick to the convention used in the code I'm near.  Here - you're working next to `minijinja`, so we try to stick with types it supports natively.  for `jaq`, e.g. it's a bit more gnarly.

*if* you're never going to return a `&str`, then I don't think you need to use `Cow`.

zeitlinger: Some are `&str` is returned, e.g. https://github.com/open-telemetry/weaver/blob/97876c8d006b50e93e183096a0bf6783f00bdfbd/crates/weaver_forge/src/extensions/prom.rs#L20

jsuereth:
@zeitlinger Yes, I'm suggesting we should *also* have one that returns a single name with chosen caveats, for codegen.  You may also want the docs to indicate how this is intended to be used.

For filters, we think about two primary locations: 

- Codegen
- Documentation

I believe your use case is related to constructing dashboards or otherwise consuming metrics which MAY  have a set of configuration applied to them.  However, as folks in the Prometheus ecosystem are looking at using weaver to define metrics independently of OTEL - I think it'd be nice to re-use what you set up here to get a single "good" metric name for codegen in prometheus.

Maybe @ArthurSens could comment on that use case.

crates/weaver_forge/src/extensions/prom.rs
```diff
@@ -0,0 +1,386 @@
...
+        for (input, want) in tests {
+            assert_eq!(want, sanitize_name(&input.into()), "input: {input}");
+        }
+    }
+
+    #[test]
+    fn test_get_unit_suffixes() {
+        let test_cases = vec![
+            // Direct match
+            ("g", Some(Cow::Borrowed("grams"))),
```
jsuereth, line 286: Minor Nit: I *think* you can do comparisons here without putting the `&str` into a cow.  It'd be nice if we don't rely on ownership output of this function in case it changes.

zeitlinger:
@ArthurSens @jsuereth all added now

@matt-hensley fyi

crates/weaver_forge/src/extensions/otel.rs
```diff
@@ -212,14 +218,89 @@ pub(crate) fn screaming_snake_case_const(input: &str) -> String {
...
     if name.is_undefined() {
         name = input.get_attr("key")?;
     }
     Ok(name)
 }
 
+pub(crate) fn parse_translation_strategy(s: &str) -> Result<TranslationStrategy, minijinja::Error> {
+    match s {
+        "NoTranslation" => Ok(TranslationStrategy::NoTranslation),
+        "UnderscoreEscapingWithoutSuffixes" => Ok(TranslationStrategy::UnderscoreEscapingWithoutSuffixes),
```
jsuereth, line 232: Nit: Would it be easier to have two arguments?

One for escaping and a second for suffixes?

crates/weaver_forge/README.md
```diff
@@ -595,6 +603,82 @@ The following filters are available (the code for all available extension can be
...
+  - `"NoTranslation"`: Returns the original metric name unchanged
+  - `"UnderscoreEscapingWithoutSuffixes"`: Sanitizes the name (dots → underscores) without adding unit/instrument suffixes
+  - `"NoUTF8EscapingWithSuffixes"`: Adds unit and instrument suffixes without name sanitization
+  - `"UnderscoreEscapingWithSuffixes"`: Sanitizes the name and adds unit/instrument suffixes
+
+  Example usage:
+  
+  ```jinja
+  {# Returns: "http.request.duration" #}
+  {{ metric|prometheus_metric_name(translation_strategy="NoTranslation") }}
```
jsuereth, line 631: I'm thinking you can have the translation strategy be optional (pick a default) and use two parameters (each with defaults).  Something like:

```
{{ metric | prometheus_metric_name(escaping=true, suffixes=true) }} -> http_request_duration_seconds
{{ metric | prometheus_metric_name(escaping=true, suffixes=false) }} -> http_request_duration
{{ metric | prometheus_metric_name(escaping=false) }} -> http.request.duration
```

zeitlinger: the translation strategy is defined in the spec - as @ArthurSens pointed out above: https://github.com/open-telemetry/weaver/pull/1051#issuecomment-3660280500

I think it's good to stick to that naming there and not have a default value for it.

jsuereth: Agreed - There is a default value in the spec, but I believe we're trying to shift away from what it is now.

Thanks for pointing that one out! merging

PR #1063, opened 2025-12-08
Cleanups to v2 DIFF format

jsuereth:
- There was a bug in checking deprecated fields for signals
- Update Attribute + Signal to use the same templated diff function

### DIFF
--- a/crates/weaver_resolved_schema/src/v2/mod.rs
+++ b/crates/weaver_resolved_schema/src/v2/mod.rs
@@ -536,8 +536,8 @@ fn diff_signals_by_hash<T: Signal>(
     latest: &HashMap<&str, &T>,
     baseline: &HashMap<&str, &T>,
 ) -> Vec<SchemaItemChange> {
-    let mut changes = Vec::new();
-    for (signal_id, latest_signal) in latest.iter() {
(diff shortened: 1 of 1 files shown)


crates/weaver_resolved_schema/src/v2/mod.rs
```diff
@@ -577,21 +518,29 @@ pub fn convert_v1_to_v2(
...
+fn diff_signals_by_hash<T: Signal>(
+    latest: &HashMap<&str, &T>,
+    baseline: &HashMap<&str, &T>,
+) -> Vec<SchemaItemChange> {
     let mut changes = Vec::new();
-    let baseline_signals: HashMap<&SignalId, &T> = baseline.iter().map(|s| (s.id(), s)).collect();
-    let latest_signals: HashMap<&SignalId, &T> = latest.iter().map(|s| (s.id(), s)).collect();
-    for (signal_id, _) in latest_signals.iter() {
-        let baseline_signal = baseline_signals.get(signal_id);
+    for (signal_id, latest_signal) in latest.iter() {
```
jerbly: Can we use `&signal_id` here to avoid the double deref in the loop?

jsuereth: done!

PR #1064, opened 2025-12-08
Update-Markdown now supports v2

jsuereth:
This takes an approach where we have guaranteed strings to access portions of the resolved schema:

`registry.` vs. `refinements.` access those portions, then, you can dive deeper via the signal type, e.g. `registry.spans.{id}`, `refinements.events.{id}`

For no good reason, I added `registry.attributes.{id}` if you want to pull an attribute snippet.

### DIFF
--- a/crates/weaver_semconv_gen/README.md
+++ b/crates/weaver_semconv_gen/README.md
@@ -72,7 +72,7 @@ in your current template directory.
 You can specify a different template by adding a `template:{template file}` prefix to your weaver snippet, e.g.
 
 ```md
-<!-- weaver template:my_custom_template.j2 .registry.metrics[] | select(.name = "my_custom_metric") | first -->
+<!-- weaver template:my_custom_template.j2 .signals.metrics[] | select(.name = "my_custom_metric") -->
 my_custom_template will be rendered with "my_custom_metric"'s data, right here.
 <!-- endweaver -->
 ```
--- a/crates/weaver_semconv_gen/data_v2/templates.md
+++ b/crates/weaver_semconv_gen/data_v2/templates.md
@@ -25,3 +25,7 @@ Custom Snippet Name
 <!-- weaver template:registry.md.j2 {value:.registry_url} -->
 todo/1.0.0
 <!-- endweaver -->
+
+<!-- weaver template:registry.md.j2 { value: .signals.metrics[] | select(.name == "test.metric") | .unit } -->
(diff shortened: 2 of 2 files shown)


jsuereth:
@lmolkova PTAL at this again - I was able to implement the JQ idea I mentioned.

crates/weaver_semconv_gen/README.md
```diff
@@ -43,4 +48,31 @@ The template will be passed the following context variables:
...
+<!-- endweaver -->
+```
+
+By default, weaver will send the result of the jq query at the `snipped.md.j2` template file defined
+in your current template directory.
+
+You can specify a different template by adding a `template:{template file}` prefix to your weaver snippet, e.g.
+
+```md
+<!-- weaver template:my_custom_template.j2 .registry.metrics[] | select(.name = "my_custom_metric") | first -->
```
lmolkova: when I run it, I don't see `registry` object in the context,  `{{debug()}}` shows I have `attributes`, `metrics`, etc there without `registry` wrapping them. 

Discussed during the call that it's probably a bug around jq/jinja

jsuereth: Ah, that's because this is the forge registry.  I call it `.signals` and `.refinements` there.  I'll update the docs once I get a working JQ query.

jsuereth: Fixed!

crates/weaver_semconv_gen/data_v2/templates.md
```diff
@@ -0,0 +1,31 @@
+<!-- semconv registry.spans.trace.test -->
```
lmolkova: I really like this syntax, but I can't make it work on semconv

e.g. with `<!-- weaver template:debug.j2 registry.spans.http.client -->` I get 

✖ Filter 'registry.spans.http.client' failed: undefined filter

Diagnostic report:

  × weaver registry update-markdown failed.

----------------------------------

`<!-- weaver template:debug.j2 .signals.spans[] | select(.type == "http.client") -->` works fine though

lmolkova (approved):

crates/weaver_semconv_gen/README.md
jsuereth: Ah I may have missed the one - it should be `.signals` though I had updated it, I'll fix it when it get back.

PR #1066, opened 2025-12-10
Support continuous live-check

jerbly:
Now that we can emit findings as log_records, we can support continuously running live-check. These changes aid this mode of operation:
- `--inactivity-timeout=0` -> If this is set to zero then it never times out
- `--output=none` -> if this is set to none then no template engine is loaded and nothing is rendered out
- `--no-stats` -> if this is set then statistics are not accumulated over the running time of live-check which has the potential to otherwise store a lot of info in memory.

We may want a convenience option to set all these up with one flag but I'm trying not to add more options. These individual tweaks have their own merit independently.

### DIFF
--- a/crates/weaver_live_check/README.md
+++ b/crates/weaver_live_check/README.md
@@ -274,6 +274,14 @@ Each policy finding is emitted as an OTLP log record with the following structur
 - `weaver.finding.signal_type`: Signal type (e.g., "span", "event", "metric")
 - `weaver.finding.signal_name`: Signal name (e.g., event name or metric name)
 
+## Continuous Running Sessions
+
+For long-running live-check sessions (for example, in a staging environment), there are a few settings to consider:
+
+- `--inactivity-timeout=0`: If this is set to zero then weaver never times out.
+- `--output=none`: If this is set to none then no template engine is loaded and nothing is rendered out to the console or files.
+- `--no-stats`: If this is set then statistics are not accumulated over the running time of live-check which has the potential to otherwise store a lot of info in memory.
+
 ## Usage examples
 
 Default operation. Receive OTLP requests and output findings as it arrives. Useful for debugging an application to check for telemetry problems as you step through your code. (ctrl-c to exit, or wait for the timeout)


crates/weaver_live_check/src/lib.rs
```diff
@@ -584,6 +599,9 @@ impl LiveCheckStatistics {
 
     /// Increment the total number of entities by type
     pub fn inc_entity_count(&mut self, entity_type: &str) {
+        if self.disabled {
```
jsuereth: Would it be easier to make this a `dyn` trait and have a "disable" instance and then the real thing and just return the empty instance?

Example:

```rust
pub trait LiveCheckStatistics {
   fn inc_entity_count(...);
   ...
}

struct NoStatistics {}
impl LiveCheckStatistics for NoStatistics { ... }

struct CumualtiveStatistics {}
impl LiveCheckStatistics for CumulativeStatistics { ... }

fn my_live_check_runner() {
  let stats: Box<dyn LiveCheckStatistics> = Box::new(...);
}
```

jerbly: ...or maybe an enum. I'll take a look.

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ All notable changes to this project will be documented in this file.
 - Add Log support for emit and live-check ([#1042](https://github.com/open-telemetry/weaver/pull/1042) by @jerbly)
 - Add OTLP log emission for policy findings in live-check. Whenever a PolicyFinding is created, a log_record is emitted to your configured OTLP endpoint. ([#1045](https://github.com/open-telemetry/weaver/pull/1045) by @jerbly)
 - Deprecate `weaver registry search` command. This command is not compatible with V2 schema and will be removed in a future version. Users should search the generated documentation instead. ([#1057](https://github.com/open-telemetry/weaver/pull/1057) by @jerbly)
+- Add support for continuous live-check sessions with `--inactivity-timeout=0`, `--output=none` and `--no-stats` ([#1066](https://github.com/open-telemetry/weaver/pull/1066) by @jerbly)
```
lmolkova, line 18: not blocking: could you please add a few lines about it in the live-check readme?

jerbly: updated

crates/weaver_live_check/src/stats.rs
```diff
@@ -0,0 +1,400 @@
...
+pub enum LiveCheckStatistics {
+    /// Cumulative statistics mode - tracks all telemetry data
+    Cumulative(CumulativeStatistics),
+    /// Disabled statistics mode - no accumulation (for long-running sessions)
+    Disabled(DisabledStatistics),
+}
+
+impl LiveCheckStatistics {
+    /// Add a live check result to the stats
+    pub fn maybe_add_live_check_result(&mut self, live_check_result: Option<&LiveCheckResult>) {
```
jsuereth, line 275: Nice! I like this division

jsuereth (approved):

PR #1072, opened 2025-12-11
fix clippy on main

zeitlinger:
due to 1.92 release of rust

### DIFF
--- a/crates/weaver_version/src/lib.rs
+++ b/crates/weaver_version/src/lib.rs
@@ -641,8 +641,8 @@ mod tests {
         let mut version = None;
 
         for (v, _) in versions.versions_asc() {
-            if let Some(version) = version.as_ref() {
-                assert!(v > *version);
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -156,6 +156,7 @@ unused_extern_crates = "deny"
 
 unsafe_op_in_unsafe_fn = "warn"
 unused_macro_rules = "warn"
+unused_assignments = "allow"
```
zeitlinger, line 159: I couldn't get it to work with use site:

```rust
    #[allow(unused_assignments)]
    pub src: NamedSource<Cow<'static, str>>,
```

jsuereth: I was unable to get this to work either. I think this is a bug/regression that somehow got fixed / rolled-back prior to the 1.92 release - at least upon my spelunking into rust issue tracker.

PR #1073, opened 2025-12-11
Fix global codeowners

lmolkova:
> If you want to match two or more code owners with the same pattern, all the code owners must be on the same line. If the code owners are not on the same line, the pattern matches only the last mentioned code owner.

https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#codeowners-syntax

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -14,7 +14,7 @@
 
 # Global owners, will be the owners for everything in the repo
 
-* @open-telemetry/weaver-maintainers @open-telemetry/weaver-approvers
+* @open-telemetry/weaver-approvers
 
 # Semantic conventions schema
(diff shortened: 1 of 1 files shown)


CODEOWNERS
```diff
@@ -14,8 +14,7 @@
 
 # Global owners, will be the owners for everything in the repo
 
-* @open-telemetry/weaver-maintainers
-* @open-telemetry/weaver-approvers
+* @open-telemetry/weaver-maintainers @open-telemetry/weaver-approvers
```
lmolkova: we could also do


```suggestion
* @open-telemetry/weaver-approvers
```

since approvers group should (and does) include all maintainers

jsuereth (approved):
I'm good either way here.

PR #1074, opened 2025-12-11
Add vertica to list of words, as it's a product.

jsuereth:
Replaces #1062


lmolkova (approved):
let's test my approver power

PR #1076, opened 2025-12-13
Initial Weaver API and UI

jerbly:
This is the very beginning of an API layer with embedded UI in weaver. At this stage the command is labelled `Experimental` while we continue developing this.

Run `weaver serve` to start the Axum server. Then go to http://127.0.0.1:8080 

image

At the moment you can browse and search the Signals and Attributes, explore the Forge Resolved and Semconv V2 Schemas, and read and try the OpenAPI spec for this new Weaver API (via built-in Rapidoc).

The idea is to add many weaver functions into the UI like:
- A playground for JQ, Rego policies etc.
- All schemas: semconv, live-check-samples etc.
- Live-check playground.
- Hot loadable registries (file picker / git repo picker)
- Registry stats on the Dashboard via the `stats` crate - visualized with graphs and charts
- ...

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,7 +8,7 @@ ARG TARGETPLATFORM
 # renovate: datasource=node-version depName=node
 ARG NODE_VERSION=24
 RUN curl -fsSL https://deb.nodesource.com/setup_${NODE_VERSION}.x | bash - && \
-    apt-get install -y nodejs
+  apt-get install -y nodejs
 
 # Copy UI package files first for better layer caching
 COPY ui/package.json ui/package-lock.json /build/ui/
@@ -27,7 +27,6 @@ COPY data /build/data
 COPY src /build/src
 COPY tests /build/tests
 COPY defaults /build/defaults
-COPY schemas /build/schemas
 COPY cross-arch-build.sh /build/cross-arch-build.sh
 
 # Build weaver


src/serve/handlers.rs
```diff
@@ -0,0 +1,369 @@
...
+        Some(entity) => Json(json!(entity)).into_response(),
+        None => (
+            StatusCode::NOT_FOUND,
+            Json(json!({"error": "Entity not found", "type": entity_type})),
+        )
+            .into_response(),
+    }
+}
+
+/// Search across all types.
```
jsuereth: Overall - I think we want more rustdoc on these APIs here, or some way to denote what this API is over time.  E.g. what are the valid inputs/outputs, what are the expectations.

src/serve/ui.rs
```diff
@@ -0,0 +1,9 @@
+// SPDX-License-Identifier: Apache-2.0
+
+//! Embedded UI assets for the serve command.
+
+use include_dir::{include_dir, Dir};
+
+/// Embedded UI distribution files.
+/// These are built from the ui/ directory using `npm run build`.
+pub(crate) static UI_DIST: Dir<'_> = include_dir!("ui/dist");
```
jsuereth, line 9: I think we'll need a build.rs that will create the UI distribution via npm for this create.

src/serve/types.rs
```diff
@@ -0,0 +1,241 @@
...
+    Span(SpanSearchResult),
+    /// An event result.
+    Event(EventSearchResult),
+    /// An entity result.
+    Entity(EntitySearchResult),
+}
+
+/// Attribute search result.
+#[derive(Debug, Serialize)]
+pub struct AttributeSearchResult {
```
jsuereth: Should we just embedd the Forge-Schema here?

Something like 

```rust
/// The found attribute
[#serde(flatten)]
pub result: Attribute,
```

That way we don't have to keep copy-pasting things here as the model evolves.  E.g. It'd be nice if annotations show up here / get rendered.

src/serve/ui.rs
jerbly: Yes, for sure.

src/serve/types.rs
jerbly: I'll look at that but *SearchResult are just for the list view. When you select the item to get the details it's the Forge Schema item e.g. http://127.0.0.1:8080/api/v1/metrics/http.client.active_requests gets you the full item.

src/serve/handlers.rs
jerbly: 100% - this was just done quickly to see if we liked the idea... I intend to make an OpenAPI schema that is then surfaced in the UI as well.

src/serve/ui.rs
jerbly: I've added the build.rs and updated the github actions. Hence the green build here now.

src/serve/types.rs
jerbly: I did switch this over to using the full struct via an Arc in the ScoredResult.

.github/workflows/ci.yml
```diff
@@ -31,6 +31,14 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
+      - name: Setup Node.js
```
jsuereth, line 34: We probably need this step in Cargo-dist too

jerbly: I think I've done this now, added system dependencies to `dist-workspace.toml` - hard to test though.

src/serve/handlers.rs
jerbly: OpenAPI doc generation and viewer is now included.

schemas/forge.schema.v2.json
jsuereth, line 1: Do we need to submit this file for the whole thing to hang together?

Can we instead just generate this file as part of the build? (note: I was looking into that before I ran out of time for V2 related things)

crates/weaver_forge/Cargo.toml
```diff
@@ -49,6 +49,10 @@ include_dir.workspace = true
 schemars.workspace = true
 dirs.workspace = true
 log.workspace = true
+utoipa = { version = "5.3", optional = true }
```
jsuereth: Nit - Should we put this on the workspace, or is it really only used by forge?

src/serve/handlers.rs
```diff
@@ -0,0 +1,306 @@
...
+        (status = 200, description = "Requested schema", content_type = "application/json"),
+        (status = 404, description = "Schema not found")
+    ),
+    tag = "schemas"
+)]
+pub async fn get_schema(Path(name): Path<String>) -> impl IntoResponse {
+    let name = name.trim_start_matches('/');
+
+    let schema = match name {
+        "forge" => include_str!("../../schemas/forge.schema.v2.json"),
```
jsuereth: Nit: Take a look at `weaver registry schema` command, we can actually generate these strings from Schemars, and I'd suggest we look at doing that.

src/serve/handlers.rs
```diff
@@ -0,0 +1,306 @@
...
+        "semconv" => include_str!("../../schemas/semconv.schema.v2.json"),
```
jsuereth: Second nit, should the names match this: https://github.com/open-telemetry/weaver/blob/main/src/registry/json_schema.rs#L58

src/serve/handlers.rs
```diff
@@ -0,0 +1,306 @@
...
+        (status = 200, description = "Registry overview", body = RegistryOverview)
+    ),
+    tag = "registry"
+)]
+pub async fn registry_overview(State(state): State<Arc<AppState>>) -> impl IntoResponse {
+    let registry = &state.registry;
+
+    let overview = RegistryOverview {
+        registry_url: registry.registry_url.clone(),
+        counts: RegistryCounts {
```
jsuereth, line 91: Nit: Think about re-using the `weaver registry stats` output here.

crates/weaver_search/src/lib.rs
```diff
@@ -0,0 +1,435 @@
...
+use weaver_forge::v2::{
+    attribute::Attribute, entity::Entity, event::Event, metric::Metric,
+    registry::ForgeResolvedRegistry, span::Span,
+};
+use weaver_semconv::stability::Stability;
+
+use super::types::{ScoredResult, SearchResult, SearchType};
+
+/// Search context for performing fuzzy searches across the registry.
+pub struct SearchContext {
```
jsuereth, line 27: This is an idea for the future, but similar to how some UIs work or JAQ, we may want to turn this into a:

```
pub struct SearchContext<'a> {
  items: Vec<SearchableItem<'a>>
}
impl <'a> SearchContext<'a> {
  fn from_registry(registry: &'a ForgeRegistry) -> SearchContext<'a> { ... }
}
```

This could help reduce the memory overhead of copying all the data for search if we hit large / many repositories.  We just need to make sure the out registry's lifetime is protected when we create the SearchContext so lifetime magic all works together.

Dockerfile
```diff
@@ -4,15 +4,22 @@ WORKDIR /build
...
 COPY Cargo.lock /build/Cargo.lock
+COPY build.rs /build/build.rs
 COPY .cargo /build/.cargo
 COPY crates /build/crates
 COPY data /build/data
 COPY src /build/src
 COPY tests /build/tests
 COPY defaults /build/defaults
+COPY schemas /build/schemas
+COPY ui /build/ui
```
jsuereth: Nit - for repeatability reasons in Docker, it'd be nice if we didn't copy *all* the directories here, as the transfer of all the node dependencies may take some time, and it may lead to builds which succeed locally but fail remotely (where the node modules directory doesn't exist in `ui` dir).

That may be annoying here, but it's the same reason I'm not copying `target` into the docker file.

jsuereth (commented):
Overall this looks great!

lots of nits - but my primary concerns are around build / release setup for this.

In order of priority:

- Let's use schemars to generate json schema on demand vs. forcing files to be checked in.
- Let's track dependencies at the workspace if they're used in multiple crates
- For Docker Build - is there a way we can make rennovate aware of NPM version so it can issue autoupgrade PRs for us?   If the nodesource repo is the best bet, that's fine, just wanted to check what we can do.
- For Docker Build (part two) - Can we try to optimise not pulling in `node_modules` or other huge directories via `COPY`?  That can dramatically slow down the build on some architectures.  I didn't check the current docker build on this PR to see how long it takes vs. our current docker build.  If it's not a huge deal right now, we can tweak that aspect later.

I think once those are shored up, we should merge this.  It lays the foundation we need to continue to make progress.

src/serve/handlers.rs
jerbly: Agreed. I actually put a TODO in another place for this exact thing... perhaps we should start issues for API/UI TODOs after the merge?

schemas/forge.schema.v2.json
jerbly: Changed to generate schemas on the fly

src/serve/handlers.rs
jerbly: Changed to generate schemas on the fly

Dockerfile
jerbly: Done

crates/weaver_search/src/lib.rs
jerbly: I'll not do that in this PR. The SearchContext is created once with one clone of the registry items, from that point on they're all Arcs. I attempted to do this with lifetimes but backed-out when it got tough.

jerbly:
@jsuereth - I believe I've covered the key points you raised. Can you take another look? Thanks.

src/serve/handlers.rs
lquerel: We should also be a bit more explicit in the URL, with something like `/api/v1/registry/stats`.

src/serve/handlers.rs
```diff
@@ -0,0 +1,316 @@
...
+        },
+    };
+
+    Json(overview)
+}
+
+/// Get a specific attribute by key.
+#[utoipa::path(
+    get,
+    path = "/api/v1/attribute/{key}",
```
lquerel: Just in case we one day support attributes that are not tied to a registry, and to follow the same hierarchy we use at the CLI level, I suggest using the following path: `/api/v1/registry/attributes/{key}`.

src/serve/handlers.rs
```diff
@@ -0,0 +1,316 @@
...
+            Json(json!({"error": "Attribute not found", "key": key})),
+        )
+            .into_response(),
+    }
+}
+
+/// Get a specific metric by name.
+#[utoipa::path(
+    get,
+    path = "/api/v1/metric/{name}",
```
lquerel: See my comment on attributes.

src/serve/handlers.rs
```diff
@@ -0,0 +1,316 @@
...
+            Json(json!({"error": "Metric not found", "name": name})),
+        )
+            .into_response(),
+    }
+}
+
+/// Get a specific span by type.
+#[utoipa::path(
+    get,
+    path = "/api/v1/span/{type}",
```
lquerel: Same here.

src/serve/handlers.rs
```diff
@@ -0,0 +1,316 @@
...
+            Json(json!({"error": "Span not found", "type": span_type})),
+        )
+            .into_response(),
+    }
+}
+
+/// Get a specific event by name.
+#[utoipa::path(
+    get,
+    path = "/api/v1/event/{name}",
```
lquerel: Same here.

lquerel: Same here.

src/serve/handlers.rs
```diff
@@ -0,0 +1,316 @@
...
+            Json(json!({"error": "Event not found", "name": name})),
+        )
+            .into_response(),
+    }
+}
+
+/// Get a specific entity by type.
+#[utoipa::path(
+    get,
+    path = "/api/v1/entity/{type}",
```
lquerel: Same here.

src/serve/handlers.rs
```diff
@@ -0,0 +1,316 @@
...
+            Json(json!({"error": "Entity not found", "type": entity_type})),
+        )
+            .into_response(),
+    }
+}
+
+/// Search across all types or list all items (browse mode).
+#[utoipa::path(
+    get,
+    path = "/api/v1/search",
```
lquerel: If this search is only intended to target objects within a registry, then I think we should use the path `/api/v1/registry/search`.

lquerel: If this search is only intended to target objects within a registry, then I think we should use the path `/api/v1/registry/search`.

crates/weaver_search/src/lib.rs
```diff
@@ -0,0 +1,435 @@
...
+///
+/// Scoring weights:
+/// - Exact name/key match: 100 points
+/// - Name/key starts with query: 80 points
+/// - Name/key contains query: 70 points
+/// - All query tokens found in name: 60 points
+/// - Brief contains query: 40 points
+/// - Note contains query: 20 points
+/// - Deprecated items: score divided by 10 (heavily demoted)
+fn score_match(query: &str, item: &SearchableItem) -> u32 {
```
lquerel, line 286: A long time ago, this project used Tantivy ([https://github.com/quickwit-oss/tantivy](https://github.com/quickwit-oss/tantivy)) for a TUI that essentially provided the same capabilities as this web UI. This is not for this PR, but I think it would be beneficial to add Tantivy back to our dependencies to get a more modern full-text search with scoring, fuzzy search, and similar features.

An example of this integration is visible here: https://github.com/open-telemetry/weaver/tree/586df7c6652be4f6c6398b0469fb2a6ad25eb79d/src/search

jerbly: This looks like a good crate and it's recently updated. Another item for the todo list for the UI.

jerbly:
> Great work! I mainly made comments on the organization of the URL paths to make future evolutions easier.

@lquerel - thanks for the review, I have reorganized the URLs and handlers to group them under `registry` and added a TODO for Tantivy.

Dockerfile
```diff
@@ -4,15 +4,30 @@ WORKDIR /build
...
+COPY Cargo.toml Cargo.lock build.rs /build/
 COPY .cargo /build/.cargo
+
+# Copy source files
 COPY crates /build/crates
 COPY data /build/data
 COPY src /build/src
 COPY tests /build/tests
 COPY defaults /build/defaults
+COPY schemas /build/schemas
```
jsuereth: Is this still needed now that you serve `json_schema!` macros?

jsuereth (approved):
One last nit

Dockerfile
jerbly: oops, thanks, missed that one.

PR #1085, opened 2025-12-17
Document generated file header best practice

### LINKED ISSUE #440: Add Comment in Generated Code Indicating Source and Generated Status as Best Practice in Jinja Templates
As @marcalff mentioned in the following comment https://github.com/open-telemetry/weaver/issues/227#issuecomment-2450408270 it would be a good practice to systematically add a comment in generated code to indicate:

1.	That it is generated code.
2.	Where it was generated from.

This practice should be strongly recommended as a best practice in the Jinja templates.

cnaples79:
## Summary
- add guidance to code generation docs to include a header comment noting generated files and their source template

## Rationale
- issue #440 requests a best practice to mark generated code with its source; documenting it makes templates easier to audit and discourages manual edits

## Testing
- not run (docs-only change)

Fixes #440

### DIFF
--- a/docs/codegen.md
+++ b/docs/codegen.md
@@ -111,6 +111,10 @@ This output in turn is passed to the `metrics.md.j2` template, evaluated by the 
 
 ## Tips and Tricks
 
+#### Mark generated files
+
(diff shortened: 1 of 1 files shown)


lmolkova:
Thanks for the contribution @cnaples79 !

Closing this PR due to inactivity, but you're welcome to come back to it.

PR #1108, opened 2026-01-03
Replace regex with semver crate

### LINKED ISSUE #579: Replace the regex in `from_semconv_specs` with the `https://docs.rs/semver/latest/semver/` crate.
In the `weaver_semconv/src/registry.rs` method `from_semconv_specs` (see todo), we should use: https://docs.rs/semver/latest/semver/ and URL parser that can give us the last element of the path to send to the parser.

See this PR comment: https://github.com/open-telemetry/weaver/pull/400/#discussion_r1905412697

ArthurSens:
Fixes #579 

Trying to learn Rust with some good first issues :)

### DIFF
--- a/crates/weaver_semconv/src/registry.rs
+++ b/crates/weaver_semconv/src/registry.rs
@@ -10,11 +10,36 @@ use crate::schema_url::SchemaUrl;
 use crate::semconv::{SemConvSpecV1WithProvenance, SemConvSpecWithProvenance};
 use crate::stats::Stats;
 use crate::Error;
-use regex::Regex;
 use std::collections::HashMap;
 use std::path::Path;
-use std::sync::LazyLock;
 use weaver_common::result::WResult;
+
+/// Extracts a semantic version from a path string.
+///
+/// This function searches through path segments (separated by '/') looking for
+/// a segment that starts with 'v' followed by a valid semantic version.
+///
+/// # Arguments
+///
+/// * `path` - A path string that may contain a version segment (e.g., "/path/v1.26.0/file")
+///
+/// # Returns
+///
+/// * `Some(String)` - The version string including the 'v' prefix (e.g., "v1.26.0")
+/// * `None` - If no valid semantic version is found in the path
+fn extract_semconv_version(path: &str) -> Option<String> {
+    for segment in path.split('/') {
+        if segment.starts_with('v') && segment.chars().nth(1).is_some_and(|c| c.is_ascii_digit()) {
+            // Try to parse it with semver (strip the 'v' first)
+            if let Some(version_part) = segment.strip_prefix('v') {
+                if semver::Version::parse(version_part).is_ok() {
+                    return Some(segment.to_owned());
+                }
+            }
+        }
+    }
+    None
+}
 
 /// A semantic convention registry is a collection of semantic convention
 /// specifications indexed by group id.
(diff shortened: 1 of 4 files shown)


crates/weaver_semconv/src/registry.rs
```diff
@@ -137,16 +158,11 @@ impl SemConvRegistry {
...
             // No registry manifest found.
-            // Try to infer the manifest from the registry path by detecting the
-            // presence of the following pattern in the registry path: v\d+\.\d+\.\d+.
-            if let Some(captures) = VERSION_REGEX.captures(registry_repo.registry_path_repr()) {
-                if let Some(captured_text) = captures.get(1) {
-                    semconv_version = captured_text.as_str().to_owned();
-                }
-            }
+            // Try to infer the manifest from the registry path by detecting a
+            // valid semantic version segment (e.g., v1.26.0) in the path.
```
ArthurSens, line 169: I've changed the comment here since semver accepts more than just `v\d+\.\d+\.\d+`, e.g. release candidates

crates/weaver_semconv/src/registry.rs
```diff
@@ -12,12 +12,37 @@ use crate::registry_repo::RegistryRepo;
...
+///
+/// # Arguments
+///
+/// * `path` - A path string that may contain a version segment (e.g., "/path/v1.26.0/file")
+///
+/// # Returns
+///
+/// * `Some(String)` - The version string including the 'v' prefix (e.g., "v1.26.0")
+/// * `None` - If no valid semantic version is found in the path
+fn extract_semconv_version(path: &str) -> Option<String> {
```
jsuereth, line 44: My only concern here (and also one with the previous regex) is that the `v` is not required, see: https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/schemas#schema-url

ArthurSens: Hmmmm, gotcha. Reading the link you shared here, it also says that version is always the last part of the URL, so if I'm understanding correctly, I don't need to try to parse every segment and check if it's a valid semver... just the last segment is enough, right?

lquerel: I don't think the documentation on `schema_url` is relevant here, because in this context Weaver does not derive a version from a `schema_url`, but from a semantic conventions registry (local, Git, or archive). For recent registries, Weaver retrieves the version directly from the manifest file, but for older registries this manifest does not exist, and in that case the version is derived from the archive path, for example. See the code of the `from_semconv_specs`.

crates/weaver_semconv/src/registry.rs
```diff
@@ -447,4 +463,56 @@ mod tests {
...
     }
+
+    #[test]
+    fn test_extract_semconv_version() {
+        use super::extract_semconv_version;
+
+        // Test with GitHub release URL pattern
+        assert_eq!(
+            extract_semconv_version(
+                "https://github.com/open-telemetry/semantic-conventions/releases/download/v1.26.0/semantic-conventions.zip"
```
lquerel, line 494: The SemConv archive URLs have the following format. We should test with them. 

https://github.com/open-telemetry/semantic-conventions/archive/refs/tags/v1.38.0.zip
https://github.com/open-telemetry/semantic-conventions/archive/refs/tags/v1.38.0.tar.gz

lquerel (changes requested):
We should change the tests to validate with the real archive URLs.

crates/weaver_semconv/src/registry.rs
jsuereth: I agree we need to be flexible.  What I'm saying is that *eventually* I'd like for SchemaURL to work.  Right now it's forcing the `v` as required, I think we should allow it as optional.

crates/weaver_semconv/src/registry.rs
ArthurSens: Do we have a list of file format suffixes that Weaver allows? Semver won't be able to parse those suffixes, so I believe I'll have to manually remove this suffixes before parsing the version

crates/weaver_semconv/src/registry.rs
ArthurSens: I'm a bit lost here if in this PR we want to solve `v` being optional or not 😅

jsuereth: If it wasn't optional before, you can leave it as is for now, I'll clean that up in a follow up.

For context - I'm working on new resolution specification where we can resolve "resolved schema" directly from a Schema URL - so I'll need to either merge in your PR and fix in my branch or have you fix it so my local fix doesn't get blown away.

I'd rather merge yours in and fix later if you're nervous about the amount of rust.

Issue #798, opened 2025-06-20
Merge TallyCat into Weaver for Telemetry Schema Inference and Metadata Governance


nicolastakashi:
Hi all 👋

Following a great discussion with the Weaver maintainers, I’d like to propose merging [TallyCat](https://github.com/nicolastakashi/tallycat) into the Weaver project to provide native support for runtime telemetry schema inference, registry persistence, and metadata governance.

---

## 🔍 What is TallyCat?

TallyCat is a schema-first metadata catalog and governance layer for OpenTelemetry signals—starting with metrics and expanding to logs and traces.

It’s inspired by [OpenMetadata](https://open-metadata.org/), but focused 100% on telemetry. Its core capabilities include:

- **Live Schema Inference**  
  Receives full OTLP signals directly (no custom connector needed) and infers telemetry schemas in real time.

- **Registry Persistence**  
  Inferred schemas are written to a local DuckDB instance for fast querying, historical tracking, and versioning.

- **Web UI Registry Browser**  
  A simple, performant UI to search, navigate, and edit the inferred telemetry schema catalog.

---

## 💡 Long-Term Vision

TallyCat is designed to evolve into a full observability metadata governance control plane. Future goals include:

- **Data Lineage for Observability**  
  Connecting attributes to dashboards, alerts, queries, and service ownership.

- **Telemetry Schema Builders**  
  Allowing users to define new schemas via the UI and publish them declaratively.

- **Advanced Registry Usability**  
  Help users:
  - Choose consistent units and naming
  - Find which services produce specific telemetry
  - Understand schema variants across environments

- **Policy-Driven Validation and Control**  
  Define schema policies that can be:
  - Deployed to collectors
  - Used to block or reroute telemetry pipelines
  - Enforce schema contracts in CI/CD or GitOps workflows

---

## 🧩 Proposed Integration with Weaver

| Capability                      | Suggested Integration Path        | Description |
|--------------------------------|-----------------------------------|-------------|
| Schema inference from OTLP     | `weaver registry live-infer`      | TallyCat extracts schemas directly from OTLP signals at runtime |
| Registry Web UI                | `weaver registry browse` or `weaver registry ui` | Enables interactive schema exploration and editing |
| Schema registry persistence    | DuckDB backend (pluggable)        | Can evolve into multi-backend support in the future |

---

## 🛠 Tech Stack

TallyCat is built using modern, fast, and developer-friendly tooling:

### Backend
- **Language**: Go (Golang)
[...]

PR #1113, opened 2026-01-05
MCP Server

jerbly:
weaver-mcp

This PR adds `weaver registry mcp` an MCP (Model Context Protocol) server that exposes the semantic conventions registry to LLMs. This enables natural language queries for finding and understanding conventions while writing instrumentation code.

| Tool | Description |
|------|-------------|
| `search` | Search across all registry items (attributes, metrics, spans, events, entities) |
| `get_attribute` | Get detailed information about a specific attribute by key |
| `get_metric` | Get detailed information about a specific metric by name |
| `get_span` | Get detailed information about a specific span by type |
| `get_event` | Get detailed information about a specific event by name |
| `get_entity` | Get detailed information about a specific entity by type |
| `live_check` | Validate telemetry samples against the registry |

### DIFF
--- a/crates/weaver_mcp/src/service.rs
+++ b/crates/weaver_mcp/src/service.rs
@@ -2,8 +2,8 @@
 
 //! MCP service implementation using rmcp SDK.
 //!
-//! This module provides the `WeaverMcpService` which implements all 7 tools
-//! for querying and validating against the semantic convention registry.
+//! This module provides the `WeaverMcpService` which implements all tools
+//! for querying, debugging and validating against the semantic convention registry.
 
 use std::path::PathBuf;
 use std::sync::Arc;
@@ -30,7 +30,7 @@ use crate::McpConfig;
 
 /// MCP service for the semantic convention registry.
 ///
-/// This service exposes 7 tools for querying and validating against the registry:
+/// This service exposes tools for querying, debugging and validating against the registry:
 /// - `search` - Search across all registry items
 /// - `get_attribute` - Get a specific attribute by key
 /// - `get_metric` - Get a specific metric by name


crates/weaver_mcp/src/service.rs
```diff
@@ -0,0 +1,846 @@
...
+/// This service exposes 7 tools for querying and validating against the registry:
+/// - `search` - Search across all registry items
+/// - `get_attribute` - Get a specific attribute by key
+/// - `get_metric` - Get a specific metric by name
+/// - `get_span` - Get a specific span by type
+/// - `get_event` - Get a specific event by name
+/// - `get_entity` - Get a specific entity by type
+/// - `live_check` - Validate telemetry samples against the registry
+#[derive(Clone)]
+pub struct WeaverMcpService {
```
jsuereth, line 42: I know this may be crazy, but is there any chance we can re-use the vanilla API server and just "wrap" API methods with MCP definitions instead of having two instances of serving / application flow?

I think there's a LOT of commonality between the two pieces, and it'd be ideal if we could share it.

jerbly: Wrapping a web api with mcp is an anti-pattern. I did move all the internals of search and get* into the weaver_search crate to make it common. I think it's likely they'll diverge, for example I'm experimenting with multiple tools for live-check.

jsuereth: > Wrapping a web api with mcp is an anti-pattern.

Interesting - tell me more?

I'm suggesting MCP is basically API++.  Where we can delegate to API functions we should.  If you add *more* that's great.

E.g. I'd be nice if we had a consistent "AppContext" structure that has our `Arc<State>` in it and a set of methods on this that both the MCP + API server would use.

I'm *not* suggesting we directly call the API from MCP - just that we take small amount of time to but in a shared abstraction they can both use to interact with a 'served weaver' thing.

jerbly: Actually @lquerel first pointed this out to me in this comment https://github.com/open-telemetry/weaver/issues/798#issuecomment-3025205231 where I had a similar idea to you. Subsequently I have read this same advice quite a few times. The video link is good, there's also a good reddit thread if I recall correctly.

Did you see that I moved the common internals to the weaver_search crate?

jsuereth: Thanks for the link! I'll take a look.

> Did you see that I moved the common internals to the weaver_search crate?

I did not, but that's great!

I was actually thinking of naming the create `weaver_debugger` or some such - i.e. if I were to add "jq", "rego" or "jinja" debugging tools, what crate should that live in?

jerbly: `weaver_devtools`??

crates/weaver_mcp/src/service.rs
```diff
@@ -0,0 +1,846 @@
...
+    DisabledStatistics, LiveCheckRunner, LiveCheckStatistics, Sample, VersionedRegistry,
+};
+use weaver_search::{SearchContext, SearchType};
+use weaver_semconv::stability::Stability;
+
+use crate::McpConfig;
+
+/// MCP service for the semantic convention registry.
+///
+/// This service exposes 7 tools for querying and validating against the registry:
```
jsuereth: Nit - let's update this to say `exposes tools for querying, debugging and validating against the registry.  For example:`

src/registry/mcp.rs
```diff
@@ -0,0 +1,82 @@
...
+pub struct RegistryMcpArgs {
+    /// Registry arguments.
+    #[command(flatten)]
+    pub registry: RegistryArgs,
+
+    /// Diagnostic arguments.
+    #[command(flatten)]
+    pub diagnostic: DiagnosticArgs,
+
+    /// Advice policies directory. Set this to override the default policies.
```
jsuereth, line 30: My only remaining hangup on this PR is these two `advice_*` arguments.  Basically I think our current "when to load rego, jq + jinja" story is a bit off.  It's unclear when `weaver.yml` controls it, when you pull from the same directory as a registry and when you specify via the command line.

I think that's bigger than the scope of this PR, but let's sort that out going forward.

jsuereth (approved):
Thanks for all the cleanups!

I think this is good to merge, but we should track some work to make it clear how "extensions" are loaded in weaver, e.g. Rego policies.

crates/weaver_mcp/src/service.rs
jsuereth: Oooh - I like that better!

src/registry/mcp.rs
jerbly: I agree, also how to specify builtins to use AND custom, or when to override the builtin with a custom. For example, often I want all the otel policies plus my own on top.

Issue #1126, opened 2026-01-10
Allow suppression of warnings for dependencies.


michaelvanstraten:
When using registry dependencies in `registry_manifest.yaml`, those dependencies are resolved and warnings from their specs are merged into the dependent registry’s warnings. This results in hard-to-parse output from tools such as `weaver registry check` because, especially when using `https://github.com/opentelemetry/semantic-conventions`, there are many deprecation warnings.

It would be helpful to add an additional property in `registry_manifest.yaml` to allow suppressing warnings on a per-dependency basis. Additionally, a global CLI or config-file flag to disable all third-party warnings would be useful.

### Example

```yaml
# registry_manifest.yaml
dependencies:
  - name: otel-semconv
    uri: https://github.com/opentelemetry/semantic-conventions.git
    ref: v1.26.0
    suppress_warnings: true
```

An alternative would be to display warnings only when the corresponding signal/attribute is actually used. However, this would likely be more difficult to implement. I’ve attached a screenshot of the warnings below.

These warnings are coming from `https://github.com/opentelemetry/semantic-conventions`, even though I am not using any of the referenced attributes.

<img width="2122" height="741" alt="Image" src="https://github.com/user-attachments/assets/b9c019f6-a207-4033-b7b0-b76ecf98c94e" />

<sub>**Tip**: [React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

PR #1136, opened 2026-01-13
Rework Resolution Engine to allow using resolved schema

jsuereth:
Reworks the resolution engine to allow using *resolved* schema as a dependency.

- Breaks phases formally into `Loaded` and `Resolved`
- `Loaded` can now work with a loaded `ResolvedTelemetrySchema` (v1 or v2) or a definition schema (raw / unresolved).
- Resolve now takes loaded, and will recurse down dependencies as needed.

TODOS

- [x] Remove legacy resolve code from code base
- [x] Handle uniqueness of group / signal ids across dependencies
- [x] More tests

### DIFF
--- a/crates/weaver_resolver/src/dependency.rs
+++ b/crates/weaver_resolver/src/dependency.rs
@@ -184,6 +184,8 @@ trait UnresolvedAttributeLookup {
 
 impl UnresolvedAttributeLookup for V1Schema {
     fn lookup_group_attributes(&self, id: &str) -> Option<Vec<UnresolvedAttribute>> {
+        // TODO - We should try to reconstruct a map which can do the lookup of dependencies.
+        // This likely involves a different algorithm where we can allocate lookup hashes per-resolved repository.
         self.group(id).map(|g| {
             let attributes: Vec<UnresolvedAttribute> = g
                 .attributes
@@ -223,6 +225,10 @@ impl UnresolvedAttributeLookup for V2Schema {
 
 impl UnresolvedAttributeLookup for Vec<ResolvedDependency> {
     fn lookup_group_attributes(&self, id: &str) -> Option<Vec<UnresolvedAttribute>> {
+        // TODO - this algorithm is only viable when we know there's only one dependency.
+        // Going forward we need to allow this method to find ambiguous imports and
+        // issue an error statement that allows resolving the ambiguity by using a
+        // dependency reference, e.g. `dep#id` vs just `id`.  Details TBD.
         self.iter().find_map(|d| d.lookup_group_attributes(id))
     }
 }
--- a/crates/weaver_resolver/src/loader.rs
+++ b/crates/weaver_resolver/src/loader.rs
@@ -23,7 +23,6 @@ const MAX_DEPENDENCY_DEPTH: u32 = 10;
 pub enum LoadedSemconvRegistry {
     /// The semconv repository was unresolved and needs to be run through resolution.
     Unresolved {
-        // TODO - We need the imports here.
         /// The specification of this raw repository.
         repo: RegistryRepo,
         /// The raw definition schema for this repository.
@@ -196,13 +195,14 @@ fn load_semconv_repository_recursive(
     // Either load a fully resolved repository, or read in raw files.
     if let Some(manifest) = registry_repo.manifest() {
         if let Some(resolved_url) = manifest.resolved_schema_url.as_ref() {
-            todo!("Loading resolved telemetry schema ({resolved_url}) is currently unsupported.");
+            todo!("Loading resolved telemetry schema ({resolved_url}) is currently unsupported.\nSee https://github.com/open-telemetry/weaver/issues/1134");
         } else {
             if manifest.dependencies.len() > 1 {
                 todo!("Multiple dependencies is not supported yet.")
             }
             // Load dependencies.
             let mut loaded_dependencies = vec![];
+            let mut non_fatal_errors = vec![];
             for d in manifest.dependencies.iter() {
                 match RegistryRepo::try_new(&d.name, &d.registry_path) {
                     Ok(d_repo) => {
@@ -215,8 +215,10 @@ fn load_semconv_repository_recursive(
                             dependency_chain,
                         ) {
                             WResult::Ok(d) => loaded_dependencies.push(d),
-                            // TODO - Should we always ignore warnings on loaded dependencies?
-                            WResult::OkWithNFEs(d, _) => loaded_dependencies.push(d),
+                            WResult::OkWithNFEs(d, nfes) => {
+                                loaded_dependencies.push(d);
+                                non_fatal_errors.extend(nfes);
+                            }
                             WResult::FatalErr(err) => return WResult::FatalErr(err),
                         }
                     }
@@ -224,7 +226,9 @@ fn load_semconv_repository_recursive(
                 }
             }
             // Now load the raw repository.
+            // TODO - Allow ignoring dependency warnings - https://github.com/open-telemetry/weaver/issues/1126.
             load_definition_repository(registry_repo, follow_symlinks, loaded_dependencies)
+                .extend_non_fatal_errors(non_fatal_errors)
         }
     } else {
         // This is a raw repository with *no* manifest.
--- a/crates/weaver_common/src/result.rs
+++ b/crates/weaver_common/src/result.rs
@@ -72,6 +72,21 @@ where
         }
     }
 
+    /// Extends a WResult with additional non-fatal errors.
+    ///
+    /// If the result was `Ok``, this moves to an `OkWithNFEs`.
+    /// If the result was a `FatalError` this method is ignored.
+    pub fn extend_non_fatal_errors(self, non_fatal_errors: Vec<E>) -> Self {
+        match self {
+            WResult::Ok(result) => WResult::OkWithNFEs(result, non_fatal_errors),
+            WResult::OkWithNFEs(result, mut items) => {
+                items.extend(non_fatal_errors);
+                WResult::OkWithNFEs(result, items)
+            }
+            WResult::FatalErr(e) => Self::FatalErr(e),
+        }
+    }
+
     /// Creates a new [`WResult`] with a successful result.
     pub fn with_non_fatal_errors(result: T, non_fatal_errors: Vec<E>) -> Self {
         if non_fatal_errors.is_empty() {
@@ -199,7 +214,7 @@ mod tests {
     use miette::Diagnostic;
     use serde::Serialize;
 
-    #[derive(thiserror::Error, Debug, PartialEq, Serialize, Diagnostic)]
+    #[derive(thiserror::Error, Debug, PartialEq, Serialize, Diagnostic, Clone)]
     enum TestError {
         #[error("Warning")]
         #[diagnostic(severity(Warning))]
@@ -214,6 +229,30 @@ mod tests {
         fn compound(errors: Vec<TestError>) -> Self {
             TestError::CompoundError(errors)
         }
+    }
+
+    #[test]
+    fn test_extend_nfes() -> Result<(), TestError> {
+        let warnings = vec![TestError::Warning];
+        let result: WResult<i32, TestError> =
+            WResult::Ok(42).extend_non_fatal_errors(warnings.clone());
+        match result {
+            WResult::OkWithNFEs(value, nfes) => {
+                assert_eq!(value, 42);
+                assert_eq!(nfes, warnings);
+            }
+            _ => panic!("Failed to add warning to Ok"),
+        }
+        let result2: WResult<i32, TestError> = WResult::OkWithNFEs(42, vec![TestError::Warning])
+            .extend_non_fatal_errors(warnings.clone());
+        match result2 {
+            WResult::OkWithNFEs(value, nfes) => {
+                assert_eq!(value, 42);
+                assert_eq!(nfes, vec![TestError::Warning, TestError::Warning]);
+            }
+            _ => panic!("Failed to add warning to OkWithNFEs"),
+        }
+        Ok(())
     }
 
     #[test]


crates/weaver_forge/src/v2/registry.rs
```diff
@@ -18,7 +18,7 @@ use crate::{
 
 /// A resolved semantic convention registry used in the context of the template and policy
 /// engines.
-///
+///registry_manifest: None,
```
lmolkova: nit

```suggestion
///
```

crates/weaver_resolved_schema/src/v2/mod.rs
```diff
@@ -55,7 +59,8 @@ pub struct ResolvedTelemetrySchema {
     pub registry: Registry,
     /// Refinements for the registry
     pub refinements: Refinements,
-    // TODO - versions, dependencies and other options.
+    /// The manifest of the registry.
+    pub registry_manifest: Option<RegistryManifest>,
```
lmolkova, line 64: it shows up in the serialized resolved schema and in the JSON schema, but I don't think it should - it's already part of schema_url

crates/weaver_resolver/src/attribute.rs
```diff
@@ -63,28 +70,35 @@ impl AttributeCatalog {
...
+        let gc_map: HashMap<AttributeRef, AttributeRef> = ordered
+            .iter()
+            .map(|(_, attr_ref)| {
                 let new_attr_ref = AttributeRef(next_id);
                 next_id += 1;
                 (*attr_ref, new_attr_ref)
             })
             .collect();
-
+        // assert_eq!(gc_map.len(), self.attribute_refs.len());
```
lmolkova: intended?

crates/weaver_resolver/src/attribute.rs
```diff
@@ -241,6 +263,69 @@ impl AttributeCatalog {
...
+                    group_id: g.id.to_owned(),
+                })
+        });
+        result
+    }
+}
+
+impl AttributeLookup for V2Schema {
+    fn lookup_attribute(&self, key: &str) -> Option<AttributeWithGroupId> {
+        todo!("Lookup {key} on v2 schema.")
```
lmolkova, line 324: is it not used yet?

crates/weaver_semconv/src/manifest.rs
```diff
@@ -141,8 +160,8 @@ mod tests {
             RegistryManifest::try_from_file("tests/test_data/valid_semconv_registry_manifest.yaml")
                 .expect("Failed to load the registry configuration file.");
         assert_eq!(config.name, "vendor_acme");
-        assert_eq!(config.semconv_version, "0.1.0");
-        assert_eq!(config.schema_base_url, "https://acme.com/schemas/");
+        assert_eq!(config.version, "0.1.0");
+        assert_eq!(config.repository_url, "https://acme.com/schemas/");
```
lmolkova, line 164: super-nit, just to make it closer to real use-case (I use tests as a source of documentation a lot)

```suggestion
        assert_eq!(config.repository_url, "https://github.com/acme/semantic-conventions/");
```

crates/weaver_semconv/src/stats.rs
```diff
@@ -16,6 +16,4 @@ pub struct Stats {
     pub group_breakdown: HashMap<GroupType, usize>,
     /// Number of attributes.
     pub attribute_count: usize,
-    /// Number of metrics.
-    pub metric_count: usize,
```
lmolkova, line 20: not blocking, but curious - why is it relevant in scope of this PR?

crates/weaver_resolver/src/registry.rs
```diff
@@ -71,23 +72,45 @@ pub struct UnresolvedGroup {
...
+        .map(group_from_spec)
+        .collect();
+    let mut ureg = UnresolvedRegistry {
+        registry: Registry::new(repo.registry_path_repr()),
+        groups,
+        imports,
+        dependencies,
+    };
 
+    // Now we to the resolution.
```
lmolkova:
```suggestion
    // Now we do the resolution.
```

crates/weaver_resolved_schema/src/v2/mod.rs
lmolkova, line 53: nit, could we skip optional fields ?

Here's how it currently looks like when running this branch with `registry resolve  -r .\model\  --v2 --format yaml` on semconv

```yaml
file_format: resolved/2.0.0
schema_url: ''
registry_id: main // i'll fix it separately in my PR
registry_manifest: null
...  
```

lmolkova: also, feel free to leave as is, and I'll polish in my docs PR

crates/weaver_resolver/src/loader.rs
```diff
@@ -0,0 +1,468 @@
...
+        });
+    }
+    // Add current registry to visited set and dependency chain
+    let _ = visited_registries.insert(registry_id.clone());
+    dependency_chain.push(registry_id.clone());
+
+    // Either load a fully resolved repository, or read in raw files.
+    if let Some(manifest) = registry_repo.manifest() {
+        if let Some(resolved_url) = manifest.resolved_schema_url.as_ref() {
+            todo!("Loading resolved telemetry schema ({resolved_url}) is currently unsupported.");
```
lmolkova: I stumbled upon it when I tried running it against semconv  to see what happens, I'm fine doing it as a follow up, but also, shouldn't it be trivial now? Just deserialize resolved schema?

crates/weaver_resolved_schema/src/v2/mod.rs
jsuereth: Interesting - Yeah. I'll have to figure out where to move this then.

crates/weaver_resolver/src/attribute.rs
jsuereth: This was debugging, I'll remove this.

crates/weaver_resolver/src/attribute.rs
jsuereth: It is not.  That's a follow up.  We don't have the ability to specify a resolved schema to load.

This just sets up the possibility and baseline tests.

Effectively, now, instead of resolving raw repositories by throwing all groups into one bundle, we resolve dependencies independently and use the `ResolvedTelemetrySchema` (V1) as a dependency.  Next step is the ability to *load* these resolved schemas instead of only allowing pointing at raw repositories.

crates/weaver_semconv/src/stats.rs
jsuereth: Oh, I went through and removed things that weren't used/needed from the resolution algorithm.

You're right I could have done that separately.  Sometimes when you're deep refactoring, you get lost :)

crates/weaver_resolved_schema/src/v2/mod.rs
jsuereth: Yeah - that's just an oversight.  Let's fix that in follow up PR.

crates/weaver_resolver/src/loader.rs
jsuereth: Not quite.

My plan was to *first* check the `version` or `file_format` of the file, then delegate to the appropriate loader.

We don't have an `EitherV1OrV2` struct we can parse right now, and given the hell the definition schema version causes, I wanted to avoid that.

jsuereth: Can we leave this as a follow up?  I'd like to get this one in (which is quite hefty) so we can just focus that on on loading repositories of different file formats from URLs strings :)

lmolkova (approved):

crates/weaver_resolved_schema/src/v2/mod.rs
lmolkova: maybe just `#[serde(skip_serializing)]`?

jsuereth: Added

crates/weaver_resolver/src/attribute.rs
jsuereth: Removed

crates/weaver_resolver/src/attribute.rs
```diff
@@ -63,28 +70,34 @@ impl AttributeCatalog {
...
-        let gc_map: HashMap<AttributeRef, AttributeRef> = self
+        let mut ordered: Vec<(attribute::Attribute, AttributeRef)> = self
             .attribute_refs
-            .values()
-            .map(|attr_ref| {
+            .iter()
+            .map(|(a, ar)| (a.clone(), *ar))
+            .collect();
+        ordered.sort_by(|(ln, _), (rn, _)| ln.cmp(rn));
+        // assert_eq!(ordered.len(), self.attribute_refs.len());
```
lquerel, line 85: The kind of thing that could use `debug_assert_eq!`.

crates/weaver_resolver/src/attribute.rs
```diff
@@ -241,6 +262,69 @@ impl AttributeCatalog {
...
+        match self {
+            ResolvedDependency::V1(schema) => schema.lookup_attribute(key),
+            ResolvedDependency::V2(schema) => schema.lookup_attribute(key),
+        }
+    }
+}
+
+impl AttributeLookup for V1Schema {
+    fn lookup_attribute(&self, key: &str) -> Option<AttributeWithGroupId> {
+        // TODO - fast lookup, not a table scan...
```
lquerel, line 287: I thought we originally had a map. Why did we lose that? It;s possible my memory is failing me.

crates/weaver_resolver/src/lib.rs
```diff
@@ -2,583 +2,158 @@
...
                 }
                 WResult::FatalErr(e) => return WResult::FatalErr(e),
             }
         }
-
-        WResult::OkWithNFEs(specs, non_fatal_errors)
+        let registry_id: String = repo.id().to_string();
+        let manifest = repo.manifest().cloned();
+        let mut attr_catalog = AttributeCatalog::default();
+        // TODO - Do something with non_fatal_errors if we need to.
```
lquerel, line 94: Why not merge the non fatal errors coming from the dependencies with the `WResult` returned by this method? It feels like, at the moment, we're simply dropping those errors.

crates/weaver_resolver/src/dependency.rs
```diff
@@ -0,0 +1,334 @@
...
+            })
+        };
+
+        // Filter imports to only include those from the current registry
+        let current_registry_imports: Vec<_> = imports.iter().collect();
+
+        let metrics_imports_matcher = build_globset(
+            current_registry_imports
+                .iter()
+                .find_map(|i| i.imports.metrics.as_ref()),
```
lquerel, line 70: Can't we have imports across multiple files? With `find_map` here, we’re only finding the first one.

crates/weaver_resolver/src/loader.rs
```diff
@@ -0,0 +1,468 @@
...
+                        // so we need to make sure the dependency chain only include direct dependencies of each other.
+                        match load_semconv_repository_recursive(
+                            d_repo,
+                            follow_symlinks,
+                            max_dependency_depth - 1,
+                            visited_registries,
+                            dependency_chain,
+                        ) {
+                            WResult::Ok(d) => loaded_dependencies.push(d),
+                            // TODO - Should we always ignore warnings on loaded dependencies?
```
lquerel: I don't think we should ignore non-fatal errors on loaded dependencies. We don't really have any guaranties that they are well-formed.

crates/weaver_resolver/src/attribute.rs
jsuereth: Ordering.  You switched to a linear order for serialization.  

I'll need to lead a map on the side.

jsuereth: Ordering.  You switched to a linear order for serialization.  

I'll need to lead a map on the side.

crates/weaver_resolver/src/lib.rs
jsuereth: We are, there's a bug on this being the preferred default.

I can write in the flag that will allow ignoring/dropping or preserving, but I'm going to default to ignoring.

The issue at least before, was using current rules and rego policies on older repositories.  That may be less of an issue pulling resolved registry but it's still not as useful as you'd think

jsuereth: We are, there's a bug on this being the preferred default.

I can write in the flag that will allow ignoring/dropping or preserving, but I'm going to default to ignoring.

The issue at least before, was using current rules and rego policies on older repositories.  That may be less of an issue pulling resolved registry but it's still not as useful as you'd think

crates/weaver_resolver/src/dependency.rs
jsuereth: Yeah.  This is a temporary quick implementation while we only support one.  This method will need to be fleshed out to support multiple later, but at least it's contained in one spot now.

jsuereth: Yeah.  This is a temporary quick implementation while we only support one.  This method will need to be fleshed out to support multiple later, but at least it's contained in one spot now.

crates/weaver_resolver/src/loader.rs
jsuereth: Updated this to previous behavior and have TODO around https://github.com/open-telemetry/weaver/issues/1126.

PTAL at that issue and comment your thoughts.  I'll defer fixing it.

crates/weaver_resolver/src/attribute.rs
jsuereth: Good call! I missed this before merging, I'll remember it going forward.

crates/weaver_resolver/src/dependency.rs
lquerel: Hum. Maybe I misunderstood what you were trying to do here. It doesn't feel like this is about supporting multiple dependencies, but rather about the `imports` section in a semconv file, which allows defining a subset of metrics to import, the wildcard mechanism. Am I mistaken, or does using `find_map` prevent capturing all imports?

PR #1138, opened 2026-01-14
Implement `weaver registry infer` command

ArthurSens:
### TLDR
Implements weaver registry infer command that generates a semantic convention registry YAML file by inferring the schema from incoming OTLP telemetry data.

### Description
This PR adds a new weaver registry infer subcommand that starts a gRPC server to receive OTLP messages (traces, metrics, logs) and automatically infers a semantic convention schema from the observed telemetry. The command processes incoming data, deduplicates attributes across signals, and collects up to 5 unique example values per attribute to help document the inferred schema.
The inferred schema is written to a single registry.yaml file in the specified output directory (default: ./inferred-registry/). The output follows the standard semantic convention format with separate groups for resources, spans, metrics, and events. Resource attributes are currently accumulated into a single resource group; entity-based grouping (via OTLP EntityRef) is not yet supported but documented for future implementation.

### Testing
Tested by using weaver registry emit to send OTLP telemetry to the infer command's gRPC endpoint. The generated registry.yaml file was verified to contain the expected groups (resources, spans, metrics, events) with properly inferred attribute types and example values.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ All notable changes to this project will be documented in this file.
 - New feature ([#1100](https://github.com/open-telemetry/weaver/issues/1100)) - Set `--output=http` to have live-check send its report as the response to `/stop`. ([#1193](https://github.com/open-telemetry/weaver/pull/1193) by @jerbly)
 - Unified output handling across all registry subcommands. Builtin output formats (json, yaml, jsonl) are now available in `registry stats`, `registry diff`, and `registry resolve`. `registry stats` also supports `--templates` for custom text output templates. ([#1200](https://github.com/open-telemetry/weaver/pull/1200) by @jerbly)
 - New feature ([#1152](https://github.com/open-telemetry/weaver/issues/1152)) - Live-check with `--emit-otlp-logs` will now include the attributes from the resource in the emitted log record, this helps to identify the source of the finding in a multi-source environment. ([#1206](https://github.com/open-telemetry/weaver/pull/1206) by @jerbly)
-- New Experimental feature: `weaver registry infer` command that listens for OTLP telemetry and infers a semantic convention registry file from the received data. Supports spans, metrics, events, and resource attributes. Includes configurable gRPC address/port, admin server for graceful shutdown, and inactivity timeout.
+- New Experimental feature: `weaver registry infer` command that listens for OTLP telemetry and infers a semantic convention registry file from the received data. Supports spans, metrics, events, and resource attributes. Includes configurable gRPC address/port, admin server for graceful shutdown, and inactivity timeout. ([#1138](https://github.com/open-telemetry/weaver/pull/1138) by @ArthurSens)
 
 # [0.21.2] - 2026-02-03
 


src/registry/infer.rs
```diff
@@ -0,0 +1,723 @@
...
+            .entry(attr.name.clone())
+            .or_insert_with(|| AccumulatedAttribute::new(attr.name.clone(), attr.r#type.clone()));
+        entry.add_example(&attr.value);
+    }
+
+    fn add_span(&mut self, span: SampleSpan) {
+        let entry = self
+            .spans
+            .entry(span.name.clone())
+            .or_insert_with(|| AccumulatedSpan::new(span.name.clone(), span.kind.clone()));
```
ArthurSens, line 153: Some extra thoughts here: When the same span name is received multiple times with different `kind` values, only the first kind is preserved.

Not sure what to do to be honest, should I use more than just span/metric/event/resource name as the identifier? Maybe use all the fields that identify a particular telemetry type?

src/registry/infer.rs
```diff
@@ -0,0 +1,723 @@
...
+    }
+
+    fn add_sample(&mut self, sample: Sample) {
+        match sample {
+            Sample::Resource(resource) => self.add_resource(resource),
+            Sample::Span(span) => self.add_span(span),
+            Sample::Metric(metric) => self.add_metric(metric),
+            Sample::Log(log) => self.add_event(log.event_name, log.attributes),
+            Sample::Attribute(attr) => self.add_resource_attribute(attr),
+            _ => {} // Ignore other sample types for now
```
nicolastakashi: I think that would be useful some log entry, so if people push like a profile or a entity this won't work and will help understand why.

src/registry/infer.rs
```diff
@@ -0,0 +1,723 @@
...
+                    .or_insert_with(|| {
+                        AccumulatedAttribute::new(attr.name.clone(), attr.r#type.clone())
+                    });
+                attr_entry.add_example(&attr.value);
+            }
+        }
+    }
+
+    fn add_metric(&mut self, metric: SampleMetric) {
+        let instrument = match &metric.instrument {
```
nicolastakashi, line 174: What does it means, sorry maybe I'm not familiar with the concept of metric.instrument, is this the metric type?

jerbly: Kind of, it's `gauge`, `updowncounter`, `histogram`...

src/registry/infer.rs
jerbly: There currently is no identifier for a span so anything done here is guessing. Live-check cannot do span comparisons at the moment because of this too. It just looks at the attributes within.

src/registry/infer.rs
```diff
@@ -0,0 +1,723 @@
...
+                brief: String::new(),
+                span_kind: Some(span_kind_to_string(&span.kind)),
+                metric_name: None,
+                instrument: None,
+                unit: None,
+                name: None,
+                attributes,
+            });
+
+            // Span events as separate event groups
```
jerbly, line 283: Interesting... are span_events and log_events the same thing in semconv? In live-check I just check the attributes for span_events, whereas logs with event_name are checked against event definitions. @jsuereth / @lmolkova ?

jerbly:
> Opening as a draft first, manually tested and it seemed to work :)
> 
> Some question that I have:
> 
> 1. Should we build v2 schemas instead of v1?
> 2. I've created an object called `YamlGroup` to serialize the YAML file because I couldn't find another object that already does this. Don't we have something like that already? Could we re-use objects the unserialize YAML to also do the serialization somehow?
> 3. Is code organized correctly? I'm still struggling to understand when code should go to a separate crate and when it should be in the CLI module.
> 4. Do we want to implement entity inference already? I'm not sure how stable Entities are

1. IMO, we should make v2.
2. We should use the `weaver_semconv` crate to make the structure and then Serialize that to YAML.
3. As it stands it's OK. See below for further thoughts that might change this...
4. I'm not sure either, this has also not been done in live-check. @jsuereth to comment.

Overall I'm wondering what the intent of this command is. What you have made takes samples and aggregates them into entirely new definitions, I guess to use as a starting point model?

What I had in mind would have been more embedded in live-check, maybe a `--infer` option to live-check. You would then be comparing samples with an existing model, the otel semconv model by default. The inference would then be to create a new model that depends on and extends the model you're comparing with. This would make a definition with `imports`, `refs` and `extends`.

Also, live-check would be highlighting any items which would be troublesome to make an inference for: e.g. an attribute named `MyAttr` would fail policy checks around naming conventions (should be `some_namespace.my_attr` for example).

jerbly:
> So with your idea, if we add a `--infer` flag to live-check, instead of a json output we would get the YAML file as done in this PR so far?
> 
> I can work with that :)

No, I'm doing a bad job trying to explain this I think.

> * The output report could be interpreted as extentions in a child registry: We can infer this information if the OTLP message includes Samples that were not present before, is that correct?

I'm thinking the command could be: `weaver registry live-check -r https://github.com/open-telemetry/semantic-conventions/archive/refs/tags/v1.38.0.zip[model] --infer new` - this would collect samples and compare them with the otel registry. Let's say one of the samples is for `metric.X` with attributes: `server.address` and `server.port`. `metric.X` is not found in the otel registry but `server.address` and `server.port` are. The inferred output would be a new registry defining `metric.X` with **references** to `server.address` and `server.port`. Since we ran `--infer new`, weaver would also create a `registry_manifest.yaml` declaring the dependency on `https://github.com/open-telemetry/semantic-conventions/archive/refs/tags/v1.38.0.zip[model]`.

> * The output report could be interpreted as modifications to the registry: This is the part where I'm not understaning how we could tell. If our registry has a sample called metric.X, the OTLP message doesn't include this Sample but includes metric.Y... How do I know the difference between a Sample that was renamed or a Sample that was removed completely and a new unrelated one was added?

**This use case could be a later phase.**
In this case, the command could be: `weaver registry live-check -r my_model_dir --infer modify` - in this case, new registry files are created but suffixed with `_inferred`. Those registry files are a copy of the original with modifications made to them with any changes inferred from the live-check result. For example, let's say we used the registry generated in the example above. We receive a sample of `metric.X` with the server attributes but now also the attribute `error.type`. The registry is modified to add this attribute to the metric. This would retain any non-inferrable fields in the original registry e.g. `brief`, `note`, `annotations`.

I think we would need options to determine if weaver should add or overwrite when it finds differences. And, if you want weaver to remove definitions if they were not received in the samples.

`--infer modify` is quite a bit more complicated and I'm not sure it's worth it. But `--infer new`, where we're making a dependent child registry I think is important and inline with our multi-registry philosophy.

src/registry/infer.rs
ArthurSens: Just some comments from the SIG meeting today:

both span and log events can be mapped to events. Eventually, span events will be deprecated and we can remove this functionality from infer and add a warning in live check if they are present

src/registry/infer.rs
ArthurSens: Being a bit more specific for the discussion we had in our call @nicolastakashi, summaries are indeed deprecated in the OTLP proto, and Weaver is very explicit about not supporting it:

https://github.com/open-telemetry/weaver/blob/b55473ef359ef9f34c61f075055e6002d85ff6bf/src/registry/otlp/conversion.rs#L128

src/registry/infer.rs
```diff
@@ -0,0 +1,1209 @@
...
+            .or_insert_with(|| AccumulatedSpan::new(span.name.clone(), span.kind.clone()));
+
+        for attr in span.attributes {
+            let attr_entry = entry
+                .attributes
+                .entry(attr.name.clone())
+                .or_insert_with(|| {
+                    AccumulatedAttribute::new(attr.name.clone(), attr.r#type.clone())
+                });
+            attr_entry.add_example(&attr.value);
```
jerbly: Let's say I have messy real world telemetry. I may receive `attr=42` and then in another sample `attr="hello"`. This would make an `attr` of type `int` with examples `[42,"hello"]`.

How should mismatched types be handled?

src/registry/infer.rs
```diff
@@ -0,0 +1,1209 @@
...
+    /// Seconds of inactivity before auto-stop (0 = never).
+    #[arg(long, default_value = "60")]
+    inactivity_timeout: u64,
+}
+
+/// Accumulated attribute with examples
+#[derive(Debug, Clone)]
+struct AccumulatedAttribute {
+    name: String,
+    attr_type: Option<PrimitiveOrArrayTypeSpec>,
```
jerbly: Does this need to be optional? Attributes must have a type.

src/registry/infer.rs
```diff
@@ -0,0 +1,1209 @@
...
+            attributes: HashMap::new(),
+            events: HashMap::new(),
+        }
+    }
+}
+
+#[derive(Debug, Clone)]
+struct AccumulatedMetric {
+    name: String,
+    instrument: Option<InstrumentSpec>,
```
jerbly: Should this be optional?

src/registry/infer.rs
```diff
@@ -0,0 +1,1209 @@
...
+                    });
+                attr_entry.add_example(&attr.value);
+            }
+        }
+    }
+
+    fn add_metric(&mut self, metric: SampleMetric) {
+        let instrument = match &metric.instrument {
+            SampleInstrument::Supported(i) => Some(i.clone()),
+            SampleInstrument::Unsupported(_) => None,
```
jerbly: If we receive and unsupported instrument, we can't infer a semconv for it by definition. So we should reject the sample as not inferable.

src/registry/infer.rs
```diff
@@ -0,0 +1,1209 @@
...
+struct InferredRegistry {
+    groups: Vec<GroupSpec>,
+}
+
+/// Convert an accumulated attribute to a weaver_semconv AttributeSpec.
+fn accumulated_to_attribute_spec(attr: &AccumulatedAttribute) -> AttributeSpec {
+    let attr_type = attr
+        .attr_type
+        .clone()
+        .unwrap_or(PrimitiveOrArrayTypeSpec::String);
```
jerbly: As mentioned in an earlier comment. IMO we should not need to have an Optional type and therefore this goes away. I'm guessing this optionality comes from `Sample*` where we allow missing type or value. This makes sense for `live-check`, we can just compare an attribute name on its own for validity. It doesn't make sense for `infer`, this data is mandatory to make a semconv definition.

I would recommend removing the `Option`s where data is mandatory and rejecting samples with `None` types rather than carrying the Option all through the code.

src/registry/infer.rs
```diff
@@ -0,0 +1,1209 @@
...
+        deprecated: None,
+        annotations: None,
+        role: None,
+    }
+}
+
+/// Convert a vector of JSON values to the appropriate Examples type.
+///
+/// Uses serde to automatically match the JSON values to the correct Examples variant.
+fn json_values_to_examples(values: &[Value]) -> Option<Examples> {
```
jerbly: I mentioned this above too. All examples must be the same type and must match the type of the attribute. If you've determined `int` they must all be `int`. (Of course one strategy to handle mismatching types is to change the type to `Any` and then you can have `Any` examples).

You could build the `Examples` as they arrive in the samples rather than this two stage process. Perhaps a function `add_example(&value, &examples) -> Result<Examples, Error>` where it will make a new `Examples` given the current examples and value (perhaps changing a single example to an array).

src/registry/infer.rs
```diff
@@ -0,0 +1,1209 @@
...
+        serde_json::from_value::<Examples>(values[0].clone()).ok()
+    } else {
+        // For multiple values, create an array and let serde figure out the type
+        let arr = Value::Array(values.to_vec());
+        serde_json::from_value::<Examples>(arr).ok()
+    }
+}
+
+fn sanitize_id(name: &str) -> String {
+    name.replace(['/', ' ', '-', '.'], "_")
```
jerbly: Do we want to convert `.` to `_`? The namespace separator in OTel is `.`

src/registry/infer.rs
```diff
@@ -0,0 +1,1209 @@
...
+        .to_lowercase()
```
jerbly: Perhaps there's a way to use the `convert_case` crate to snake_case first and then deal with invalid chars. This would convert `HelloWorld` to `hello_world` too which is preferable to `helloworld`.

src/registry/infer.rs
```diff
@@ -0,0 +1,1209 @@
...
+    admin_port: u16,
+
+    /// Seconds of inactivity before auto-stop (0 = never).
+    #[arg(long, default_value = "60")]
+    inactivity_timeout: u64,
+}
+
+/// Accumulated attribute with examples
+#[derive(Debug, Clone)]
+struct AccumulatedAttribute {
```
jerbly: Could we accumulate right into the semconv structs and avoid these additional structs and the two stage process?

src/registry/infer.rs
ArthurSens: Ok gotcha, that makes sense. That allows removing the Option from instrument as well then

src/registry/infer.rs
ArthurSens: Good catch, I didn't think of this. I'm updating the PR to ignore attribute values that differ from the original value received.

This will probably create some race conditions, but not sure what else could be done here 🤔

src/registry/infer.rs
ArthurSens: whoops, force of habit 😅

ArthurSens:
> * optimizing with a single pass to accumulate and translate could be fun, not essential

I think I made it work for attributes at least, but I'm struggling a bit to make it work for metrics, spans and events. The hashmap is useful for quick lookups, and I'm not sure how to do the deduplication without the hashmaps 😬 
 
> I think, if we're not supporting v2 in this PR that's ok (it's marked experimental) but we should quickly move on to that in a follow-up. I'd also suggest, in the next PR, we move the main conversion code out to either one of the existing crates or a new one.

Happy to tackle both!

src/registry/infer.rs
```diff
@@ -0,0 +1,1269 @@
...
+                        attributes: Vec::new(),
+                        live_check_result: None,
+                    };
+                    for attribute in resource.attributes {
+                        sample_resource
+                            .attributes
+                            .push(sample_attribute_from_key_value(&attribute));
+                    }
+                    accumulator.add_sample(Sample::Resource(sample_resource));
+                }
```
jerbly: nit: this resource accumulation block is repeated for each signal. Maybe we can be more DRY here?

jerbly (approved):
Overall: Looks good for a first pass.

Perhaps when adding the v2 support there can be some refactoring to make this more idiomatic Rust. The conversion logic between `Sample*` types and `Accumulated*/AttributeSpec` types could use Rust's conversion traits:

- `From<&SampleAttribute> for AttributeSpec` - Replace `attribute_spec_from_sample()` with a `From` impl
- `From<&AccumulatedSpan> for GroupSpec` (and similar for Metric/Event) - Replace the inline conversion in `to_semconv_spec()`

Maybe add an `Accumulate` trait - Something like:
```rust
trait Accumulate {
    fn accumulate(&self, acc: &mut AccumulatedSamples);
}
```
Implement for `SampleResource`, `SampleSpan`, `SampleMetric`, etc. This would let `add_sample` become simply `sample.accumulate(self)`.

But, this is a great addition to weaver, let's get the first iteration in.

ArthurSens:
> Overall: Looks good for a first pass.
> 
> Perhaps when adding the v2 support there can be some refactoring to make this more idiomatic Rust. The conversion logic between `Sample*` types and `Accumulated*/AttributeSpec` types could use Rust's conversion traits:
> * `From<&SampleAttribute> for AttributeSpec` - Replace `attribute_spec_from_sample()` with a `From` impl
> * `From<&AccumulatedSpan> for GroupSpec` (and similar for Metric/Event) - Replace the inline conversion in `to_semconv_spec()`
> 
> Maybe add an `Accumulate` trait - Something like:
> 
> ```rust
> trait Accumulate {
>     fn accumulate(&self, acc: &mut AccumulatedSamples);
> }
> ```
> 
> Implement for `SampleResource`, `SampleSpan`, `SampleMetric`, etc. This would let `add_sample` become simply `sample.accumulate(self)`.
> 
> But, this is a great addition to weaver, let's get the first iteration in.

Aha, trait is the next chapter in the rust book 😛, but gotcha! I'll focus on that in the next PRs :)

jsuereth:
I'm fine merging, as this is a good first step in an evolution, but i"m deferring to @jerbly here.  He did the detailed review and has the vision for how it merges together with live-check over time.  I wasn't sure if any of his comments were blocking, but wanted to check.

jerbly:
> If I'm understanding #1206 correctly, including the parent resource in the inferred span doesn't affect the final result.
> 

You're correct. What you have now is fine.

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ All notable changes to this project will be documented in this file.
...
 - Unified output handling across all registry subcommands. Builtin output formats (json, yaml, jsonl) are now available in `registry stats`, `registry diff`, and `registry resolve`. `registry stats` also supports `--templates` for custom text output templates. ([#1200](https://github.com/open-telemetry/weaver/pull/1200) by @jerbly)
 - New feature ([#1152](https://github.com/open-telemetry/weaver/issues/1152)) - Live-check with `--emit-otlp-logs` will now include the attributes from the resource in the emitted log record, this helps to identify the source of the finding in a multi-source environment. ([#1206](https://github.com/open-telemetry/weaver/pull/1206) by @jerbly)
+- New Experimental feature: `weaver registry infer` command that listens for OTLP telemetry and infers a semantic convention registry file from the received data. Supports spans, metrics, events, and resource attributes. Includes configurable gRPC address/port, admin server for graceful shutdown, and inactivity timeout.
```
jerbly:
```suggestion
- New Experimental feature: `weaver registry infer` command that listens for OTLP telemetry and infers a semantic convention registry file from the received data. Supports spans, metrics, events, and resource attributes. Includes configurable gRPC address/port, admin server for graceful shutdown, and inactivity timeout. ([#1138](https://github.com/open-telemetry/weaver/pull/1138) by @ArthurSens)
```

Issue #1100, opened 2025-12-29
Live-check: return report in HTTP response when /stop is called


lmolkova:
Trying out weaver-live check to validate gen-ai instrumentations in python in https://github.com/open-telemetry/opentelemetry-python-contrib/compare/main...lmolkova:opentelemetry-python-contrib:add-weaver-live-check-for-gen-ai?expand=1 (mostly https://github.com/lmolkova/opentelemetry-python-contrib/blob/5a53eb3833a3560dab4c5b41ebe3d9e7f849e101/util/opentelemetry-util-genai/tests/weaver_container.py)

It would be much easier (no volume mapping, permissions, IO handling, reading logs) to return response or errors when /stop is invoked in HTTP response.

I.e. `/stop` would return 

- full json report (when validation succeed)
- list of errors (in json) if validation has failed (e.g. because of jq or rego errors) 

Also got this feedback at KubeCon from @pellared

PR #1145, opened 2026-01-20
[proposal] Expand out of the box defaults for weaver 

jsuereth:
A proposal for how to set up OpenTelemetry ecosystem to allow building out weaver defaults, and distribute them in the release of weaver.

### DIFF
--- a/docs/specs/default-templates/default_templates.md
+++ b/docs/specs/default-templates/default_templates.md
@@ -186,7 +186,7 @@ Loading dynamic configuration is a vector of attack, and concern for security. W
 | `diff_templates/ansi` | stays in weaver |
 | `diff_templates/ansi-stats` | stays in weaver |
 | `diff_templates/markdown` | removed* |
-| `diff_templates/yaml` | removed* |
+| `diff_templates/yaml` | stays in weaver* |
 | `jq/advice.jq` | stays in weaver |
 | `jq/semconv.jq` | stays in weaver |
 | `live_check_templates/ansi` | stays in weaver |


docs/specs/default-templates/default_templates.md
```diff
@@ -0,0 +1,133 @@
...
+Today, weaver provides a set of extension points via:
+
+- Policies: rego files
+- Templates: jinja and jq via `weaver.yaml`
+
+These are located in the `defaults` director and are loaded into weaver's binary via `include*!` macros within rust.
+
+While this gives us a baseline mechanism to offer a "batteries-loaded" weaver experience, it suffers from the following:
+
+* It bloats the binary size of weaver by including all templates in the binary.
```
lmolkova, line 20: orthogonal to the proposal, can we review existing defaults and potentially remove some of them (e.g. having 5 different diff templates in the artifact does not seem to be necessary)

docs/specs/default-templates/default_templates.md
```diff
@@ -0,0 +1,133 @@
...
+- ZIP - need the ability to verify signatures.
+- GIT - need the ability to reference specific commits.
+
+### Weaver Templates Repository
+
+- Creation of weaver templates repository
+- Each directory is a theme/name addressable distro 
+    - `codegen/java`, `codegen/go`, etc.
+    - `docs/markdown`, `docs/html`, etc.
+    - `checks/semconv`, `checks/backwards-compatibility`, etc.
```
lmolkova: Do you envision otel semconv reusing those? 

(similar question, would otel-go codegen leverage common go codegen)?

The risk is that otel semconv policies/templates would be perpetually out of sync with `opentelemetry-weaver-templates` and there will be a lot of copy-paste. 

I'm fine with this risk, and I'd be against *requiring* semconv / language SIGs to reuse common templates.

docs/specs/default-templates/default_templates.md
jsuereth: absolutely!!!

docs/specs/default-templates/default_templates.md
jsuereth: I don't think we require *generation templates* here.

I do think we could pull out a set of "backward compatibility" and "stability" requirements we enforce generically and require these as part of the "federated semconv" work in the stability OTEP.

But, e.g. naming convention rules, the specific Markdown snippets we create, no - I do not think it makes sense to *force* those to be shared.  Overtime they may be, because maintaining two things will become silly once the "new" codegen is up to speed, robust and generally used everywhere

docs/specs/default-templates/default_templates.md
```diff
@@ -0,0 +1,133 @@
...
+templates:
+  codegen:
+    java: https://github.com/open-telemetry/opentelemetry-weaver-templates.git:<sha>\codegen/java
+  docs:
+    markdown: https://github.com/open-telemetry/opentelemetry-weaver-templates.git:<sha>\docs/markdown
+policies:
+    checks:
+      semconv: https://github.com/open-telemetry/opentelemetry-weaver-templates.git:<sha>\checks/semconv
+    advice:
+      semconv: https://github.com/open-telemetry/opentelemetry-weaver-templates.git:<sha>\advice/semconv
```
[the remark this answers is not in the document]
jsuereth: Note for readers -  If you read the implementation of `parse` for `VirtualDirectoryRef`, you'll see `\` was a choice made for disambiguating between a URL and a path inside the directory that URL represents (e.g. a git repository or zip archive).  Do we like the use of `\`?

docs/specs/default-templates/default_templates.md
```diff
@@ -0,0 +1,133 @@
...
+This document outlines how Weaver can enable federated development and inclusion of out of the box templates and policies for code generation, documentation, error reporting and stability enforcement.
+
+## Background
+
+Today, weaver provides a set of extension points via:
+
+- Policies: rego files
+- Templates: jinja and jq via `weaver.yaml`
+
+These are located in the `defaults` director and are loaded into weaver's binary via `include*!` macros within rust.
```
lquerel:
```suggestion
These are located in the `defaults` directory and are loaded into weaver's binary via `include*!` macros within rust.
```

docs/specs/default-templates/default_templates.md
```diff
@@ -0,0 +1,133 @@
...
+
+- Language, Policy and Documentation experts are able to contribute templates and policies independently of the main contribution of weaver.
+    - They can depend on a *stable* version of weaver for development
+    - They can independently verify/test their templates.
+    - They can independently release their templates.
+- Weaver can decide which versions of templates and policies to
+  include by default in any release.
+    - Weaver can depend on *stable* versions of templates & policies
+    - Weaver can `smoke test` that its inclusion is working.
+    - Weaver can (optionally) bundle templates with its executable or docker container to avoid the need for network connection for out of the box behavior.
```
lquerel, line 41: I checked, and today our **default artifacts** (templates, policies, and jq expr) are about 115 KiB uncompressed, so embedding them systematically in the executable is not a big deal in my opinion. It has the benefit of being self-contained, always accessible, and harder for an external actor to modify.

docs/specs/default-templates/default_templates.md
```diff
@@ -0,0 +1,133 @@
...
+- Out of the box templates and policies can be referred to via "simple" names, e.g. how `ansi` works for diagnostic output today.
+- Weaver users are able to override out of the box defaults.
+
+## Design Proposal
+
+The design relies on three key capabilities:
+
+- Expanding weaver's `VirtualDirectory` capabilities to support additional verification features on remote directories (archive downloads and git repositories).
+- Creation of a `opentelemetry-weaver-templates` repository for contributors to out of the box templates for weaver.
```
lquerel: It's not the most important thing, but having a slightly less specific name would be good to signal that people can contribute not only templates, but also policies for Weaver check or live-check. During the meeting, the following repository names were proposed:

* opentelemetry-weaver-packages
* opentelemetry-weaver-artifacts
* opentelemetry-weaver-contrib

lmolkova: from the SIG call:

minor: `templates` does not cover policies, maybe `extensions`?

from me (also minor):
can we remove opentelemetry prefix ? We'd have https://github.com/open-telemetry/weaver and https://github.com/open-telemetry/weaver-extensions|templates - and both can be used outside of OTel context (Prometheus).  It'd still be in otel org, but otel is not in the repo name.

lmolkova: the important question: would we host default codegen templates for Prometheus in the repo? I would be happy to as long as Prometheus community is interested and does not provide alternative /cc @ArthurSens @vesari

docs/specs/default-templates/default_templates.md
```diff
@@ -0,0 +1,133 @@
...
+
+First, we create a configuration file for weaver for what to include in each release:
+
+```yaml
+templates:
+  codegen:
+    java: https://github.com/open-telemetry/opentelemetry-weaver-templates.git:<sha>\codegen/java
+  docs:
+    markdown: https://github.com/open-telemetry/opentelemetry-weaver-templates.git:<sha>\docs/markdown
+policies:
```
lquerel, line 127: Two comments:
- We should support multiple policy packages that can all be applied together, for example OTel-level policies, enterprise-level policies, and team-level policies.
- We have two types of policies, those for `weaver check` and those for `weaver live-check`. I think it would be more consistent to reuse these names to label the policy sections, so instead of `advice` we would have `live-check`.

docs/specs/default-templates/default_templates.md
```diff
@@ -0,0 +1,133 @@
...
+- ZIP - need the ability to verify signatures.
+- GIT - need the ability to reference specific commits.
+
+### Weaver Templates Repository
+
+- Creation of weaver templates repository
+- Each directory is a theme/name addressable distro 
+    - `codegen/java`, `codegen/go`, etc.
+    - `docs/markdown`, `docs/html`, etc.
+    - `checks/semconv`, `checks/backwards-compatibility`, etc.
```
lmolkova: another nit from the call: we might want to separate schema checks from telemetry check and then call them accordingly, e.g.  `schema-check` and `telemetry|live-check`.

From my side (and after discussions with @xrmx), providing JSON in response to live-check `/stop` API (https://github.com/open-telemetry/weaver/issues/1100) would be the preferred way for instrumentation tests - having to deal with report files is complicated.  Writing python/any code is preferrable over JQ filters, so I, at the moment, don't see a lot of future for the live-check templates beyond `ansi`

docs/specs/default-templates/default_templates.md
ArthurSens: I didn't propose that because I felt like I would be hijacking an idea even before it came into reality, but absolutely 😀 

In https://github.com/prometheus/prometheus/pull/17868, I was adding rego policies and templates to generate Prometheus Go SDK code to the Prometheus repository, but discoverability of those files would be terrible if they were hosted there.

martinjt:
Something I think I might have missed here, are you proposing multiple templates here? This is why I mentioned not bundling.

I'm thinking more about markdown than codegen, but it is probably applicable to codegen in some ways.

So I could have markdown for Hugo, something specific to confluence, or any number of other things. I could even have a html one with different styles too. I thought this was about providing the ability for people to provide a location to collaborate on different ways to represent the information in their schema, as opposed to just one of each.

I'm thinking now, that your intent isn't to have that though, and its more about keeping this as a single implementation of each? Perhaps I skimmed the proposal too quickly the first time.

docs/specs/default-templates/default_templates.md
jsuereth: I like `opentelemetry-weaver-packages` best of this, WDYT?

jsuereth: Yes - My thought was we'd happily host these if they have owners.

docs/specs/default-templates/default_templates.md
jsuereth: That's true today - If we start taking in, e.g. out of the box markdown generation or codegen, that's where the problem lies.

This goes into the decision of "what should be baked in" and "what should require the internet"

ArthurSens (approved):
:shipit:

docs/specs/default-templates/default_templates.md
jsuereth: Fair - I've added a table to denote which existing defaults would remain in weaver and a suggested directory structure for packages.

docs/specs/default-templates/default_templates.md
jsuereth: I think I agree.  For now I've used `telmetry` and `live-check` but I can switch to `check` and `live-check`.

docs/specs/default-templates/default_templates.md
lmolkova: wdyt about calling it `weaver-packages` then? removing otel from the repo name

docs/specs/default-templates/default_templates.md
```diff
@@ -0,0 +1,197 @@
...
+- Weaver should use git HASH when resolving from git repositories.
+
+## Survey of Weaver's current built-ins
+
+| Package | Action |
+|--|--|
+| `diagnostic_templates/ansi` | stays in weaver |
+| `diagnostic_templates/gh_workflow_command` | stays in weaver |
+| `diagnostic_templates/json` | stays in weaver* |
+| `diff_templates/json` | stays in weaver* |
```
lmolkova, line 185: not blocking, we can change it later - why this one stays and yaml is removed? 

I probably see the principle behind templates list: one ansi, one structured/formal template. If it's the case, then yaml is the first class citizen for diff

docs/specs/default-templates/default_templates.md
```diff
@@ -0,0 +1,197 @@
...
+- Weaver should allow signature verification of packages it downloads as ZIP.
+- Weaver should use git HASH when resolving from git repositories.
+
+## Survey of Weaver's current built-ins
+
+| Package | Action |
+|--|--|
+| `diagnostic_templates/ansi` | stays in weaver |
+| `diagnostic_templates/gh_workflow_command` | stays in weaver |
+| `diagnostic_templates/json` | stays in weaver* |
```
lmolkova, line 184: can we remove it? are these signs that CI jobs that run weaver would commonly need json for diagnostics?

docs/specs/default-templates/default_templates.md
```diff
@@ -0,0 +1,197 @@
...
+| `diff_templates/json` | stays in weaver* |
+| `diff_templates/ansi` | stays in weaver |
+| `diff_templates/ansi-stats` | stays in weaver |
```
lmolkova, line 187: do we need it? diff usage is super-limited today and ansi-stats can be built as an opinionated cusomization

docs/specs/default-templates/default_templates.md
jsuereth: Plan is to remove the JINJA part of this, and generically support JSON/YAML/etc. output via serde.  Personally, I don't really care too much about JSON, but we do use it for our own testing purposes, so it'd be hard to just kill.

I think ANSI / Github-Workflow defaults should remain baked in.

docs/specs/default-templates/default_templates.md
jsuereth: I'd be ok removing, personally.  I'm not aware of significant usage and agree this could be provided somewhere else.

docs/specs/default-templates/default_templates.md
jsuereth: That may just be a mistake on my part.  Both should have the `*` denoting they are NO longer templates, but first class SERDE supported export to avoid the wonkyness.  I'll update this before  merging.

docs/specs/default-templates/default_templates.md
jerbly: See #1157 for initial support for live-check with builtin serde output in json, jsonl and yaml.

docs/specs/default-templates/default_templates.md
jerbly: See #1157

PR #1106, opened 2026-01-02
Add doc summarizing info on v2 schemas

lmolkova:
Cleanup and document v2 schemas:

- add `head_schema_url` and `baseling_schema_url` to diff
- add missing JSON schemas
- rename forge to materialized in json-schema command
- add md overview

PR #1154, opened 2026-01-23
Replace version in definition schema with file_format

lmolkova:
Align definition schema marker with resolved schema introduced in https://github.com/open-telemetry/weaver/pull/1136

Extracted from https://github.com/open-telemetry/weaver/pull/1106

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ All notable changes to this project will be documented in this file.
 - Use `schema_url` in registry manifest, dependencies, and resolved schema instead of `registry_url`. Parse registry name and version
   from it. ([#1202](https://github.com/open-telemetry/weaver/pull/1202) by @lmolkova)
 - Default to `manifest.yaml` for registry manifest file, deprecate `registry_manifest.yaml` and add warning when it's used. ([#1202](https://github.com/open-telemetry/weaver/pull/1202) by @lmolkova)
+- 💥 BREAKING CHANGE 💥 (Fixes [#760](https://github.com/open-telemetry/weaver/issues/760)) - Auto-escaping is now off by default (`none`) for all templates, regardless of file extension. To opt in, set `auto_escape: html` or `auto_escape: json` per template in `weaver.yaml`. Within a template, `{% autoescape false %}` blocks can selectively disable escaping for sections. Use `|tojson` for explicit JSON/YAML value escaping where needed. ([#1239](https://github.com/open-telemetry/weaver/pull/1239) by @jerbly)
 - 💥 BREAKING CHANGE 💥 Replace `version: "2"` with `file_format: definition/2` for v2 definition schema ([#1154](https://github.com/open-telemetry/weaver/pull/1154) by @lmolkova)
 
 # [0.21.2] - 2026-02-03
--- a/crates/weaver_forge/src/config.rs
+++ b/crates/weaver_forge/src/config.rs
@@ -143,6 +143,32 @@ pub enum ApplicationMode {
     Each,
 }
 
+/// Auto-escape mode for a template's output.
+/// `none` (default): no escaping.
+/// `html`: escape `<`, `>`, `&`, `"`, `'`, `/` for HTML/XML.
+/// `json`: serialize values as JSON (suitable for JSON/YAML embedding).
+///
+/// This mirrors `minijinja::AutoEscape` but adds serde support for
+/// deserialization from `weaver.yaml`.
+#[derive(Serialize, Deserialize, Debug, Clone, Default, PartialEq)]
+#[serde(rename_all = "snake_case")]
+pub(crate) enum AutoEscapeMode {
+    #[default]
+    None,
+    Html,
+    Json,
+}
+
+impl From<&AutoEscapeMode> for minijinja::AutoEscape {
+    fn from(mode: &AutoEscapeMode) -> Self {
+        match mode {
+            AutoEscapeMode::None => minijinja::AutoEscape::None,
+            AutoEscapeMode::Html => minijinja::AutoEscape::Html,
+            AutoEscapeMode::Json => minijinja::AutoEscape::Json,
+        }
+    }
+}
+
 /// A template configuration.
 #[derive(Serialize, Deserialize, Debug, Clone)]
 #[serde(rename_all = "snake_case")]
@@ -173,6 +199,14 @@ pub(crate) struct TemplateConfig {
     /// The default value of this path is the same as the input file path.
     /// This file path can be a Jinja expression referencing the parameters.
     pub(crate) file_name: Option<String>,
+    /// Auto-escape mode for this template's output.
+    /// `none` (default): no escaping.
+    /// `html`: escape `<`, `>`, `&`, `"`, `'`, `/` for HTML/XML.
+    /// `json`: serialize values as JSON (suitable for JSON/YAML embedding).
+    /// Within a template, `{% autoescape false %}` blocks can selectively
+    /// disable escaping for sections.
+    #[serde(default)]
+    pub(crate) auto_escape: AutoEscapeMode,
 }
 
 fn default_filter() -> String {
--- a/crates/weaver_forge/src/file_loader.rs
+++ b/crates/weaver_forge/src/file_loader.rs
@@ -66,6 +66,20 @@ impl EmbeddedFileLoader {
         local_dir: PathBuf,
         target: &str,
     ) -> Result<Self, Error> {
+        let target_local_dir = local_dir.join(target);
+        if target_local_dir.exists() {
+            let fs_loader = FileSystemFileLoader::try_new(local_dir, target)?;
+            log::debug!(
+                "Using local templates from `{}`",
+                target_local_dir.display()
+            );
+            return Ok(Self {
+                target: target.to_owned(),
+                embedded_dir,
+                fs_loader: Some(fs_loader),
+            });
+        }
+
         let target_embedded_dir =
             embedded_dir
                 .get_dir(target)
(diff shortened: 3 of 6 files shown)


crates/weaver_forge/data/registry/http.yaml
```diff
@@ -1,4 +1,4 @@
-version: "2"
+file_format: definition/2.0.0
```
jsuereth: Nit: I'm not sure we want a three-version rule on language syntax.  Or at least we should answer some questions:

1. When do we *fail to load* a file because of major/minor versions?  E.g. do we need something in place where weaver will *fail* to resolve `definition/2.1.0` right now because we haven't defined the syntax there?
2. When will we bump this version?  Should we do so on any major feature addition to the syntax? (probably). We should open a ticket to have some kind of automated tracker to let us know to do this so it doesn't get lost.
3. Do we need a tool to migrate from version 2.x -> 2.y?  I had started working on such a thing for 1->2.

I'm on board moving this direction of using `file_format` everywhere, and we can make these decisions over time, but I'd like to write down our versioning/loading policy a bit more (it can be in follow on bugs) and make sure our release that uses this syntax can "scale" to future versions/usage.

crates/weaver_semconv/src/semconv.rs
```diff
@@ -14,25 +14,25 @@ use std::path::Path;
...
-#[serde(tag = "version")]
+#[serde(tag = "file_format")]
 pub enum Versioned {
     /// Version 1 of the semantic convention schema.
-    #[serde(rename = "1")]
+    #[serde(rename = "definition/1.0.0")]
     V1(SemConvSpecV1),
     /// Version 2 of the semantic convention schema.
-    #[serde(rename = "2")]
+    #[serde(rename = "definition/2.0.0")]
```
jsuereth: This will ONLY handle the literal string `definition/2.0.0`.  If we were to try a version bump, like `definition/2.1.0` then this would fail to resolve previous versions.

I think there's some HARD implications here for us to sort through.  E.g. I'd *prefer* If we had a custom deserialization path for SERDE that would resolve version string and then delegate to a deserializer of choice.

That custom path would be responsible for 

- knowing if the version that came is compatible with versions we can represent in the current weaver.
- Choosing the right Rust structure to deserialize into.

crates/weaver_forge/data/registry/http.yaml
lmolkova: That's a great point. I don't really see a point in following semver for the definition. I hope we'll deprecate v1 and remove support for it at some point, then this won't be necessary at all.

But I still want a bit of consistency with `file_format` (primarily i think `version` is ambiguous) , so I updated the PR to do`file_format: definition/2`.

crates/weaver_semconv/src/semconv.rs
```diff
@@ -8,31 +8,81 @@ use crate::provenance::Provenance;
...
+
+        let is_v2 = matches!(file_format, Some("definition/2")) || matches!(version, Some("2"));
+        let is_v1 = matches!(file_format, Some("definition/1")) || matches!(version, Some("1"));
+
+        _ = mapping.remove(&file_format_key);
+        _ = mapping.remove(&version_key);
+
+        let cleaned = serde_yaml::Value::Mapping(mapping);
+        if is_v2 {
+            let v2 = serde_yaml::from_value::<SemConvSpecV2>(cleaned).map_err(de::Error::custom)?;
```
jsuereth: Can we make sure the errors are still meaningful here?

crates/weaver_semconv/src/lib.rs
```diff
@@ -296,12 +296,12 @@ pub enum Error {
         error: String,
     },
 
-    /// This indicates the file version used is not yet stable.
-    #[error("Version `{version}` schema file format is not yet stable: {provenance}")]
+    /// This indicates the file format (version) used is not yet stable.
+    #[error("Version `{file_format}` schema file format is not yet stable: {provenance}")]
```
jsuereth: Nit: Should this now read "File Format"?

jsuereth (commented):
Overall looks good, two major questions:

- One on errors and what they look like wrapped now
- Another on whether we should update the file format error message

crates/weaver_semconv/src/semconv.rs
lmolkova: I rewrote this piece - custom deserialization here is now only used with in-memory conventions (for tests). Actual logic happens in `from_file*` methods which deserialize v1 or v2, but not the variant. 

Messages are the same, but I believe we should no longer see errors that show both variants - they can no longer happen.

I can spend more time cleaning up, but I'd rather burn it alltogether once we move to v2

crates/weaver_semconv/src/semconv.rs
```diff
@@ -8,31 +8,46 @@ use crate::provenance::Provenance;
...
+    /// Semantic convention specification that does NOT include a file_format tag.
     NoVersion(SemConvSpecV1),
 }
 
+impl<'de> Deserialize<'de> for SemConvSpec {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        // this custom deserialization is used in tests, in real
```
jsuereth: If this is true, let's add #[cfg(test)] to prevent this from being public or document this with `///` on the impl itself.

crates/weaver_semconv/src/semconv.rs
jsuereth: I understand the frustration, but I also expect we're not quite done with this section and we'll need to maintain this code / make improvements during the v2 rollout.

I'm ok with some shortcuts, but I was struggling to follow some of this, so I'd like the minimum amount of cleanup.

crates/weaver_semconv/src/semconv.rs
```diff
@@ -167,18 +182,162 @@ fn provenance_path_to_name(path: &str) -> String {
...
+
+fn from_yaml_value(
+    yaml_value: serde_yaml::Value,
+    provenance: &str,
+    unversioned_validator: Option<&JsonSchemaValidator>,
+    versioned_validator_v1: Option<&JsonSchemaValidator>,
+    versioned_validator_v2: Option<&JsonSchemaValidator>,
+    warnings: &mut Vec<Error>,
+) -> Result<SemConvSpec, Error> {
+    let (file_format, version, is_v1, is_v2) =
```
jsuereth: I'm kinda struggling to go through this code and understand it architecturally. Additionally, the Json Schema validator is becoming a mess here, which is just compounding the minor cleanup and then giant mess *i* made when adding v2, assuming we'd clean this up later.

A few suggestions:

- Should this `check_version_and_determing_format` return an `enum` instead of a bunch of loose booleans? That's a more "rust" way to make sure you're covering all cases.
- Should we move the handling of different version validation into `JsonSchemaValidator`, or create a *new* thing which uses it called `MultiVersionJsonSchemaValidator` you can use instead of passing around a bunch of optional values?  Another option is the `check_version_and_determine_format` call can return the appropriate validator and error if `version` isn't defined.

jsuereth:
Overall, I'm not happy with the error messages here, example:

```
ℹ No registry manifest found: dummy-repo/registry_manifest.yaml

Diagnostic report:

  × "Missing required property: \"type\".\n\nMissing required property: \"brief\".\n\nMissing required property: \"stability\".\n\nMissing required property: \"instrument\".
  │ \n\nMissing required property: \"unit\".\n\nMissing required property: \"brief\".\n\nMissing required property: \"stability\"."
  ```

File: crates/weaver_semconv/src/semconv.rs (hunk 6 above)
Comment (lmolkova) [reply crates/weaver_semconv/src/semconv.rs]: check the new version out, it's much shorter thanks to your suggestions

File: crates/weaver_semconv/src/semconv.rs (hunk 5 above)
Comment (lmolkova) [reply crates/weaver_semconv/src/semconv.rs]: removed the whole struct - not needed

**Comment by jsuereth** [comment]
@lmolkova I'll take a look, and the formatting I was able to confirm head vs. this PR that this PR *was* the cause as of yesterday.  I'll re-evaluate the updates, thanks for taking a look!

File: crates/weaver_semconv/src/semconv.rs [hunk 7]
```diff
@@ -167,18 +155,147 @@ fn provenance_path_to_name(path: &str) -> String {
...
+    /// Explicit `file_format: definition/1` (or legacy `version: 1`).
+    V1,
+    /// Explicit `file_format: definition/2` (or legacy `version: 2`).
+    V2,
+    /// No `file_format` or `version` field — treated as V1.
+    Unversioned,
+}
+
+impl FileFormat {
+    fn detect(
```
jsuereth, line 173: Nit: can you make sure the goal of these methods is documented?

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ All notable changes to this project will be documented in this file.
 - Use `schema_url` in registry manifest, dependencies, and resolved schema instead of `registry_url`. Parse registry name and version
   from it. ([#1202](https://github.com/open-telemetry/weaver/pull/1202) by @lmolkova)
 - Default to `manifest.yaml` for registry manifest file, deprecate `registry_manifest.yaml` and add warning when it's used. ([#1202](https://github.com/open-telemetry/weaver/pull/1202) by @lmolkova)
+- 💥 BREAKING CHANGE 💥 Replace `version: "2"` with `file_format: definition/2' for v2 definition schema ([#1154](https://github.com/open-telemetry/weaver/pull/1154) by @lmolkova)
```
jsuereth: Look like a mismatched backtick here for `file_format: definition/2`

jsuereth (approved):
I'm good with this as an interim state we'll clean up.

- I'm not super happy with creating temp files, but it's just for tests.  That could be cleaned up, but we can do so later.
- The code is a LOT easier to read now, thank you.  Think we should be able to maintain this for the time it takes to migrate to v2.

Issue #1132, opened 2026-01-12
jsonl exporter for registry live-check


xrmx:
**Is your feature request related to a problem? Please describe.**

I'm running weaver in CI with popen and it would be great to communicate with it using stdout / stderr without touching the filesystem. The current --format=json option outputs a stream of json object, e.g.:

```
{
  "...": ...
}
{
  "..": ...
}
````

Which are harder to parse since the whole output is not a valid json document.

**Describe the solution you'd like**

Providing a json lines (https://jsonlines.org)  formatter that is using newline as document separator would make it easier to consume it, e.g.:

```
{"...": ...}
{"...": ...}
```

**Describe alternatives you've considered**

I've tried --output but looks like there's a race condition there between the different signals, e.g. metrics override traces.

**Additional context**

<sub>**Tip**: [React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

PR #1157, opened 2026-01-26
Live check output formats

jerbly:
Feature ([#1132](https://github.com/open-telemetry/weaver/issues/1132)) - Live-check: builtin output rendering for json, jsonl and yaml.

Added an `OutputProcessor` to weaver_forge which handles json, jsonl, yaml for builtin formats, templates or mute (no output).

The end goal is to unify all weaver output in a standard way. **This first PR tackles this only for Live-check only.**

A follow-up PR will fit this for other commands: `diff` and `resolve`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -623,9 +623,9 @@ checksum = "3d52eff69cd5e647efe296129160853a42795992097e8af39800e1060caeea9b"
 
 [[package]]
 name = "convert_case"
-version = "0.8.0"
+version = "0.11.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "baaaa0ecca5b51987b9423ccdc971514dd8b0bb7b4060b983d3664dad3f1f89f"
+checksum = "affbf0190ed2caf063e3def54ff444b449371d55c58e513a95ab98eca50adb49"
 dependencies = [
  "unicode-segmentation",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -64,7 +64,7 @@ opentelemetry-stdout = { version = "0.31.0", features = [
 ] }
 tokio = { version = "1.47.1", features = ["full"] }
 rouille = "3.6.2"
-convert_case = "0.7.1"
+convert_case = "0.11"
 log = { version = "0.4.28", features = ["std"] }
 utoipa = "5.3"
 rand = "0.9.2"
--- a/crates/weaver_forge/Cargo.toml
+++ b/crates/weaver_forge/Cargo.toml
@@ -27,7 +27,6 @@ minijinja = { version = "2.12.0", features = [
     "macros",
 ] }
 minijinja-contrib = { version = "2.12.0", features = ["pycompat"] }
-convert_case = "0.8.0"
 jaq-core = "2.2.1"
 jaq-std = "2.1.2"
 jaq-json = { version = "1.1.3", features = ["serde_json"] }
@@ -36,6 +35,7 @@ regex = "1.11.3"
 markdown = "=1.0.0"
 textwrap = "0.16.2"
 
+convert_case.workspace = true
 itertools.workspace = true
 thiserror.workspace = true
 serde.workspace = true
--- a/crates/weaver_forge/src/config.rs
+++ b/crates/weaver_forge/src/config.rs
@@ -26,9 +26,7 @@ use std::fmt::{Display, Formatter};
 use std::path::Path;
 use std::sync::OnceLock;
 
-use convert_case::pattern::Pattern;
-use convert_case::Converter;
-use convert_case::{pattern, Boundary};
+use convert_case::{Boundary, Converter, Pattern};
 use dirs::home_dir;
 use globset::{Glob, GlobSet, GlobSetBuilder};
 use serde::{Deserialize, Serialize};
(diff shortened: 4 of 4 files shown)


crates/weaver_forge/src/output_processor.rs
```diff
@@ -0,0 +1,484 @@
...
+            BuiltinFormat::Yaml => "yaml",
+            BuiltinFormat::Jsonl => "jsonl",
+        }
+    }
+
+    /// Serialize data to string
+    fn serialize<T: Serialize>(&self, data: &T) -> Result<String, Error> {
+        match self {
+            BuiltinFormat::Json => {
+                serde_json::to_string_pretty(data).map_err(|e| Error::SerializationError {
```
jsuereth, line 42: Nit: Should we have a `json_pretty` and `json` as separate options?

crates/weaver_forge/src/output_processor.rs
```diff
@@ -0,0 +1,484 @@
...
+
+    /// Whether this format appends a newline after each generate call
+    fn is_line_oriented(&self) -> bool {
+        matches!(self, BuiltinFormat::Jsonl)
+    }
+}
+
+/// Output processor - handles output generation with builtin formats or templates
+#[allow(clippy::large_enum_variant)]
+pub enum OutputProcessor {
```
jsuereth: Nit:

This may be better as a `trait`.  If we need a container to wrap it, we could do something like:

```
struct OutputProcessorOpt(dyn OutputProcessor);

impl <T: OutputProcessor> From<T> : OutputProcessorImpl {
   fn from(p: T) -> OutputProcessorImpl { OutputProcessorImpl(p) }
}
```

crates/weaver_forge/src/output_processor.rs
```diff
@@ -0,0 +1,484 @@
...
+                    engine,
+                    path,
+                    directive,
+                })
+            }
+        }
+    }
+
+    /// Open/create the output file if not already open.
+    fn open(&mut self) -> Result<(), Error> {
```
jsuereth: These kinds of "implementation deferred" functions are why I think it may be easier to use a `trait` instead of enum here.

It's a bit hard to follow the stateful behavior across the enum when all the methods are in the same `impl`.

crates/weaver_forge/src/output_processor.rs
```diff
@@ -0,0 +1,484 @@
...
+            if *directive == OutputDirective::File && file.is_none() {
+                let filename = format!("{}.{}", prefix, format.extension());
+                *file = Some(Self::create_file(path, &filename)?);
+            }
+        }
+        Ok(())
+    }
+
+    /// Create/truncate a file and return the handle
+    fn create_file(path: &std::path::Path, filename: &str) -> Result<File, Error> {
```
jsuereth: Nit - this could be a free-standing helper function.

crates/weaver_forge/src/output_processor.rs
```diff
@@ -0,0 +1,484 @@
...
+                    writeln!(f, "{content}")
+                } else {
+                    write!(f, "{content}")
+                }
+                .map_err(|e| Error::SerializationError {
+                    error: e.to_string(),
+                })
+            }
+            OutputDirective::Stderr => {
+                unreachable!("OutputProcessor does not support Stderr directive")
```
jsuereth: Why's that?  Should we support stderr?

crates/weaver_forge/src/output_processor.rs
```diff
@@ -0,0 +1,484 @@
...
+        match self {
+            OutputProcessor::Builtin { directive, .. }
+            | OutputProcessor::Template { directive, .. } => *directive == OutputDirective::File,
+            OutputProcessor::Mute => false,
+        }
+    }
+
+    /// Returns the builtin format if this is a builtin processor
+    #[must_use]
+    pub fn builtin_format(&self) -> Option<BuiltinFormat> {
```
jsuereth: NIt: Is this needed publicly? Anyway we could hide this?

src/registry/live_check.rs
```diff
@@ -172,28 +172,43 @@ fn default_advisors() -> Vec<Box<dyn Advisor>> {
     ]
 }
 
+/// Generate output for a complete report - handles JSONL special case
+fn generate_report(
+    output: &mut OutputProcessor,
+    samples: Vec<Sample>,
+    stats: LiveCheckStatistics,
+) -> Result<(), weaver_forge::error::Error> {
+    // Special JSONL handling: one line per sample, stats at end
```
jsuereth: Could we create something like - "needs_multiple_events()" or "is_streaming()" on the OutputProcessor instead?

If we were to use `jsonl` other places *or* if we add other new built-in formats, It'd be nice if we didn't need to alter this logic.

src/registry/live_check.rs
```diff
@@ -172,28 +172,43 @@ fn default_advisors() -> Vec<Box<dyn Advisor>> {
...
+    if output.builtin_format() == Some(weaver_forge::output_processor::BuiltinFormat::Jsonl) {
+        for sample in &samples {
+            output.generate(sample)?;
+        }
+        match stats {
+            LiveCheckStatistics::Cumulative(_) => output.generate(&stats),
+            LiveCheckStatistics::Disabled(_) => Ok(()),
+        }
+    } else if matches!(output, OutputProcessor::Mute) {
```
jsuereth: Nit: If Mute processor does nothing, why not just call it?

src/registry/live_check.rs
```diff
@@ -172,28 +172,43 @@ fn default_advisors() -> Vec<Box<dyn Advisor>> {
...
-            OutputDirective::Stdout // Will be ignored when no_output is true
-        } else {
-            output = path_buf.clone();
-            OutputDirective::File
-        }
-    } else {
-        OutputDirective::Stdout
-    };
+
+    let mut output = OutputProcessor::new(
```
jsuereth, line 203: I like this cleanup!

ui/package-lock.json
jsuereth, line 1: We need to fix this somehow.

crates/weaver_forge/src/output_processor.rs
jerbly: I think `jsonl` is the json_ugly. The way I've written `jsonl` handling leaves it to the caller to determine where they want the newlines in the output. You could just have one for the entire output.

crates/weaver_forge/src/output_processor.rs
jerbly: I'll look at a refactoring... the clippy thing is not a big deal IMO since you make one of these not thousands.

crates/weaver_forge/src/output_processor.rs
jerbly: As above... I'll look at that refactoring.

crates/weaver_forge/src/output_processor.rs
jerbly: This and `open` are supposed to be the hidden internals. I guess this might be useful elsewhere??

crates/weaver_forge/src/output_processor.rs
jerbly: I will revisit this when I integrate OutputProcessor with the other weaver commands. StdErr should be reserved for the logger output so we don't mix artifacts with runtime information. I think this may change for diagnostic output, but I need to get to that.

src/registry/live_check.rs
jerbly: I went back and forth on this a few times and landed on this API. This handling is specifically for jsonl in live-check's report mode. Most things, I think, will have a single object (albeit large) output. The API here allows you to call `generate` multiple times and it will append to the file. The first call to `generate` opens the file.

Again, if this is a bad fit for the other commands I'll refactor.

jerbly: Oh, wait, I think you're talking about the call below what you've highlighted. Instead finding out if it's JSONL find out if the format requires multiple calls to generate. Got it.

crates/weaver_forge/src/output_processor.rs
jsuereth: I'd make it free-standing and hidden to the file, and if we want to publicize it later we can easily do so without moving it around.

crates/weaver_forge/src/output_processor.rs
jsuereth: Yeah - I'm relying on diagnostic output for the weaver-package tests, hence why I'm really curious how we do this going forward :)

PR #1160, opened 2026-01-26
Allow resolving from resolved schema using Manifest file.

### LINKED ISSUE #1134: Support resolved schema as input
We should expect resolved schema to be provided in `--registry`  and `--baseline-registry` options across all commands and should heavily optimize for it (no resolution, no validation beyond deserialization).

It could makes sense to support different inputs:
- `http://otel.io/schemas/1.future.0`  - link/path to manifest - this  is convenience for now, but can let us evolve/optimize manifest format
- `https://link-to-resolved-schema.yaml` 

Manifest/resolved schema should also be importable in multi-registry scenarios.
Definition/source schema can still be supported, but we should update docs to encourage using resolved schema whenever possible.

This only makes sense for v2 scenarios.

jsuereth:
Fixes #1134

Should allow specifying  a schema like in [OTEP 4815](https://github.com/open-telemetry/opentelemetry-specification/pull/4815).

- Can specify a manifest instead of a directory, e.g. `http://my-schema.org/manifest/1.0.0`
- Will use the `resolved_schema_url`, if available, on that manifest to resolve schema instead of grabbing `.yaml` files from the directory.
- Allow relative URLs for `resolved_schema_url`, although this needs to be tested a bit.

Thing not working yet:

- Extending groups from V2
- Actually using `manifest.yaml` vs. `registry_manifest.yaml` as per OTEP 4815.


crates/weaver_resolver/src/dependency.rs
```diff
@@ -133,11 +123,230 @@ impl ImportableDependency for V1Schema {
...
+                                sampling_relevant: None,
+                                note: attr.common.note.clone(),
+                                stability: Some(attr.common.stability.clone()),
+                                deprecated: attr.common.deprecated.clone(),
+                                prefix: false,
+                                tags: None,
+                                annotations: Some(attr.common.annotations.clone()),
+                                value: None,
+                                role: None,
+                            })
```
jerbly: nit: a helper function would be good for this, it's repeated a few times.

crates/weaver_common/src/vdir.rs
```diff
@@ -130,6 +130,60 @@ pub enum VirtualDirectoryPath {
...
+            LocalFolder { path } => LocalFolder { path: f(path) },
+            LocalArchive { path, sub_folder } => LocalArchive {
+                path,
+                sub_folder: {
+                    let result = f(sub_folder.unwrap_or("".to_owned()));
+                    if result.is_empty() {
+                        None
+                    } else {
+                        Some(result)
+                    }
```
jerbly: nit: this could be extracted to a function - it's duplicated.
double-nit: `unwrap_or("".to_owned())` can be `unwrap_or_default()`

crates/weaver_resolver/src/loader.rs
```diff
@@ -237,12 +235,36 @@ fn load_semconv_repository_recursive(
...
+    match from_vdir(path) {
+        Ok(resolved) => WResult::Ok(LoadedSemconvRegistry::ResolvedV2(resolved)),
+        Err(err) => WResult::FatalErr(err),
+    }
+}
+
+/// Reads a serialized object with serde from the given virtual directory path.
+fn from_vdir<T: serde::de::DeserializeOwned>(f: &VirtualDirectoryPath) -> Result<T, Error> {
+    let path = VirtualDirectory::try_new(f).map_err(|e| Error::InvalidUrl {
+        url: format!("{f}"),
```
jerbly:
```suggestion
        url: f.to_string(),
```

crates/weaver_resolver/src/dependency.rs
```diff
@@ -133,11 +123,230 @@ impl ImportableDependency for V1Schema {
...
+                            prefix: false,
+                            tags: None,
+                            annotations: Some(attr.common.annotations.clone()),
+                            value: None,
+                            role: Some(weaver_semconv::attribute::AttributeRole::Descriptive),
+                        }));
+                    }
+                    Group {
+                        id: e.id().to_owned(),
+                        r#type: weaver_semconv::group::GroupType::Event,
```
jerbly:
```suggestion
                        r#type: weaver_semconv::group::GroupType::Entity,
```

crates/weaver_semconv/src/registry_repo.rs
```diff
@@ -75,9 +78,38 @@ impl RegistryRepo {
...
+        let manifest = self.manifest.as_ref()?;
+        let resolved_url: &str = manifest.resolved_schema_url.as_ref()?;
+        match get_path_type(resolved_url) {
+            weaver_common::PathType::RelativePath => {
+                // We need to understand if the manifest URL is the same as the registry URL.
+                let vdir_was_manifest_file = self.manifest_path()? == self.registry.path();
+                Some(self.registry.vdir_path().map_sub_folder(|path| {
+                    if vdir_was_manifest_file {
+                        match Path::new(&path).parent() {
+                            Some(parent) => format!("{}/{resolved_url}", parent.display()),
```
jerbly, line 93: Is this right? `parent.display()` will use os dependent separators, so backslashes in Windows.

crates/weaver_resolver/src/error.rs
```diff
@@ -13,6 +13,27 @@ use weaver_semconv::provenance::Provenance;
 #[must_use]
 #[non_exhaustive]
 pub enum Error {
+    /// There was an issue resolving definition schema.
+    #[error(transparent)]
+    #[diagnostic(transparent)]
+    FailToResolveDefinition(#[from] weaver_semconv::Error),
```
jerbly, line 19: The `#[from]` here will make any `weaver_semconv::Error` a `FailToResolveDefinition` error even if it has nothing to do with resolving definitions.

jerbly (commented):
This is a big one... I'm going to try this branch out. In the meantime, I found a couple of bugs and some nits.

crates/weaver_semconv/src/registry_repo.rs
jsuereth: It will only do that if it's parsed or you use `join`.  I wrote this on windows, and yes it's a bit awkward.  Probably the thing that needs the MOST attention in this PR.

If you want me to refactor this, I'd be happy to try a bit.

crates/weaver_resolver/src/error.rs
jsuereth: The only errors that (should) be coming from that crate, in this context, are things relating to resolving definitions.

The underlying error should still be reported, but you have context that the error in question caused resolver to fail.

crates/weaver_common/src/vdir.rs
jsuereth: done

crates/weaver_resolver/src/dependency.rs
jsuereth: Repeated... with subtlety.   I'll work on refactoring this.  PTAL

crates/weaver_semconv/src/manifest.rs
```diff
@@ -68,7 +68,11 @@ pub struct RegistryManifest {
 pub struct Dependency {
     /// The name of the dependency.
     pub name: String,
-    /// The registry path of the dependency.
+    /// The path to the dependency.
+    ///
+    /// This can be either:
+    /// - A manifest of a published registry
+    /// - A directory containing the raw definition.
     pub registry_path: VirtualDirectoryPath,
```
lmolkova, line 76: can we call it accordingly? could be confusing to see `registry_path: http://otel.io/schemas` ...

```suggestion
    pub registry_location: VirtualDirectoryPath,
```

or 


```suggestion
    pub registry: VirtualDirectoryPath,
```

crates/weaver_resolver/src/error.rs
```diff
@@ -13,6 +13,27 @@ use weaver_semconv::provenance::Provenance;
...
+        registry_id: String,
+        /// A string representing the dependency chain.
+        chain: String,
+    },
+
+    /// We've reached the maximum dependency depth for this registry.
+    #[error("Maximum dependency depth reached for registry `{registry}`. Cannot load further dependencies.")]
+    MaximumDependencyDepth {
+        /// The registry which has too many dependencies.
+        registry: String,
```
lmolkova, line 34: can we align on `registry` vs `registry_id` (above) vs `registry_url` ?

crates/weaver_resolver/src/loader.rs
jsuereth: Fixed

crates/weaver_semconv/src/manifest.rs
jsuereth: Since manifest work today and folks use them, I didn't want to issue breaking changes.

I'm happy to move to something else and declare an `alias` here so we don't break.

Personally, I prefer `registry_uri` or `registry_location` better.  The key thing to denote is this is a "virtual directory path" in Weaver vernacular, so it can point at directories *inside* a git repository or zip archive.


Should I update that now?

crates/weaver_resolver/src/error.rs
jsuereth: Sure - I'm fine with either.  Which do you think makes a better error message?  URL or ID?

crates/weaver_semconv/src/manifest.rs
lmolkova: now is great, I can also do it in https://github.com/open-telemetry/weaver/pull/1106 (`registry_uri` sounds good and we'd align `registy_url` with it)

crates/weaver_resolver/src/error.rs
lmolkova: I was playing with it in the https://github.com/open-telemetry/weaver/pull/1106 and I'm thinking:

- registry_url | uri is required and necessary
- `id` is now equivalent to name, it's not essential - can be optional on dependency

what if we had `name()` method that'd return name if it's known or fell back to url if not. We'd use it when reporting dependency errors and then called it `registry_name` (since that's what it called on dependency and in the manifest). 

OR we should rename `dependency.name` and `manifest.name` to `id` and then use registry_id everywhere

jsuereth: So - we need a "name" that's separate from "version" so we can isolate if we have a repeated/diamond dependency.  That's actually orthogonal to this discussion, but made me realize we may have a bug in our dependency tracking on that front.

I'm a fan of `name()` method that falls back from human readable to the `url`.  I need to sort through the whole codebase for `id` and fix it.

Do you think we can do that in a separate PR or do you want to clean that up now?

lmolkova: yeah, let's do it as a follow-up - that'd be easier to review

lmolkova (approved):

crates/weaver_semconv/src/manifest.rs
jsuereth: Since this is a rather large PR - I'd rather do a dedicate PR for that change.

I think we could also look at moving from `registry_manifest.yaml` -> `manifest.yaml` in that one.

crates/weaver_resolver/src/error.rs
jsuereth: Yes! Let's do that in a follow up PR - This gets enough done I can start fragmenting my work into separate branches again without everything conflicting.

crates/weaver_resolver/src/dependency.rs
```diff
@@ -130,14 +121,210 @@ impl ImportableDependency for V1Schema {
...
+                    deprecated: m.common.deprecated.clone(),
+                    attributes: m
+                        .attributes
+                        .iter()
+                        .map(|ar| {
+                            // TODO - this should be non-panic errors.
+                            let attr = self
+                                .attribute_catalog
+                                .attribute(&ar.base)
+                                .expect("Unable to find attr on catalog, invalid registry!");
```
jerbly: Here's an example of a panic.

crates/weaver_resolver/src/dependency.rs
```diff
@@ -130,14 +121,210 @@ impl ImportableDependency for V1Schema {
...
+                    deprecated: e.common.deprecated.clone(),
+                    attributes: e
+                        .attributes
+                        .iter()
+                        .map(|ar| {
+                            // TODO - this should be non-panic errors.
+                            let attr = self
+                                .attribute_catalog
+                                .attribute(&ar.base)
+                                .expect("Unable to find attr on catalog, invalid registry!");
```
jerbly: Another panic

crates/weaver_resolver/src/dependency.rs
```diff
@@ -130,14 +121,210 @@ impl ImportableDependency for V1Schema {
...
+                    include_all || entities_imports_matcher.is_match(entity_type)
+                })
+                .map(|e| {
+                    let mut attributes = vec![];
+                    for ar in e.identity.iter() {
+                        // TODO - this should be non-panic errors.
+                        let attr = self
+                            .attribute_catalog
+                            .attribute(&ar.base)
+                            .expect("Unable to find attr on catalog, invalid registry!");
```
jerbly: Another panic

crates/weaver_resolver/src/dependency.rs
```diff
@@ -130,14 +121,210 @@ impl ImportableDependency for V1Schema {
...
+                            ar.requirement_level.clone(),
+                            Some(AttributeRole::Identifying),
+                        )));
+                    }
+                    for ar in e.description.iter() {
+                        // TODO - this should be non-panic errors.
+                        let attr = self
+                            .attribute_catalog
+                            .attribute(&ar.base)
+                            .expect("Unable to find attr on catalog, invalid registry!");
```
jerbly: Another panic

jsuereth: Thanks! Fixed

crates/weaver_resolver/src/dependency.rs
jsuereth: Thanks! Fixed

crates/weaver_resolver/src/dependency.rs
jsuereth: Thanks! Fixed

crates/weaver_resolver/src/dependency.rs
jsuereth: Thanks! Fixed

PR #1171, opened 2026-01-31
Fix for gix 0.78 update

jerbly:
Replaces #1158 

Updating gix to 0.78 introduced aws-lc-rs as the default crypto backend for rustls. When cross-compiling from glibc to musl (Alpine), aws-lc-sys C code references __isoc23_* functions that don't exist in musl, causing linker errors.

~~I tried a bunch of different work arounds but with no luck. So this PR switches from Alpine to Debian trixie-slim for the runtime image, and target *-unknown-linux-gnu instead of *-unknown-linux-musl.~~

~~Runtime image size increases from 5MB to 70MB. :(
It also means that the musl builds we just added to the cargo-dist config will also have to be removed.~~

**UPDATE:**
@jsuereth - I followed your idea:
- Switched from rustls to native-tls with vendored OpenSSL
- cross-arch-build.sh detects the CPU architecture and configures the right musl toolchain
(FYI - Claude helped a lot!)

Docker image size is now ~60MB (previously ~50MB), Debian slim version was ~180MB

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -215,7 +215,7 @@ jobs:
           | sarif-fmt
         continue-on-error: true
       - name: Upload
-        uses: github/codeql-action/upload-sarif@19b2f06db2b6f5108140aeb04014ef02b648f789 # v4.31.11
+        uses: github/codeql-action/upload-sarif@b20883b0cd1f46c72ae0ba6d1090936928f9fa30 # v4.32.0
         with:
           sarif_file: clippy-results.sarif
           wait-for-processing: true
--- a/.github/workflows/ossf-scorecard.yml
+++ b/.github/workflows/ossf-scorecard.yml
@@ -42,6 +42,6 @@ jobs:
       # Upload the results to GitHub's code scanning dashboard (optional).
       # Commenting out will disable upload of results to your repo's Code Scanning dashboard
       - name: "Upload to code-scanning"
-        uses: github/codeql-action/upload-sarif@19b2f06db2b6f5108140aeb04014ef02b648f789 # v4.31.11
+        uses: github/codeql-action/upload-sarif@b20883b0cd1f46c72ae0ba6d1090936928f9fa30 # v4.32.0
         with:
           sarif_file: results.sarif
--- a/.github/workflows/publish-docker.yml
+++ b/.github/workflows/publish-docker.yml
@@ -43,7 +43,7 @@ jobs:
       run: |
           docker run --rm ${{ env.TEST_WEAVER_TAG }} --help
     - name: Log in to Docker Hub
-      uses: docker/login-action@5e57cd118135c172c3672efd75eb46360885c0ef # v3.6.0
+      uses: docker/login-action@c94ce9fb468520275223c153574b00df6fe4bcc9 # v3.7.0
       if: github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/tags/')
       with:
         username: ${{ secrets.DOCKER_USERNAME }}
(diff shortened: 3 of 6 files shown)


jsuereth:
@jerbly This is a HUGE change.

Thanks so much for looking into this.

I really do NOT want to lose the MUSL build.  The huge increase in container size seems problematic to me.  I was looking into alternatives, and wanted to propose a few:

- WDYT about moving to use OpenSSL here instead of the AWS fork?
- I was looking at boring crypto as another option, but we lack support through gix for that.

I think we COULD update the `alpine` build image to include libssl if needed, so I may take what you have here and explore that option.

cross-arch-build.sh
```diff
@@ -1,19 +1,53 @@
...
-  RUST_TARGET=aarch64-unknown-linux-musl
-  if [ "${TARGETPLATFORM}" != "${BUILDPLATFORM}" ]; then
-    apt-get update && apt-get install -y gcc-aarch64-linux-gnu
-  fi
-else
-  echo "Unsupported target platform: ${TARGETPLATFORM}"
-  exit 1
-fi
+# Install build dependencies
+apt-get update && apt-get install -y musl-tools musl-dev perl
```
jsuereth: Nit - we should probably move this into the Dockerfile. I have a PR for fixing this build for windows where I may add that one

cross-arch-build.sh
```diff
@@ -1,19 +1,53 @@
...
+  linux/amd64)
+    RUST_TARGET=x86_64-unknown-linux-musl
+    if [ "${HOST_ARCH}" = "x86_64" ]; then
+      # Native x86_64 - musl-tools provides musl-gcc
+      export CC_x86_64_unknown_linux_musl=musl-gcc
+      export CARGO_TARGET_X86_64_UNKNOWN_LINUX_MUSL_LINKER=musl-gcc
+    else
+      # Cross-compiling to x86_64 from another arch
+      apt-get install -y wget
+      wget -q https://musl.cc/x86_64-linux-musl-cross.tgz
```
jsuereth, line 19: Nit: Can you add a note that we should be verifying this artifact via a checksum if we can?

cross-arch-build.sh
```diff
@@ -1,19 +1,53 @@
...
+  linux/arm64*)
+    RUST_TARGET=aarch64-unknown-linux-musl
+    if [ "${HOST_ARCH}" = "aarch64" ]; then
+      # Native arm64 - musl-tools provides musl-gcc
+      export CC_aarch64_unknown_linux_musl=musl-gcc
+      export CARGO_TARGET_AARCH64_UNKNOWN_LINUX_MUSL_LINKER=musl-gcc
+    else
+      # Cross-compiling to arm64 from x86_64
+      apt-get install -y wget
+      wget -q https://musl.cc/aarch64-linux-musl-cross.tgz
```
jsuereth, line 35: Nit: Same comment about verifying checksums.

jsuereth (approved):
Woot!  Very nice change / fix here, thanks for doing this.

The 10MB boost is crazy for SSL, but I guess that's the ropes.  In my experiments I was looking and I think the three options were OpenSSL, AWS's SSL and Google's "boringcrypto" (which had lots of caveats to not rely on it).  It seems *gix* itself didn't give an option to use boring-crypto, but reqwest does.  I also wasn't sure if that one would also have issues with MUSL.

In any case, glad we have something working now!  I think at some point we may want a way to swap between crypto libraries, but for now, let's stick with one that work!

cross-arch-build.sh
jerbly: Done

cross-arch-build.sh
jerbly: added a checksum check

cross-arch-build.sh
jerbly: added a checksum check

PR #1173, opened 2026-02-01
Switch to `pnpm` and force its usafe for UI - Remove `npm` calls from `build.rs`

jsuereth:
This Fixes a few things in the build.

- First it forces a specific version for node and using `pnpm` as a package manager.
- Second, it only issues warnings from `build.rs` if node is not build, and warnings if it is not up to date. It should not longer try to call node executables from inside cargo.
- Last, it tries to do more fine-grained detection of staleness checks in build.rs to give better guidance on what you need to call to fix things.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -43,7 +43,7 @@ fn is_package_lock_stale(dir: &Path) -> Result<bool, Box<dyn std::error::Error>>
     if package_timestamp > lock_timestamp {
         return Ok(true);
     }
-    return Ok(false);
+    Ok(false)
 }
 
 /// Helper function to determine if NPM project is out of date.
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -131,12 +131,20 @@ jobs:
             curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
             echo "$HOME/.cargo/bin" >> $GITHUB_PATH
           fi
+      - name: "Setup pnpm"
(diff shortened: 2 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -18,6 +18,8 @@ Run `cargo test --all` to run the tests.
 
 Run `cargo build` to build a local binary for any additional tests. The resulting binary will be placed in the `./output` directory.
 
+NOTE: You will need to run `pnpm build` in the `ui` directory prior to running any `cargo` commands.
```
jerbly, line 21: should we add it to the `justfile`?

jsuereth: huh, sure. I don't use JUST but I can do that.

jsuereth: Actually - I don't use just, @jerbly Could you do that in a follow-up? I'm not really sure what I should be doing in justfiles.

build.rs
```diff
@@ -26,56 +26,64 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+fn check_ui() -> Result<(), Box<dyn std::error::Error>> {
+    let ui_dir = Path::new("ui");
 
-    if !status.success() {
-        return Err("Failed to build UI".into());
+    // Check if UI is out of date before running.
+    if is_ui_stale(ui_dir)? {
+        return Err(
+            "Weaver UI is out of date. Please run `pnpm build` in the `ui` directory.".into(),
+        );
```
lquerel, line 79: Cool!

PR #1199, opened 2026-02-13
[chore] Fix rennovate - don't fail in build.rs for UI.

jsuereth:
This allows `cargo` commands to work without a UI - so rennovate can do dependency fixes.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -75,13 +75,13 @@ fn check_ui() -> Result<(), Box<dyn std::error::Error>> {
     // Check if UI is out of date before running.
     if is_ui_stale(ui_dir)? {
         println!("cargo::warning=Weaver UI is out of date. Please run `pnpm build` in the `ui` directory.");
-        return Ok(())
+        return Ok(());
     }
 
     // Check if we need to install packages.
     if is_package_lock_stale(ui_dir)? {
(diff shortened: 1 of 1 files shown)


build.rs
```diff
@@ -74,15 +74,14 @@ fn check_ui() -> Result<(), Box<dyn std::error::Error>> {
 
     // Check if UI is out of date before running.
     if is_ui_stale(ui_dir)? {
-        return Err(
-            "Weaver UI is out of date. Please run `pnpm build` in the `ui` directory.".into(),
-        );
+        println!("cargo::warning=Weaver UI is out of date. Please run `pnpm build` in the `ui` directory.");
+        return Ok(())
```
jerbly:
```suggestion
        return Ok(());
```

build.rs
```diff
@@ -74,15 +74,14 @@ fn check_ui() -> Result<(), Box<dyn std::error::Error>> {
...
     }
 
     // Check if we need to install packages.
     if is_package_lock_stale(ui_dir)? {
-        // TODO - Disable for CI
-        return Err("Weaver `ui/pnpm-lock.yaml` is out of date. Please run `pnpm install` in the `ui` directory.".into());
+        println!("cargo::warning=Weaver `ui/pnpm-lock.yaml` is out of date. Please run `pnpm install` in the `ui` directory.");
+        return Ok(())
```
jerbly:
```suggestion
        return Ok(());
```

PR #1200, opened 2026-02-14
Unifies output handling across all registry subcommands

jerbly:
Unifies output handling across all registry subcommands. Builtin output formats (json, yaml, jsonl) are now available in `registry stats`, `registry diff`, and `registry resolve`. `registry stats` also supports `--templates` for custom text output templates.

No breaking changes to the CLI. There is a future opportunity to unify the CLI with common terms following this PR.

- Replaces scattered format handling across all registry subcommands with `OutputProcessor`, a unified output abstraction
- `OutputTarget` enum models output destinations (Stdout/Stderr/File/Directory/Mute) with ergonomic helpers:  `from_optional_dir` and `from_optional_file`
- Makes `TemplateEngine` and `OutputDirective` crate-internal — external consumers go through `OutputProcessor`
- Converts `registry stats` text format from hardcoded `println!` to a Jinja template
  - (Required the addition of `numsort` template filter for numeric key ordering)
- Removes redundant `diff` JSON/YAML templates now handled by builtin serialization

### DIFF
--- a/crates/weaver_forge/src/output_processor.rs
+++ b/crates/weaver_forge/src/output_processor.rs
@@ -397,9 +397,8 @@ mod tests {
 
     #[test]
     fn test_mute_format() {
-        let output =
-            OutputProcessor::new("mute", "test", None, None, OutputTarget::Stdout)
-                .expect("mute format should succeed");
+        let output = OutputProcessor::new("mute", "test", None, None, OutputTarget::Stdout)
+            .expect("mute format should succeed");
         assert!(!output.is_file_output());
         assert!(!output.is_line_oriented());
     }
@@ -450,7 +449,9 @@ mod tests {
         assert!(!output.is_line_oriented());
         assert!(output.is_file_output());
 
-        output.generate(&test_data()).expect("generate should succeed");
+        output
+            .generate(&test_data())
+            .expect("generate should succeed");
 
         let file_path = path.join("test.json");
         let content = fs::read_to_string(&file_path).expect("should read output file");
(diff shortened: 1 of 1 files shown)


crates/weaver_forge/src/output_processor.rs
```diff
@@ -26,6 +64,16 @@ enum BuiltinFormat {
 }
 
 impl BuiltinFormat {
+    /// Try to parse a format name (already lowercased) into a builtin format.
+    fn from_name(name: &str) -> Option<Self> {
+        match name {
+            "json" => Some(BuiltinFormat::Json),
```
jsuereth, line 70: Something I was thinking for the existing weaver-packages test script -

Should we allow JQ filter expressions when we do built-in?  E.g. I have a few JQ filter expressions I'm using to only pull in rego template output and I had to create a new diagnostic template for this, which MAY have the same JSON/YAML bug we saw previously.

crates/weaver_forge/src/output_processor.rs
```diff
@@ -349,10 +452,38 @@ mod tests {
...
+        let temp_dir = TempDir::new().unwrap();
+        let file_path = temp_dir.path().join("output.json");
+        let mut output = OutputProcessor::new(
+            "json",
+            "test",
+            None,
+            None,
+            OutputTarget::File(file_path.clone()),
+        )
+        .unwrap();
```
jsuereth: Nit: Let's use `expect` instead of `unwrap`.

crates/weaver_forge/src/output_processor.rs
jerbly: Of course you could pre-process before outputting. But I guess we can include the pre-processing in the OutputProcessor? Let's do this in a subsequent PR.

As I said in the PR description, with this unification in place there's now more opportunity to standardize across subcommands including breaking changes to the CLI which would tidy up the interface. Perhaps we do this with that?

Issue #182, opened 2024-05-29
Allow specifying a registry for a given tag or branch


lquerel:
It seems that the most commonly used syntax for tags is the git URL followed by a #<tag-id>.

MadVikingGod:
Is your feature request related to a problem? Please describe.
Add a way to fetch a repository at a specific git tag. For example weaver registry resolve -r  https://github.com/open-telemetry/semantic-conventions.git --tag v1.25.0

Describe the solution you'd like
By specifying a tag, weaver should not fetch main but instead the tag specified.

Describe alternatives you've considered
Currently fetching the repository manually and checking out the tag requested.

lquerel:
In PR #257, the following syntax is now supported: `source[subfolder]`, and an extension for tag, branch, and commit will take the following form: `source@refspec[sub_folder]`.

I keep this issue open as the refspec notation is not yet fully supported.

sebasnabas:
Hi, I'd like to help out, so I've opened PR #1201

jerbly:
Support was added in weaver v0.22.0

PR #1201, opened 2026-02-14
Implement git refspec

sebasnabas:
Hi, I'd like to contribute to OpenTelemetry, so here's my first PR :)

This PR implements #182.
gix supports refspecs with the [`with_ref_name`](https://docs.rs/gix/0.78.0/gix/clone/struct.PrepareFetch.html#method.with_ref_name) method.
I've updated the related tests and tested it manually with git tags.


tests/custom_registry/registry_manifest.yaml
```diff
@@ -4,4 +4,4 @@ semconv_version: 0.1.0
 schema_base_url: https://acme.com/schemas/
 dependencies:
   - name: otel
-    registry_path: https://github.com/open-telemetry/semantic-conventions/archive/refs/tags/v1.30.0.zip[model]
+    registry_path: https://github.com/open-telemetry/semantic-conventions@v1.30.0[model]
```
jsuereth, line 7: nit: I think this would be `https://github.com/open-telemetry/semantic-conventions.git@v1.30.0[model]`

sebasnabas: I found that both with and without `.git` suffix work (through a debug log statement in `try_from_git_url`)

sebasnabas:
@jsuereth I just rebased and fixed the conflict.

PR #1202, opened 2026-02-14
Manifest renames

### LINKED ISSUE #1197: Update naming conventions for multi-registry manifests
As part of #1160 - We discussed certain differences / naming conventions from what Weaver has today vs. what we want in the OTEP.  A few points to nail down and change:

- Migrating from `registry_manifest.yaml` to just `manifest.yaml`.  We should (in a non-breaking way) move to the simpler named file.
- Consistency in referencing registries.  We have `_id`, `_uri` and `_name` today, we need to sort out what these mean.
- Dependency Resolution needs to understand if we have a "diamond dependency".  This means we need to see if we're depending on "the same" registry but two different versions.

## A set of principles:

- We need a `name` for a registry.  This should be unique *across all registries*, and can be used to denote if we have a dependency chain with two versions of the "same" registry.
- We need a `uri` for a registry.  This should denote the Manifest file *or* the directory where the manifest file can be found.
- We need a `uri` for where to find resolved schema.  This should denote where we load the resolved schema and will be a `VirtualDirectoryRef`. (Note: `VirtualDirectoryRef` can be a file in a virtual directory, e.g. a file in a git repo or a file inside a zip archive).

## Strawman Proposal:

- Allow `registry_manfiest.yaml` or `manfiest.yaml` to be the name of a registry manifest.
  - Prefer `manifest.yaml` if it exists.
  - Issue a warning if we see `registry_manifest.yaml` is used.
- We consistently use "Weaver URI" in documentation and code to represent `VirtualDirectoryRef` strings.
- Change `RegistryManifest` as follows:
  - `resolved_schema_url` becomes `resolved_schema_uri`.
 - `name` is removed.
 - `repository_url` is changed to be `schema_url` - I.e. this is the schema_url this registry will fill in telemetry.
- `name` of a registry will be inferred from `schema_url` - We drop the "version" component of the URL, and we get a unique name we can use to identify version conflicts in dependencies.

lmolkova:
Fixes #1197

Renames:

- `registry_manifest.yaml` is now deprecated (but supported), use `manifest.yaml` instead
- `schema_url` is now used consistently in resolved schema  and manifest instead of registry_id, registry_url, ... . It can be a source of the registry contents, but does not have to be. When specifying dependencies, it's allowed to pass `registry_path` in addition to `schema_url`. This way, schema url is essentially a combination of name and version in a certain format.
- `resolved_schema_url` -> `resolved_schema_uri`

Changes should be backward compatible, old properties of `RegistryManifest` and `Dependency` are supported, but deprecated

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -259,7 +259,7 @@ jobs:
           | sarif-fmt
         continue-on-error: true
       - name: Upload
-        uses: github/codeql-action/upload-sarif@9e907b5e64f6b83e7804b09294d44122997950d6 # v4.32.3
+        uses: github/codeql-action/upload-sarif@89a39a4e59826350b863aa6b6252a07ad50cf83e # v4.32.4
         with:
           sarif_file: clippy-results.sarif
           wait-for-processing: true
--- a/.github/workflows/ossf-scorecard.yml
+++ b/.github/workflows/ossf-scorecard.yml
@@ -42,6 +42,6 @@ jobs:
       # Upload the results to GitHub's code scanning dashboard (optional).
       # Commenting out will disable upload of results to your repo's Code Scanning dashboard
       - name: "Upload to code-scanning"
-        uses: github/codeql-action/upload-sarif@9e907b5e64f6b83e7804b09294d44122997950d6 # v4.32.3
+        uses: github/codeql-action/upload-sarif@89a39a4e59826350b863aa6b6252a07ad50cf83e # v4.32.4
         with:
           sarif_file: results.sarif
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -139,9 +139,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.101"
+version = "1.0.102"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5f0e0fee31ef5ed1ba1316088939cea399010ed7731dba877ed44aeb407a75ea"
+checksum = "7f202df86484c868dbad7eaa557ef785d5c66295e41b460ef922eca0723b842c"
 
 [[package]]
 name = "arbitrary"
@@ -494,9 +494,9 @@ checksum = "613afe47fcd5fac7ccf1db93babcb082c5994d996f20b8b159f2ad1658eb5724"
 
 [[package]]
 name = "chrono"
-version = "0.4.43"
+version = "0.4.44"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "fac4744fb15ae8337dc853fee7fb3f4e48c0fbaa23d0afe49c447b4fab126118"
+checksum = "c673075a2e0e5f4a1dde27ce9dee1ea4558c7ffe648f576438a20ca1d2acc4b0"
 dependencies = [
  "iana-time-zone",
  "js-sys",
@@ -2977,9 +2977,9 @@ dependencies = [
 
 [[package]]
 name = "minijinja"
-version = "2.15.1"
+version = "2.16.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b479616bb6f0779fb0f3964246beda02d4b01144e1b0d5519616e012ccc2a245"
+checksum = "5c54f3bcc034dd74496b5ca929fd0b710186672d5ff0b0f255a9ceb259042ece"
 dependencies = [
  "aho-corasick",
  "memo-map",
@@ -2991,9 +2991,9 @@ dependencies = [
 
 [[package]]
 name = "minijinja-contrib"
-version = "2.15.1"
+version = "2.16.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7826089e6af7bc638f69a44b100ebe7f6c64b182cfde16558d5cd38ac8adde20"
+checksum = "86201a560fde90bf270d3c9d5da0c2a5a0c25bd3242d674fcf03e6a35fcfa7ba"
 dependencies = [
  "minijinja",
  "serde",
--- a/justfile
+++ b/justfile
@@ -8,7 +8,7 @@ install:
     cargo install cargo-check-external-types@0.4.0 --locked
     cargo install git-cliff@2.12.0 --locked
     cargo install cargo-tarpaulin@0.35.2 --locked
-    cargo install cargo-nextest@0.9.128 --locked
+    cargo install cargo-nextest@0.9.129 --locked
 
 pre-push-check:
     cargo clean
--- a/ui/package.json
+++ b/ui/package.json
@@ -3,7 +3,7 @@
   "private": true,
   "version": "0.0.1",
   "type": "module",
-  "packageManager": "pnpm@10.29.3",
+  "packageManager": "pnpm@10.30.1",
   "scripts": {
     "dev": "vite",
     "build": "vite build",
@@ -14,7 +14,7 @@
     "autoprefixer": "10.4.24",
     "daisyui": "4.12.24",
     "postcss": "8.5.6",
-    "svelte": "5.50.3",
+    "svelte": "5.53.0",
     "tailwindcss": "3.4.19",
     "vite": "7.3.1"
   },
--- a/ui/pnpm-lock.yaml
+++ b/ui/pnpm-lock.yaml
@@ -10,7 +10,7 @@ importers:
     dependencies:
       marked:
         specifier: ^17.0.1
-        version: 17.0.2
+        version: 17.0.3
       rapidoc:
         specifier: ^9.3.8
         version: 9.3.8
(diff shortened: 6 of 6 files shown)


crates/weaver_semconv/src/schema_url.rs
```diff
@@ -0,0 +1,409 @@
...
+                return authority;
+            }
+
+            format!("{}/{}", authority, segments.join("/"))
+        })
+    }
+
+    /// Returns the registry version, derived from the schema URL.
+    #[must_use]
+    pub fn version(&self) -> &str {
```
jsuereth, line 81: FYI - We may want to move to using "semver" directly here, and close #1108 for this.

crates/weaver_semconv/src/schema_url.rs
```diff
@@ -0,0 +1,409 @@
...
+
+impl Eq for SchemaUrl {}
+
+impl std::hash::Hash for SchemaUrl {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        self.url.hash(state);
+    }
+}
+
+impl std::fmt::Display for SchemaUrl {
```
jsuereth, line 129: I do wish these derivation macros supported helpers to avoid this kind of trivial impl,  At least they're not hard to write....

crates/weaver_semconv_gen/src/v2.rs
```diff
@@ -451,8 +452,7 @@ mod tests {
     fn test_registry() -> ResolvedTelemetrySchema {
         ResolvedTelemetrySchema {
             file_format: "resolved/2.0.0".to_owned(),
-            schema_url: "todo/1.0.0".to_owned(),
-            registry_id: "main".to_owned(),
+            schema_url: SchemaUrl::try_new("https://todo/1.0.0".to_owned()).unwrap(),
```
jsuereth: Nit: If you're feeling particularly rusty, you could `impl TryFrom<&str> for SchemaUrl` here, and then we can have:

```
schema_url: "https://todo/1.0.0".try_into().expect("Should be valid schema url")
```

crates/weaver_semconv/src/manifest.rs
```diff
@@ -60,20 +69,126 @@ pub struct RegistryManifest {
...
+    #[serde(default, skip_serializing)] // we can read, but won't write this field
+    pub name: Option<String>,
+}
+
+impl<'de> Deserialize<'de> for Dependency {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        #[derive(Deserialize)]
```
jsuereth, line 96: This is clever!

crates/weaver_semconv/src/manifest.rs
```diff
@@ -60,20 +69,126 @@ pub struct RegistryManifest {
...
+
+    /// This field is deprecated and should not be used.
+    /// The registry name should be derived from the `schema_url` field,
+    /// which serves as a unique identifier for the dependency registry
+    /// and includes registry version.
+    #[deprecated(
+        note = "The `name` field is deprecated. The registry name should be derived from the `schema_url` field, which serves as a unique identifier for the dependency registry."
+    )]
+    #[serde(default, skip_serializing)] // we can read, but won't write this field
+    pub name: Option<String>,
```
jsuereth: Nit: Is name even used at all when deserializing? Can we remove it?

crates/weaver_semconv/src/manifest.rs
```diff
@@ -100,45 +215,31 @@ impl RegistryManifest {
...
-                error: "The registry schema base URL is required.".to_owned(),
-            });
-        }
-
-        handle_errors(errors)?;
+    /// Returns the registry name, which is derived from the schema URL.
+    /// For example, if the schema URL is `https://opentelemetry.io/schemas/sub-component/1.0.0`,
+    /// the registry name would be `opentelemetry.io/schemas/sub-component`
+    #[must_use]
+    pub fn name(&self) -> String {
```
jsuereth: Nit: This should probably return `&str` and we `to_owned()` where it's used.

crates/weaver_semconv/src/manifest.rs
```diff
@@ -100,45 +215,31 @@ impl RegistryManifest {
...
+        self.schema_url.name().to_owned()
+    }
 
-        Ok(())
+    /// Returns the registry version, which is derived from the schema URL.
+    /// For example, if the schema URL is `https://opentelemetry.io/schemas/sub-component/1.0.0`,
+    /// the registry version would be `1.0.0`
+    #[must_use]
+    pub fn version(&self) -> String {
```
jsuereth: Same as above on returning `&str`

crates/weaver_resolver/src/loader.rs
```diff
@@ -36,7 +36,7 @@ pub enum LoadedSemconvRegistry {
         dependencies: Vec<LoadedSemconvRegistry>,
     },
     /// The semconv repository is already resolved and can be used as-is.
-    Resolved(V1Schema),
+    Resolved(Box<V1Schema>),
```
jsuereth: Is this due to the warning about a large variant type?

We may want to move all the schemas to be boxes here.  Just trying to understand the rationale.

crates/weaver_semconv/src/manifest.rs
lmolkova: great catch! no, we don't need to have it as an API property, same as schema_base_url and semconv_version on manifest, removed

crates/weaver_resolver/src/loader.rs
lmolkova: yes, this particular one turned out not to be necessary, but there is https://github.com/lmolkova/weaver/blob/8677b1bfa82e9d30918051111718806031fc768f/crates/weaver_live_check/src/lib.rs#L121 that clippy demands.

Are you saying to do it consistently for the `ResolvedRegistry` too?

```rs
pub enum VersionedRegistry {
    /// v1 ResolvedRegistry
    V1(ResolvedRegistry),
    /// v2 ForgeResolvedRegistry
    V2(Box<ForgeResolvedRegistry>),
}
```

crates/weaver_semconv/src/schema_url.rs
lmolkova: I'd prefer to follow up if that's ok - there are all the questions around URL format and it'd be much easier to review a specific change

crates/weaver_resolver/src/loader.rs
jsuereth: We'll likely need to, yes.

I'm not sure how one is too large compared to the other, but given how this is used, probably better to box both

crates/weaver_semconv/src/schema_url.rs
jsuereth: That was my intention here - just an FYI not a blocking comment.

crates/weaver_semconv/src/registry_repo.rs
```diff
@@ -34,30 +40,69 @@ pub struct RegistryRepo {
...
-            manifest: None,
-        };
-        if let Some(manifest) = registry_repo.manifest_path() {
-            let registry_manifest = RegistryManifest::try_from_file(manifest)?;
-            registry_repo.id = Arc::from(registry_manifest.name.as_str());
-            registry_repo.manifest = Some(registry_manifest);
+        let registry =
+            VirtualDirectory::try_new(registry_path).map_err(Error::VirtualDirectoryError)?;
+        // Try to load manifest
+        if let Some(manifest_path) = {
```
jsuereth: It may be worth moving the logic of manifest-path into a standalone function, and then re-using it here vs. creating a temporary RegistryRepo and just calling manifest_path on that.

crates/weaver_semconv/src/registry_repo.rs
```diff
@@ -34,30 +40,69 @@ pub struct RegistryRepo {
...
+                manifest: None,
+            })
         }
-        Ok(registry_repo)
     }
 
-    /// Returns the unique identifier for the registry.
+    /// Returns the registry name (from manifest if present, otherwise top-level field).
+    #[must_use]
+    pub fn name(&self) -> Arc<str> {
```
jsuereth: Nit - If you're not putting the SchemaURL in an `Arc`, then we should not force the name into one here.

Arc is about safe sharing between threads, and we'd need to do so for the entire SchemaUrl or RegistryRepo, likely.

crates/weaver_semconv/src/registry_repo.rs
```diff
@@ -34,30 +40,69 @@ pub struct RegistryRepo {
...
+        self.schema_url.name().into()
+    }
+
+    /// Returns the registry version (from manifest if present, otherwise top-level field).
     #[must_use]
-    pub fn id(&self) -> Arc<str> {
-        self.id.clone()
+    pub fn version(&self) -> Arc<str> {
```
jsuereth: Same nit - this doesn't need to return `Arc`, can probably return `&str`

crates/weaver_semconv/src/registry_repo.rs
```diff
@@ -125,6 +177,22 @@ impl RegistryRepo {
             None
         }
     }
+
+    /// Returns the registry schema URL.
+    #[must_use]
+    pub fn schema_url(&self) -> SchemaUrl {
```
jsuereth: I'm not sure where this is used, but possibly just return `&SchemaUrl` without `#[must_use]`, and we can decide to `clone` if we need to after calling this.

jsuereth (commented):
Overall, almost ready to merge, but a bunch of rust-related cleanups.

crates/weaver_semconv/src/registry_repo.rs
lmolkova: you detected AI slop that I missed :)

PR #1203, opened 2026-02-15
chore: Restore renovate scope

thompson-tomo:
Enable renovate to process all directories not just ui.

Note when you look at the renovate dashboard you would see that renovate is only managing the ui folder which was caused by the changes in #1173


jerbly (approved):
Oof! Missed that footgun, thanks!

Issue #1075, opened 2025-12-12
Figure out how to pin cargo-dist versions w/ rennovate


jsuereth:
How to pin dependencies in cargo-dist: https://axodotdev.github.io/cargo-dist/book/ci/customizing.html#pinned-actions-commits

Need to look into how to make this work with rennovate.

PR #1205, opened 2026-02-15
ci: Pin dependencies and manage via custom manager in renovate

thompson-tomo:
Closes: #1075

This pins github actions used by cargo dist and enables renovate to update them.

This also introduces explicit version for dependencies installed via the justfile which are also managed via renovate.

The rust toolchain version in both the justfile and  toolchain file will now be managed via renovate.

Note the justfile dependencies are now all locked


jsuereth:
This looks good, but you need to regenerate cargo dist workflow with `dist init` after `cargo install cargo-dist`.

Agreed we should update the justfile, I actually don't use it, but at least we should document somewhere this behavior.

PR #1206, opened 2026-02-15
Include resource attributes for live-check otlp emitted findings

### LINKED ISSUE #1152: Allow extracting resource attributes from the source signal
When weaver is run in `live-check` mode with OTLP log export in a CI or similar environment with multiple apps pointing their OTLP streams, it's impossible to determine which app a finding was generated for.

It will be useful to have the ability to extract the resource of from the source signal. This will allow users to extract things like `service.name`, `deployment.environment.name`, `k8s.pod.name` and so on, so they could easily determine and associate a finding to their applications.

jerbly:
Closes #1152 

Live-check with `--emit-otlp-logs` will now include the attributes from the resource in the emitted log record, this helps to identify the source of the finding in a multi-source environment.

- Spans, metrics, and logs now carry a reference to their parent resource, making resource attributes accessible during live checking.
- All advisors now include resource attributes in their findings output when emitting log records.
- The OTLP ingester now extracts and attaches resource data when converting incoming signals to internal sample types.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -428,9 +428,9 @@ checksum = "1fd0f2584146f6f2ef48085050886acf353beff7305ebd1ae69500e27c67f64b"
 
 [[package]]
 name = "bytes"
-version = "1.11.0"
+version = "1.11.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b35204fbdc0b3f4446b89fc1ac2cf84a8a68971995d0bf2e925ec7cd960f9cb3"
+checksum = "1e748733b7cbc798e1434b6ac524f0c1ff2ab456fe201501e6497c8417a4fc33"
 
 [[package]]
 name = "bytesize"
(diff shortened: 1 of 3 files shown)


crates/weaver_live_check/src/otlp_logger.rs
```diff
@@ -173,6 +201,23 @@ fn flatten_finding_context(context: &JsonValue) -> Vec<LogAttribute> {
...
+        JsonValue::String(s) => Some(AnyValue::from(s.clone())),
+        JsonValue::Number(n) => {
+            if let Some(i) = n.as_i64() {
+                Some(AnyValue::from(i))
+            } else {
+                n.as_f64().map(AnyValue::from)
+            }
+        }
+        JsonValue::Bool(b) => Some(AnyValue::from(*b)),
+        JsonValue::Null | JsonValue::Object(_) | JsonValue::Array(_) => None,
```
lquerel: Should we also ignore arrays of primitive JSON values?
I'm asking because in `conversion.rs` method `maybe_to_json` arrays are preserved.

lquerel (approved):
LGTM.

In its current form, I have the impression that this PR does not allow reporting resource attributes that are arrays of primitive values. I'm thinking in particular of attributes such as `host.ip`, `host.mac`, or even `process.command_args`, which are arrays of strings and could quite easily appear in the list of resource attributes. This could be added in a separate PR or in this one, up to you.

crates/weaver_live_check/src/otlp_logger.rs
jerbly: Added

PR #1225, opened 2026-02-19
ci: Add in a custom manager for the action snippet

thompson-tomo:
This builds upon #1205 by adding a custom manager to update the github actions in build-setup.yml. This will enable files to be kept in sync.

Closes #1223


jsuereth:
Is this just updating all the files by hand or are we runnign `dist init`? 

I'm ok if the former works, just wanted to confirm - as I think that's what you're doing here.

jsuereth:
I'd prefer to just use `cargo dist install && dist init` to keep 'em in sync, but let's try this and see how it works for now.

PR #1232, opened 2026-02-22
Live check dog-fooding: doc, code-gen and live-check

jerbly:
This PR makes live-check "eat its own dog food" — the OTLP log events emitted by `--emit-otlp-logs` are now defined as a semantic convention model. Weaver is used to generate docs and code and to run live-check against tests that emit logs.

- **Model** (`crates/weaver_live_check/model/live_check.yaml`): Defines the `weaver.finding` event, its attributes (`finding_id`, `signal_type`, `sample_type`, etc.), and enumerations as a v2 semconv registry.
- **Templates** (`crates/weaver_live_check/templates/`): Jinja templates generate Rust structs and enums with `strum` derives and a Markdown reference doc — all derived generically from the model, no hardcoded names.
- **Generated code** (`crates/weaver_live_check/src/generated/`): `attributes.rs`, `events.rs`, `entities.rs` — type-safe enums for `FindingId`, `SignalType`, `SampleType`, etc., replacing hand-written string constants.
- **Generated docs** (`crates/weaver_live_check/docs/finding.md`): Auto-generated reference documentation for the finding event schema.
- **Guide** (`crates/weaver_live_check/docs/dog-fooding.md`): Explains the model, how to regenerate, and the design.
- **Integration test** (`crates/weaver_live_check/tests/livecheck_emit.rs`): End-to-end test that emits diverse findings via `OtlpEmitter` into a live-check instance using the live_check model as its registry, then asserts zero violations and 100% registry coverage — confirming the emitted OTLP log records conform to the model.
- **CI freshness check** (`cargo xtask check-generated`): Cross-platform xtask that regenerates code/docs and fails if there's any drift from what's committed. Runs on ubuntu in the test job.


crates/weaver_live_check/model/live_check.yaml
```diff
@@ -1,137 +1,295 @@
...
-        brief: "Signal name (e.g., event name or metric name)"
-        requirement_level: recommended
-        stability: development
-        examples: ["http.request.duration", "user.login.event"]
-
-  - id: weaver.live_check.finding
-    type: event
-    name: weaver.live_check.finding
+# yaml-language-server: $schema=../../../schemas/semconv-group.json
+file_format: definition/2
```
jsuereth, line 2: WOOT! v2!

I love the idea of dogfooding here, just looking through all the CI/CD changes.

One question - should we have per-crate weaver, or "global project" weaver?

jerbly: I'm not sure what you mean exactly. Feels like the code gen is for this crate so the def needs to be here?

crates/weaver_live_check/model/live_check.yaml
```diff
@@ -1,137 +1,295 @@
...
+      - `weaver.finding.context.expected`: `"http.request.method"`
+
+      The context structure varies by finding type but commonly includes details like
+      attribute names, expected values, stability levels, and deprecation reasons.
+    stability: development
+    examples:
+      - "http.method"
+      - "http.request.method"
+
+  - key: weaver.finding.sample_type
```
lmolkova:
```suggestion
  - key: weaver.finding.sample.type
```

`type` is a property of `sample`, so they should be separated by `.`. `live_check` is a good example of `_`, but when in doubt, use `.`.

(now that we're in semconv territory, let's bikeshed a little :) )

crates/weaver_live_check/model/live_check.yaml
```diff
@@ -1,137 +1,295 @@
...
+          brief: "A log record or event"
+          stability: development
+    brief: "The type of telemetry sample being validated"
+    note: |
+      This indicates what kind of OpenTelemetry data structure was analyzed.
+      The sample type helps categorize findings and understand which part of
+      the telemetry pipeline generated the issue.
+    stability: development
+
+  - key: weaver.finding.signal_type
```
lmolkova:
```suggestion
  - key: weaver.finding.signal.type
```

crates/weaver_live_check/model/live_check.yaml
```diff
@@ -1,137 +1,295 @@
...
+          brief: "Log record or event signal"
+          stability: development
+    brief: "The OpenTelemetry signal type that the finding applies to"
+    note: |
+      The signal type categorizes the finding according to the OpenTelemetry data model.
+      This helps correlate findings with specific parts of an observability system
+      (traces, metrics, or logs).
+    stability: development
+
+  - key: weaver.finding.signal_name
```
lmolkova:
```suggestion
  - key: weaver.finding.signal.name
```

crates/weaver_live_check/model/live_check.yaml
```diff
@@ -1,137 +1,295 @@
...
+    note: |
+      For metrics, this is the metric name (e.g., `http.server.request.duration`).
+      For log events, this is the event name (e.g., `exception`).
+      For spans, this is the span name (e.g., `GET /api/users`).
+      This field helps identify exactly which signal in the telemetry has the issue.
+    stability: development
+    examples:
+      ["http.server.request.duration", "exception", "GET /api/users"]
+
+  - key: weaver.finding.resource_attribute
```
lmolkova:
```suggestion
  - key: weaver.finding.resource.attribute
```

crates/weaver_live_check/model/live_check.yaml
```diff
@@ -1,137 +1,295 @@
...
+  - key: weaver.finding.context
+    type: template[any]
+    brief: "Additional contextual information about the finding"
+    note: |
+      Context attributes provide structured details about what was found during validation.
+      Each context field is flattened into a separate attribute using dot notation.
+      For example, a context containing `{"attribute_name": "http.method", "expected": "http.request.method"}`
+      would be recorded as two attributes:
+
+      - `weaver.finding.context.attribute_name`: `"http.method"`
```
lmolkova: we should probably rename it everywhere (attribute name will became v1 legacy, it's key on the proto level and in v2)
```suggestion
      - `weaver.finding.context.attribute_key`: `"http.method"`
```

crates/weaver_live_check/model/live_check.yaml
```diff
@@ -1,137 +1,295 @@
...
+            hard errors. Examples include using recommended attributes, upgrading from
+            development to stable conventions, or following best practices.
+          stability: development
+        - id: information
+          value: "information"
+          brief: "Useful context about the telemetry without requiring action"
+          note: >
+            Information findings provide helpful context for understanding the telemetry
+            or the validation results. For example, noting that a template attribute
+            was used.
```
lmolkova, line 123: Not blocking: I still don't understand why is it important to know that template attribute was used. It's the same as 'string' attribute was used or value of this attribute was 5.

crates/weaver_live_check/model/manifest.yaml
```diff
@@ -0,0 +1,6 @@
+name: weaver-live-check
+description: Weaver Live Check Semantic Conventions
+schema_url: https://github.com/open-telemetry/weaver/crates/weaver_live_check/model/v0.1.0
```
lmolkova:
```suggestion
schema_url: https://github.com/open-telemetry/weaver/crates/weaver_live_check/model/0.1.0
```

https://github.com/open-telemetry/opentelemetry-specification/blob/27094331af23ba915168c54cb503f13669464b4c/specification/schemas/README.md#schema-url

> http[s]://server[:port]/path/<version>
> Version number follows the MAJOR.MINOR.PATCH format, similar to semver 2.0.

lmolkova (commented):
this is so cool!!!

crates/weaver_live_check/model/live_check.yaml
jerbly: Not quite the same as `string` etc. It's useful to catch unintended use of a defined template-type. e.g. `http.request.header.my.mistake` is valid but might be unintentional.

lmolkova: but there is no difference between `http.request.header.all-good` and `http.request.header.my.mistake` from schema / semantic-conventions. 

My argument was that `template[string]` is a type in definition, `string` is also a type in a definition.  
We see `string` attribute over the wire and it matches `template[string]` definition type.
So we fire info advise when we see that attribute conforms to the conventions - is that useful?

I'm trying to create some mental model on when to fire which level and in my mind "everything is according to spec" is either not fired at all and only appears in metrics or has some trace/debug level, but consistently.

jerbly: I understand, I feel that template-types fall into an awkward category where you're providing keys and values. And therefore there's a different class of mistake that can be made. This is why I set this at `information` level.

Here's another example. Let's say there's an error in my code. I wanted to use: `http.request.method_original` but I accidentally wrote `http.request.header.method_original`. These are both valid according to the definition.

I realize these examples are somewhat tenuous - so if we think there's no value having this `information` level finding we can remove it. Should we also look at other `information` level findings? `opt_in_attribute_not_present`, `conditionally_required_attribute_not_present`, `undefined_enum_variant`, `extends_namespace`

#1256 allows you to exclude all `information` level findings or specific findings if you so wish.

crates/weaver_live_check/model/live_check.yaml
jsuereth: I'm asking if we should have a crate for codegen, or run weaver so it generates *all* code across all crates or if we plan to run it per-create, i.e separately.

jerbly: I guess that depends on whether we think other crates will produce telemetry in the future. As it stands today, only `emit` and `live-check` produce telemetry. `emit` spoofs telemetry from other models so it's not really a candidate for codegen.

Would it be useful to have other commands, like `check` produce Findings as OTLP?

schemas/semconv-group.json
```diff
@@ -0,0 +1,2563 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "title": "SemConvSpec",
```
lmolkova: do we need it published? I'd rather have v1 and v2 independently :)

lmolkova (approved):

schemas/semconv-group.json
jerbly: Removed and switched to v2 directly. I did this because that got around the file_format missing in the schema. But I see you have a PR to fix that now.

crates/weaver_live_check/model/live_check.yaml
```diff
@@ -1,137 +1,295 @@
...
+    type: template[any]
+    brief: "Resource attributes from the original telemetry source that produced the finding"
+    note: |
+      When a finding is emitted as an OTLP log record, the resource attributes from the
+      original telemetry signal are included with this prefix. This allows identifying
+      which service or application produced the telemetry that triggered the finding,
+      which is essential in CI environments with multiple instrumented applications.
+
+      For example, if the original telemetry had resource attribute `service.name: "my-app"`,
+      the finding log record would include `weaver.finding.resource_attribute.service.name: "my-app"`.
```
lmolkova:
```suggestion
      the finding log record would include `weaver.finding.resource.attribute.service.name: "my-app"`.
```

lmolkova (approved):

crates/weaver_live_check/model/live_check.yaml
jsuereth: I think what's here is fine for now, we can ask the question when we need a second component with telemetry.

jsuereth (approved):

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ All notable changes to this project will be documented in this file.
...
 - Unified output handling across all registry subcommands. Builtin output formats (json, yaml, jsonl) are now available in `registry stats`, `registry diff`, and `registry resolve`. `registry stats` also supports `--templates` for custom text output templates. ([#1200](https://github.com/open-telemetry/weaver/pull/1200) by @jerbly)
 - New feature ([#1152](https://github.com/open-telemetry/weaver/issues/1152)) - Live-check with `--emit-otlp-logs` will now include the attributes from the resource in the emitted log record, this helps to identify the source of the finding in a multi-source environment. ([#1206](https://github.com/open-telemetry/weaver/pull/1206) by @jerbly)
+- Live-check OTLP log findings are now dog-fooded: the event schema, attributes, and enumerations are defined in a semconv model and code-generated using Weaver's own templates. See [`finding.md`](crates/weaver_live_check/docs/finding.md) for the generated reference documentation and [`dog-fooding.md`](crates/weaver_live_check/docs/dog-fooding.md) for the full dog-fooding guide. ([#1232](https://github.com/open-telemetry/weaver/pull/1232) by @jerbly)
```
lmolkova: should we call out changes that might affect current users? 

- attribute_name → attribute_key in context
- weaver.finding.sample_type → weaver.finding.sample.type                                                                                                                                                                        
- weaver.finding.signal_type → weaver.finding.signal.type                                                                                                                                                                        
- weaver.finding.signal_name → weaver.finding.signal.name
- weaver.finding.resource_attribute.<key> → weaver.finding.resource.attribute.<key>     

?

jerbly: Yes, we must! I'll get this branch up to date and fix that this evening.

PR #1239, opened 2026-02-24
Fix #760, Disables MiniJinja auto-escaping on filename

### LINKED ISSUE #760: Weaver registry diff: template extension weirdness
Take a look at the https://github.com/open-telemetry/semantic-conventions/pull/2307

it has

```
schema-diff/
  ansi/
    schema-diff.j2
    weaver.yaml
```

I run 

```ps
weaver registry diff -r  https://github.com/open-telemetry/semantic-conventions/archive/refs/tags/v1.34.0.zip[model] \
  --baseline-registry https://github.com/open-telemetry/semantic-conventions/archive/refs/tags/v1.33.0.zip[model] \
  --diff-template .\internal\tools\scripts\schema-diff\ --diff-format ansi -o diff-output
```

and things work fine, I get something like
```yaml
    all:
      changes:
        - rename_attributes:
            attribute_map:
              code.column: code.column.number
              code.filepath: code.file.path
              ...
```
in output file.

Now, if I rename `schema-diff.j2` -> `schema-diff.yaml.j2` (and also change file name in the config), I'll get

```yaml
""
    all:
      changes:
        - rename_attributes:
            attribute_map:
              "code.column": "code.column.number"
              "code.filepath": "code.file.path"
              ...
```
(Note all the quotes)

It also reproduces with `--diff-format yaml` and corresponding folder structure.
It seems there is some escaping(?) magic that tries to be helpful based on the file extension, but it gets in the way.

jerbly:
Fixes #760 

💥 BREAKING CHANGE 💥

Auto-escaping is now off by default (`none`) for all templates, regardless of file extension. To opt in, set `auto_escape: html` or `auto_escape: json` per template in `weaver.yaml`. Within a template, `{% autoescape false %}` blocks can selectively disable escaping for sections. Use `|tojson` for explicit JSON/YAML value escaping where needed.

See: https://docs.rs/minijinja/latest/minijinja/fn.default_auto_escape_callback.html

Also fixed the order of precedence for template and format usage:
1. filesystem template
2. embedded template
3. builtin

This allows you to have a `yaml` target even though `yaml` is now a builtin.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -57,3 +57,6 @@ history-temp-repo/
 ui/node_modules/
 ui/dist/
 ui/.vite/
+
+# claude
+.claude/


crates/weaver_forge/src/lib.rs
```diff
@@ -712,6 +712,23 @@ impl TemplateEngine {
...
+        // Use `|tojson` for explicit JSON/YAML value escaping in templates.
+        // See: https://docs.rs/minijinja/latest/minijinja/fn.default_auto_escape_callback.html
+        env.set_auto_escape_callback(|name| {
+            let name = name
+                .strip_suffix(".j2")
+                .or_else(|| name.strip_suffix(".jinja"))
+                .or_else(|| name.strip_suffix(".jinja2"))
+                .unwrap_or(name);
+            match name.rsplit('.').next().unwrap_or("") {
+                "html" | "htm" | "xml" => minijinja::AutoEscape::Html,
```
jsuereth: Should this include `markdown`, `java`, or not?

I'm wondering if we should make this configurable...

jerbly: From the docs:
> The default logic for auto escaping based on file extension.
> 
> [Html](https://docs.rs/minijinja/latest/minijinja/enum.AutoEscape.html#variant.Html): .html, .htm, .xml
> [Json](https://docs.rs/minijinja/latest/minijinja/enum.AutoEscape.html#variant.Json): .json, .json5, .js, .yaml, .yml
> [None](https://docs.rs/minijinja/latest/minijinja/enum.AutoEscape.html#variant.None): all others

It seems highly unlikely that you want escaping if you're building a yaml file in the context of weaver. But, yeah we could prevent the breaking change and make this a config option instead if you think that's better?

jerbly: @jsuereth - I made this per template configurable in the weaver.yaml. This now has the added benefit that, regardless of file extension, you can enable auto escaping.

crates/weaver_forge/src/config.rs
```diff
@@ -143,6 +143,32 @@ pub enum ApplicationMode {
...
+/// Auto-escape mode for a template's output.
+/// `none` (default): no escaping.
+/// `html`: escape `<`, `>`, `&`, `"`, `'`, `/` for HTML/XML.
+/// `json`: serialize values as JSON (suitable for JSON/YAML embedding).
+///
+/// This mirrors `minijinja::AutoEscape` but adds serde support for
+/// deserialization from `weaver.yaml`.
+#[derive(Serialize, Deserialize, Debug, Clone, Default, PartialEq)]
+#[serde(rename_all = "snake_case")]
+pub(crate) enum AutoEscapeMode {
```
jsuereth, line 155: WOOT! Thank you

PR #1240, opened 2026-02-24
Fix 404 on template type attributes

jerbly:
Fixed a bug in the serve api. Attempts to get a template-type attribute would return a 404.

### DIFF
--- a/src/serve/handlers.rs
+++ b/src/serve/handlers.rs
@@ -124,8 +124,12 @@ pub async fn get_registry_attribute(
     // Remove leading slash if present (from wildcard match)
     let key = key.trim_start_matches('/');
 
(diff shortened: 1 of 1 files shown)


src/serve/handlers.rs
```diff
@@ -124,8 +124,12 @@ pub async fn get_registry_attribute(
     // Remove leading slash if present (from wildcard match)
     let key = key.trim_start_matches('/');
 
-    // O(1) lookup via SearchContext
-    match state.search_ctx.get_attribute(key) {
+    // O(1) lookup via SearchContext, checking both regular and template attributes
+    match state
```
jsuereth, line 128: Would be nice if we had tests for the API in the future.

jsuereth (approved):

PR #1250, opened 2026-02-26
Initial cut at supporting refinements in v2 syntax.

jsuereth:
Note: I used gemini-cli to aide in the writing of this PR.

- Updates `refinement` to actually be a refinement.  This means metric names, notes, etc. will be copied from the refined group vs. needing to be restated.
- Creates a `is_v2` flag to track v2 refinements through V1 so resolution engine will work.
- Update dependency + resolution to support `extends` additionally having more things that can be pulled in.

This does NOT do the following:

- Update lineage any further.  That's a BIG todo I plan to address in a different monster PR.
- Allow complicated "don't inherit" logic for attributes or annotations.  You can append-only and override.  I expect us to figure out more complicated refinement in non-breaking ways going forward.

### DIFF
--- a/schemas/semconv.schema.v2.json
+++ b/schemas/semconv.schema.v2.json
@@ -528,21 +528,21 @@
       "type": "object",
       "properties": {
         "annotations": {
-          "description": "Additional annotations for the signal.\nNote: This field is currently not propagated during resolution.",
+          "description": "Additional annotations for the signal.",
           "type": "object",
           "additionalProperties": {
             "$ref": "#/$defs/YamlValue"
           }
         },
         "brief": {
-          "description": "Refines the brief description of the signal.\nNote: This field is currently not propagated during resolution.",
+          "description": "Refines the brief description of the signal.",
           "type": [
             "string",
             "null"
           ]
         },
         "deprecated": {
-          "description": "Specifies if the signal is deprecated.\nNote: This field is currently not propagated during resolution.",
+          "description": "Specifies if the signal is deprecated.",
           "anyOf": [
             {
               "$ref": "#/$defs/Deprecated"
@@ -553,7 +553,7 @@
           ]
         },
         "description": {
-          "description": "The additional attributes to describe of the Entity.",
+          "description": "The additional attributes to describe the Entity.",
           "type": "array",
           "items": {
             "$ref": "#/$defs/AttributeRef"
@@ -564,7 +564,7 @@
           "$ref": "#/$defs/SignalId"
         },
         "note": {
-          "description": "Refines the more elaborate description of the signal.\nNote: This field is currently not propagated during resolution.",
+          "description": "Refines the more elaborate description of the signal.",
           "type": [
             "string",
             "null"
@@ -575,7 +575,7 @@
           "$ref": "#/$defs/SignalId"
         },
         "stability": {
-          "description": "Refines the stability of the signal.\nNote: This field is currently not propagated during resolution.",
+          "description": "Refines the stability of the signal.",
           "anyOf": [
             {
               "$ref": "#/$defs/Stability"
@@ -722,7 +722,7 @@
       "type": "object",
       "properties": {
         "annotations": {
-          "description": "Additional annotations for the signal.\nNote: This field is currently not propagated during resolution.",
+          "description": "Additional annotations for the signal.",
           "type": "object",
           "additionalProperties": {
             "$ref": "#/$defs/YamlValue"
@@ -736,14 +736,14 @@
           }
         },
         "brief": {
-          "description": "Refines the brief description of the signal.\nNote: This field is currently not propagated during resolution.",
+          "description": "Refines the brief description of the signal.",
           "type": [
             "string",
             "null"
           ]
         },
         "deprecated": {
-          "description": "Specifies if the signal is deprecated.\nNote: This field is currently not propagated during resolution.",
+          "description": "Specifies if the signal is deprecated.",
           "anyOf": [
             {
               "$ref": "#/$defs/Deprecated"
@@ -754,7 +754,7 @@
           ]
         },
         "entity_associations": {
-          "description": "Which resources this event should be associated with.\nNote: This field is currently not propagated during resolution.",
+          "description": "Which resources this event should be associated with.",
           "type": "array",
           "items": {
             "type": "string"
@@ -765,7 +765,7 @@
           "$ref": "#/$defs/SignalId"
         },
         "note": {
-          "description": "Refines the more elaborate description of the signal.\nNote: This field is currently not propagated during resolution.",
+          "description": "Refines the more elaborate description of the signal.",
           "type": [
             "string",
             "null"
@@ -776,7 +776,7 @@
           "$ref": "#/$defs/SignalId"
         },
         "stability": {
-          "description": "Refines the stability of the signal.\nNote: This field is currently not propagated during resolution.",
+          "description": "Refines the stability of the signal.",
           "anyOf": [
             {
               "$ref": "#/$defs/Stability"
@@ -1054,7 +1054,7 @@
       "type": "object",
       "properties": {
         "annotations": {
-          "description": "Additional annotations for the signal.\nNote: This field is currently not propagated during resolution.",
+          "description": "Additional annotations for the signal.",
           "type": "object",
           "additionalProperties": {
             "$ref": "#/$defs/YamlValue"
@@ -1068,14 +1068,14 @@
           }
         },
         "brief": {
-          "description": "Refines the brief description of the signal.\nNote: This field is currently not propagated during resolution.",
+          "description": "Refines the brief description of the signal.",
           "type": [
             "string",
             "null"
           ]
         },
         "deprecated": {
-          "description": "Specifies if the signal is deprecated.\nNote: This field is currently not propagated during resolution.",
+          "description": "Specifies if the signal is deprecated.",
           "anyOf": [
             {
               "$ref": "#/$defs/Deprecated"
@@ -1086,7 +1086,7 @@
           ]
         },
         "entity_associations": {
-          "description": "Which resources this metric should be associated with.\nNote: This field is currently not propagated during resolution.",
+          "description": "Which resources this metric should be associated with.",
           "type": "array",
           "items": {
             "type": "string"
@@ -1097,7 +1097,7 @@
           "$ref": "#/$defs/SignalId"
         },
         "note": {
-          "description": "Refines the more elaborate description of the signal.\nNote: This field is currently not propagated during resolution.",
+          "description": "Refines the more elaborate description of the signal.",
           "type": [
             "string",
             "null"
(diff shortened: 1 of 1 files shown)


crates/weaver_resolver/data/registry-test-v2-4-refinements/registry/v2-refinements.yaml
```diff
@@ -0,0 +1,65 @@
+version: '2'
```
lmolkova:
```suggestion
file_format: definition/2
```

crates/weaver_resolver/data/registry-test-v2-4-refinements/registry/v2-refinements.yaml
```diff
@@ -0,0 +1,65 @@
...
+    kind: client
+    name:
+      note: "base span"
+    brief: Base span
+    stability: stable
+
+span_refinements:
+  - id: span.refined.span
+    ref: base.span
+    kind: server
```
lmolkova: mmm, I think we should not allow to change kind when refining.

crates/weaver_resolver/data/registry-test-v2-4-refinements/registry/v2-refinements.yaml
```diff
@@ -0,0 +1,65 @@
...
+    brief: Base metric
+    stability: stable
+    attributes:
+      - ref_group: test.group
+
+metric_refinements:
+  - id: metric.refined.metric
+    ref: base.metric
+    brief: Refined metric
+    unit: ms
```
lmolkova: we shouldn't be able to update units, right?

crates/weaver_semconv/src/v2/entity.rs
```diff
@@ -27,6 +32,36 @@ pub struct Entity {
...
 
+/// A refinement of an existing entity.
+#[derive(Serialize, Deserialize, Debug, Clone, JsonSchema)]
+#[serde(deny_unknown_fields)]
+pub struct EntityRefinement {
+    /// The ID of the refinement.
+    pub id: SignalId,
+    /// The name of the entity being refined.
+    pub r#ref: SignalId,
+    /// The additionaly attributes to describe of the Entity.
```
lmolkova: I can see real @jsuereth here!

crates/weaver_resolver/data/registry-test-v2-4-refinements/expected-registry.json
```diff
@@ -0,0 +1,219 @@
...
+      "type": "entity",
+      "brief": "Base entity",
+      "stability": "stable",
+      "attributes": [
+        1
+      ],
+      "name": "base.entity",
+      "lineage": {
+        "provenance": {
+          "registry_id": "default",
```
lmolkova, line 15: unrelated, I need to clean this one up

crates/weaver_resolver/data/registry-test-v2-4-refinements/registry/v2-refinements.yaml
jsuereth: That's the major question on this PR. I just vibed out everything and I think we pare back.

crates/weaver_resolver/data/registry-test-v2-4-refinements/registry/v2-refinements.yaml
jsuereth: Nominally, yes - it's hugely problematic.  I was literally just formulating a discussion on this.

E.g. Should you be able to refine span name, or just adapt the note describing it?

crates/weaver_semconv/src/v2/entity.rs
jsuereth: Ha! sadly that's a copy-pasta by the AI of the real @jsuereth, if you can believe it.

crates/weaver_resolver/data/registry-test-v2-4-refinements/expected-registry.json
jsuereth: Yeah - I need to sort out provenance going forward.  I don't think we can declare resolved schema final until it's added, but I think we can add it in a non-breaking way.

crates/weaver_resolver/data/registry-test-v2-4-refinements/registry/v2-refinements.yaml
jsuereth: fixed

crates/weaver_resolver/data/registry-test-v2-4-refinements/registry/v2-refinements.yaml
jsuereth: Ok - I fixed this to not update units or instrument or metric name (which it was previously doing).

This implementation is now complete, resolving some of these.

crates/weaver_resolver/data/registry-test-v2-4-refinements/registry/v2-refinements.yaml
jsuereth: Ok - this is now fixed to be what I think we should support initially.

crates/weaver_resolved_schema/src/registry.rs
```diff
@@ -141,6 +141,15 @@ pub struct Group {
     #[serde(skip_serializing)]
     #[schemars(skip)]
     pub visibility: Option<AttributeGroupVisibilitySpec>,
+
+    /// Whether this group is a v2 group.
+    ///
+    /// This does NOT survive serialization, but is used when
+    /// tracking v2 groups during resolution.
+    #[serde(default)]
+    #[serde(skip_serializing)]
```
lmolkova, line 150:
```suggestion
    #[serde(skip)]
```

also skip deserializing?

crates/weaver_resolver/src/registry.rs
```diff
@@ -483,25 +505,75 @@ fn resolve_extends_references(ureg: &mut UnresolvedRegistry) -> Result<(), Error
...
+                        // Copy over fields refinements MUST use.
+                        unresolved_group.group.instrument = parent_summary.instrument.clone();
+                        unresolved_group.group.unit = parent_summary.unit.clone();
+                        unresolved_group.group.span_kind = parent_summary.span_kind;
+                        unresolved_group.group.metric_name = parent_summary.metric_name.clone();
+
+                        // Optionally copy over fields if refinements have not set them.
+                        if unresolved_group.group.stability.is_none() {
+                            unresolved_group.group.stability = parent_summary.stability.clone();
+                        }
```
lmolkova, line 551: this is where we'd validate that refinement can't be more stable than definition (if we decide to do it), right?

schemas/semconv.schema.v2.json
```diff
@@ -25,6 +25,20 @@
         "$ref": "#/$defs/Entity"
       }
     },
+    "entity_refinements": {
```
lmolkova, line 28: could you please ask AI to update https://github.com/open-telemetry/weaver/blob/main/schemas/semconv-syntax.v2.md too ?

crates/weaver_resolver/src/error.rs
```diff
@@ -153,6 +153,19 @@ pub enum Error {
         provenances: Vec<Provenance>,
     },
 
+    /// An invalid `extends` clause type.
+    #[error("Group `{group_id}` of type `{group_type}` cannot extend group `{extends_ref}` of type `{extends_type}`.")]
+    InvalidRefinement {
```
lmolkova, line 158: could we use v2 properties? we don't really support refinements on v1 anyway?

crates/weaver_resolver/src/error.rs
```diff
@@ -153,6 +153,19 @@ pub enum Error {
...
+        /// The id of the group with the issue.
+        group_id: String,
```
lmolkova:
```suggestion
        refinement_id: String,
```

crates/weaver_resolver/src/error.rs
```diff
@@ -153,6 +153,19 @@ pub enum Error {
...
+        /// The group id that is being extended.
+        extends_ref: String,
```
lmolkova:
```suggestion
        ref: String, // signal_ref ?
```

crates/weaver_resolver/src/error.rs
```diff
@@ -153,6 +153,19 @@ pub enum Error {
...
+        /// The type of the group.
+        group_type: String,
```
lmolkova:
```suggestion
        signal_type: String,
```

crates/weaver_resolver/src/error.rs
```diff
@@ -153,6 +153,19 @@ pub enum Error {
...
+        /// The type of the group being extended.
+        extends_type: String,
```
lmolkova:
```suggestion
        signal_type: String,
```

crates/weaver_resolved_schema/src/registry.rs
jsuereth: I was using deserializing for tests

crates/weaver_resolver/src/registry.rs
jsuereth: Yes!

schemas/semconv.schema.v2.json
jsuereth: :) Sure

crates/weaver_resolver/src/error.rs
jsuereth: Good point, I can update this.

schemas/semconv-syntax.v2.md
```diff
@@ -49,9 +49,13 @@ A semantic convention file starts with `file_format: definition/2` and may conta
 
 - `attributes`: Contains definitions of semantic attributes which may be applicable to all OpenTelemetry signals.
 - `entities`: Contains definitions of entities.
+- `entity_refinements`: Contains definitions of entity refinements.
```
lmolkova, line 52: there are also sections underneath to describe format in human-readable form, e.g. https://github.com/jsuereth/weaver/blob/6b909ea304d8d727a2a96a6b12dddb19b97e02de/schemas/semconv-syntax.v2.md#entities-definition

(cached and blessed AI output so that people don't need to AI that for us)

jsuereth: oh, I fixed that, need to re-push

jsuereth: huh, the commit is showing up in git locally, but not github - odd

schemas/semconv.schema.v2.json
```diff
@@ -495,6 +523,75 @@
         "stability"
       ]
     },
+    "EntityRefinement": {
+      "description": "A refinement of an existing entity.",
+      "type": "object",
+      "properties": {
+        "annotations": {
+          "description": "Additional annotations for the signal.\nNote: This field is currently not propagated during resolution.",
```
lquerel: Is it still true now that you added support to annotation merging?

lquerel (approved):
LGTM. 
nit: I believe the json schema need to be updated now that you support some merging.

schemas/semconv.schema.v2.json
jsuereth: I'll make sure this is regenerated.  Yes this is now fixed, which I wasn't planning to do at first.

PR #1254, opened 2026-02-28
Add weaver registry package command

### LINKED ISSUE #1244: Implement `weaver registry package` command

lmolkova:
Fixes #1244


crates/weaver_semconv/src/manifest.rs
```diff
@@ -475,3 +485,89 @@ registry_path: "./registry"
         );
     }
 }
+
+/// The file format version of the publication manifest.
+pub const PUBLICATION_MANIFEST_FILE_FORMAT: &str = "2.0.0";
```
lmolkova: Do we want it to be `2.0.0` or `manifest/2.0.0` ?

jerbly: To me, `file_format` expresses a type and a version `type/version`. So it should be `manifest/2.0.0`. Otherwise we would have to call it `version` or `file_version` in order to just provide `2.0.0`.

src/registry/package.rs
```diff
@@ -0,0 +1,283 @@
...
+    #[command(flatten)]
+    policy: PolicyArgs,
+
+    /// Parameters to specify the diagnostic format.
+    #[command(flatten)]
+    pub diagnostic: DiagnosticArgs,
+}
+
+fn write_yaml(path: &Path, data: &impl serde::Serialize) -> Result<(), DiagnosticMessages> {
+    let yaml = serde_yaml::to_string(data).map_err(|e| Error::InvalidParams {
```
jerbly: `InvalidParams` is not the right error - params are to do with template parameters used during the `generate` command. 

Maybe add something like:
```rust
#[error("Failed to write output file `{path}`: {error}")]
OutputWriteError { path: PathBuf, error: String },
```

Also, super-nit: you can avoid the temp string by using `serde_yaml::to_writer`.

src/registry/package.rs
```diff
@@ -0,0 +1,283 @@
...
+        params_file: path.to_path_buf(),
+        error: e.to_string(),
+    })?;
+    fs::write(path, yaml).map_err(|e| Error::InvalidParams {
```
jerbly: As above, `InvalidParams` is not the right error.

src/registry/package.rs
```diff
@@ -0,0 +1,283 @@
...
+    let resolved = weaver.resolve(loaded, &mut diag_msgs)?;
+
+    let resolved_v2: ResolvedV2 = resolved.try_into()?;
+    resolved_v2.check_after_resolution_policy(&mut diag_msgs)?;
+
+    if !diag_msgs.is_empty() {
+        return Err(diag_msgs);
+    }
+
+    fs::create_dir_all(&args.output).map_err(|e| Error::InvalidParams {
```
jerbly: As above, `InvalidParams` is not the right error.

src/registry/package.rs
lmolkova: thanks, done!

crates/weaver_semconv/src/manifest.rs
```diff
@@ -475,3 +485,121 @@ registry_path: "./registry"
...
+pub const PUBLICATION_MANIFEST_FILE_FORMAT: &str = "manifest/2.0.0";
+
+/// Represents the publication manifest for a packaged semantic convention registry.
+///
+/// This is produced by `weaver registry package` and describes the contents of
+/// a self-contained registry artifact, including its resolved schema location.
+#[derive(Serialize, Deserialize, Debug, Clone, JsonSchema)]
+pub struct PublicationRegistryManifest {
+    /// The file format version of this publication manifest.
+    /// Always `"2.0.0"`.
```
lmolkova:
```suggestion
    /// Always `"manifest/2.0.0"`.
```

docs/usage.md
```diff
@@ -53,7 +55,7 @@ Manage Semantic Convention Registry
 
 * `check` — Validates a semantic convention registry.
 * `generate` — Generates artifacts from a semantic convention registry.
-* `resolve` — DEPRECATED - Resolves a semantic convention registry. This command is deprecated and will be removed in a future version. Please use `generate` or `package` instead.
+* `resolve` — Resolves a semantic convention registry.
```
jerbly, line 58: Need to update the /// rust doc so DEPRECATED is generated here.

lmolkova: You're probably looking at specific commit, there is no change here if you look at all commits

crates/weaver_semconv/src/manifest.rs
```diff
@@ -475,3 +485,121 @@ registry_path: "./registry"
...
+
+/// The file format version of the publication manifest.
+pub const PUBLICATION_MANIFEST_FILE_FORMAT: &str = "manifest/2.0.0";
+
+/// Represents the publication manifest for a packaged semantic convention registry.
+///
+/// This is produced by `weaver registry package` and describes the contents of
+/// a self-contained registry artifact, including its resolved schema location.
+#[derive(Serialize, Deserialize, Debug, Clone, JsonSchema)]
+pub struct PublicationRegistryManifest {
```
lmolkova: let's try putting it into resolved schema crate

crates/weaver_semconv/src/manifest.rs
```diff
@@ -2,11 +2,10 @@
 
 //! Contains the definitions for the semantic conventions registry manifest.
 //!
-//! This struct is used to specify the registry, including its name, version,
-//! description, and few other details.
-//!
-//! In the future, this struct may be extended to include additional information
-//! such as the registry's owner, maintainers, and dependencies.
+//! Two manifest types are defined here:
+//! - [`RegistryManifest`]: flexible, can be a definition or publication manifest
```
lmolkova: let's try to do definition and publication manifests and variant for both

crates/weaver_semconv/src/manifest.rs
```diff
@@ -475,3 +485,121 @@ registry_path: "./registry"
...
+
+/// Represents the publication manifest for a packaged semantic convention registry.
+///
+/// This is produced by `weaver registry package` and describes the contents of
+/// a self-contained registry artifact, including its resolved schema location.
+#[derive(Serialize, Deserialize, Debug, Clone, JsonSchema)]
+pub struct PublicationRegistryManifest {
+    /// The file format version of this publication manifest.
+    /// Always `"manifest/2.0.0"`.
+    pub file_format: String,
```
lmolkova: let's put it into json-schema enum and use it for that command and also in json schema

src/registry/json_schema.rs
```diff
@@ -64,6 +66,9 @@ pub(crate) fn command(args: &RegistryJsonSchemaArgs) -> Result<ExitDirectives, D
         }
         JsonSchemaType::Diff => schema_for!(weaver_version::schema_changes::SchemaChanges),
         JsonSchemaType::DiffV2 => schema_for!(weaver_version::v2::SchemaChanges),
+        JsonSchemaType::PublicationManifest => {
```
lmolkova: let's also add definition manifest

src/registry/package.rs
```diff
@@ -0,0 +1,284 @@
...
+
+    // we require a manifest file to be present for packaging
+    let registry_manifest = main_registry_repo
+        .manifest()
+        .ok_or_else(|| Error::PackagingRequiresManifest {
+            registry: registry_path.to_string(),
+        })?
+        .clone();
+
+    let loaded = weaver.load_definitions(main_registry_repo, &mut diag_msgs)?;
```
jsuereth: I'm so happy this API is working out for us, when I see it used in commands like this.

Any friction / oddities in `weaver` here?  Should we perhaps have it also create `RegsitryRepo`?

crates/weaver_resolver/src/registry.rs
```diff
@@ -983,7 +983,7 @@ mod tests {
                 assert!(schema.is_err(), "This test is expected to fail");
                 let expected_errors: String = std::fs::read_to_string(&expected_errors_file)
                     .expect("Failed to read expected errors file");
-                let observed_errors = serde_json::to_string(&schema).unwrap();
+                let observed_errors = serde_json::to_string(&schema.err()).unwrap();
```
lmolkova, line 1058: schema does not need to be serializeable, it was the only place that required it to be, but we don't care about schema itself, only about errors.

schemas/definition-manifest.v2.json
```diff
@@ -0,0 +1,113 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "title": "DefinitionRegistryManifest",
+  "description": "Represents the definition manifest for a semantic convention registry.\n\nThis is used when developing a registry before it is published.\nSee [`PublicationRegistryManifest`] for the stricter publication form produced\nby `weaver registry package`.",
+  "type": "object",
+  "properties": {
```
lmolkova, line 6: no file format here. not sure we need one for definition, let's add it when necessary.

crates/weaver_semconv/src/manifest.rs
lmolkova: I tried, essentially it means have duplicate structure, one unnamed here and another one named there and some copy logic. but in practice it does not seem to bring any benefit.

I wonder if separation between semconv and resolved crates is even necessary? you can't use semconv without resolving them.

src/registry/package.rs
lmolkova: yep, it's super-easy to use, no friction at all! 

I think we should do a big cleanup when we remove legacy and use this chance to polish this - repo seems like an impl detail that would be cool to hide. 
I don't think it's a big deal unless we want to ship libraries and need good public api

crates/weaver_semconv/src/manifest.rs
lmolkova: created https://github.com/open-telemetry/weaver/issues/1266 to follow up

Issue #1248, opened 2026-02-26
Live-check: allow filtering findings when emitting logs over otlp


lmolkova:
<img width="2006" height="1460" alt="Image" src="https://github.com/user-attachments/assets/11b5709f-43af-474e-83b7-6439ae0cd1ed" />

Most of these findings are known, expected or something app developer has no control over (e.g. `exception.escapted` attribute is set by otel python for back-compat reasons).

I thought `advice-preprocessor` could help me, but it seems it only massages the data given to a policy and it does not seem I can use it to drop/filter things.

What would be great is some config (similar to linting suppressions) like
```yaml
min_finding_level: WARN
suppress_findings:
  - ids: ["not_stable"]
  # some future advanced capabilities to suppress more granularly, but it might not be worth it
```

It can also be cmd line params (min finding level, finding ids to suppress) for the time being.

jerbly:
Fixed in: https://github.com/open-telemetry/weaver/releases/tag/v0.23.0

PR #1256, opened 2026-03-01
Finding filters via Phase 1 of .weaver.toml config file

### LINKED ISSUE #1247: Stop emitting advice when enum has unknown value
Enums are expected to have unknown values, if some of them need strict validation, it can come as custom policy but it can't be default applied to all enums.

E.g. `error.type` is one of the most common semconv attributes and it's [designed](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/registry/attributes/error.md#error-type) to be an open set.

Adding weaver live check to a typical app with HTTP or DB calls brings A LOT of these

Image

PS: I love the ability to emit logs to my backend!

jerbly:
Closes #1247, #1248.

Adds `.weaver.toml` project-level config, completing **Phase 1** of the [package-vs-project config proposal](docs/specs/package-vs-project-config/config-proposal.md). Discovery walks up from CWD; `--config <path>` overrides. Precedence: defaults -> `.weaver.toml` -> CLI flags.

## Phase 1 (live-check)

- `[live_check]` covers every `live-check` CLI flag, including `[live_check.otlp]` and `[live_check.emit]` sub-tables.
- `[[live_check.finding_filters]]` drops findings by id, sample name, level, or signal type.

## Phase 2 PoC

Partial shared sections (`[registry]`, `[policy]`, `[diagnostics]`) included so live-check is end-to-end usable. Wired through a temporary heuristic and marked with a TODO — the proper rework converts the shared `*Args` structs to `Option<T>` and adopts the standard `CliOverrides` pattern uniformly.

## Infrastructure

- **`weaver_config` crate** split into `live_check`, `registry`, and `overrides` modules.
- **`CliOverrides` trait + `override_if_set!` macro**: generic IoC pattern subcommands can adopt with one impl.
- **Automatic CLI/config consistency test**: schemars + clap introspection catches drift between config fields and CLI args. Generic helper — future commands opt in with one line. 
- **JSON schema** at `schemas/weaver-config.json`. Use the `#:schema` directive at the top of `.weaver.toml` for editor autocomplete. Regenerate via `weaver registry json-schema --json-schema weaver-config`.

See the **Finding Modification** section in the live-check README for usage details.


crates/weaver_config/src/lib.rs
```diff
@@ -0,0 +1,290 @@
...
+//! Discovery walks up from the current working directory to find the first
+//! `.weaver.toml` file. The `--config` CLI option can override this.
+
+use schemars::JsonSchema;
+use serde::Deserialize;
+use std::path::{Path, PathBuf};
+use weaver_checker::FindingLevel;
+
+/// The filename to search for during discovery.
+const CONFIG_FILENAME: &str = ".weaver.toml";
```
jsuereth, line 22: Are we moving to `toml` from `yaml`?

-> insert meme of factory worker seeing a new thing on the assembly line.

I'm personally fine either way, but given our model is in YAML, what's the justification for a different config file format?

crates/weaver_config/src/lib.rs
```diff
@@ -0,0 +1,290 @@
...
+    /// Optional signal type scope. When set, this filter only applies to
+    /// findings with a matching signal_type.
+    pub signal_type: Option<String>,
+}
+
+/// Discover a `.weaver.toml` file by walking up from the given directory.
+///
+/// Returns the path to the first `.weaver.toml` found, or `None` if none exists.
+#[must_use]
+pub fn discover(start: &Path) -> Option<PathBuf> {
```
jsuereth, line 42: Don't we already have a lot of code that does this for WeaverConfig today?

crates/weaver_config/README.md
```diff
@@ -0,0 +1,9 @@
+# Weaver Config
+
+Project-level configuration for Weaver via `.weaver.toml`.
```
jsuereth, line 3: How does this line up with - https://github.com/open-telemetry/weaver/blob/main/crates/weaver_forge/src/config.rs#L45 ?

I know in your comment you said you'd figure this out later, but I'm not sure how this will work.  I'd love to have a better design/idea of how to do this migration before we start down the path, otherwise evaluating this PR is really hard, since a lot of the features are duplicative, just in a different spot.

Would it be easier to *first* extract WeaverConfig into its own location, then add features live-check needs?

jsuereth (commented):
I love the direction, but soft-blocking.  

- I think we need to understand what `WeaverConfig` of today is meant to do vs. this.
- We should figure out if `toml` vs. `yaml` is ok
- It's possible we want a formal split between `WeaverPackageConfig` and `WeaverRuntimeConfig`.  I can buy that, but let's not use confusing names between crates here.

jerbly:
TBH, I've never read the comment at the top of the weaver_forge config.rs - I totally didn't realise it was meant to be used for weaver configuration in general! My bad!

> * I think we need to understand what `WeaverConfig` of today is meant to do vs. this.

Yes! I would like to understand that. I've always seen the `weaver.yaml` files _belonging_ to the jinja template processes, not belonging to the instance of weaver you're running. That separation (until now) has been in command-line options - the way I see it.

> * We should figure out if `toml` vs. `yaml` is ok

We can do `yaml` - I was just following the crowd for linters etc. If we're somehow combining this with `weaver.yaml` then, of course we should stay on `yaml`.

> * It's possible we want a formal split between `WeaverPackageConfig` and `WeaverRuntimeConfig`.  I can buy that, but let's not use confusing names between crates here.

This is much better than my confusing reuse of the name, and follows my thoughts on "what the config belongs to". We would have to clarify both. But, now I've read the weaver_forge config.rs in more detail I'm not sure what we should be doing. 😕

crates/weaver_config/src/overrides.rs
```diff
@@ -0,0 +1,90 @@
...
+
+/// Trait for CLI args structs that can load and override a config section.
+///
+/// Implement this on each command's `*Args` struct to enable the generic
+/// `load_config()` flow: defaults → `.weaver.toml` → CLI overrides.
+///
+/// The field-mapping methods (`config_only_fields`, `cli_only_args`,
+/// `field_mappings`) enable a generic consistency test that verifies every
+/// config field has a CLI arg and vice versa, with no per-command test code.
+pub trait CliOverrides {
```
jsuereth, line 34: Should we have `CliOverrides<T: "Config">` here?

E.g. do we envision overriding *different* config types via the same CLI argument?

crates/weaver_config/src/registry.rs
```diff
@@ -0,0 +1,123 @@
...
+use std::path::PathBuf;
+
+use schemars::JsonSchema;
+use serde::Deserialize;
+
+/// Registry configuration — where to load the semantic convention registry from.
+#[derive(Debug, Clone, Default, Deserialize, PartialEq, JsonSchema)]
+#[serde(default)]
+#[schemars(inline)]
+pub struct RegistryConfig {
```
jsuereth, line 17: Note: This is a *non-blocking* comment, but something that's been noodling in my head for a while.

If we moved `live-check` to a mode where it *dynamically* downloaded `schema_url` on-demand based on what it sees in telemetry - what would this config look like?

Would this instead be some kind of a `schema-url -> config` map?  Curious on your thoughts - I was toying around with trying to get AI to make this change and it started floundering with complexity, so I think this is one we'd need to talk through our architecture for.

crates/weaver_config/src/overrides.rs
```diff
@@ -0,0 +1,90 @@
...
+    }
+}
+
+/// Apply a single CLI override: if `src` is `Some`, write it to `dst`.
+///
+/// Two forms:
+/// - `override_if_set!(dst, src)` — for `Option<T> → T` (clone into existing value)
+/// - `override_if_set!(dst, src, optional)` — for `Option<T> → Option<T>` (wrap in `Some`)
+#[macro_export]
+macro_rules! override_if_set {
```
jsuereth, line 79: There's a *lot* of this in the attribute/signal inheritance code - I wonder if we should move this to common eventually to share?

jsuereth (approved):
Approving this as a good first start - *caveat* I do think the `weaver.yoml` syntax is somethign I'm worried about.  I think we should issue a warning that the schema is not stable if folks use it for now, until we're happy with how it handles things.

You can see my comments on things I'm worried about breaking in the future.

PR #1260, opened 2026-03-04
Package vs Project proposal

jerbly:
Proposal for clean separation of configuration that belongs with an artifact that weaver uses, templates, versus configuration for weaver itself e.g. live-check settings in a project.


docs/specs/package-vs-project-config/config-proposal.md
```diff
@@ -0,0 +1,129 @@
+# Proposal: Package Config vs Project Config
+
+## Background
+
+Weaver already has configuration files that sit alongside the artifacts they describe. A `registry_manifest.yaml` lives at the root of a semantic convention registry and declares its name, version, and dependencies. A `weaver.yaml` lives at the root of a template package and defines how those templates are processed — syntax delimiters, comment formats, text maps, default params, and application modes.
```
lmolkova: nit, it's now just `manifest`

```suggestion
Weaver already has configuration files that sit alongside the artifacts they describe. A `manifest.yaml` lives at the root of a semantic convention registry and declares its name, version, and dependencies. A `weaver.yaml` lives at the root of a template package and defines how those templates are processed — syntax delimiters, comment formats, text maps, default params, and application modes.
```

docs/specs/package-vs-project-config/config-proposal.md
```diff
@@ -0,0 +1,129 @@
...
+
+This pattern — config adjacent to artifacts — works well because it keeps configuration close to the thing it configures. The config belongs to the package: it's authored, versioned, and distributed together with the artifacts.
+
+## Problem
+
+Today `weaver.yaml` is the only configuration mechanism, and it's being asked to serve two roles. Beyond configuring the template package it sits in, it also supports directory-walking and a `$HOME/.weaver/weaver.yaml` fallback — features designed to let users customize behavior from outside the package. This conflation creates confusion:
```
lmolkova, line 11: > it also supports directory-walking and a `$HOME/.weaver/weaver.yaml`

TIL!

docs/specs/package-vs-project-config/config-proposal.md
```diff
@@ -0,0 +1,129 @@
...
+A new `.weaver.toml` file belongs to the **project** — the codebase that uses Weaver. It configures everything that does not belong to a template package: settings that are CLI-configured today, plus new project-level settings that are cumbersome to provide on the command line.
+
+**What it configures:**
+
+- Live-check finding overrides and filters
+- Live-check OTLP and emit settings
+- Default registry path, template path, policies path
+- Any other setting that is currently a CLI flag
+
+**Where it lives:** In the project repository. Discovered by walking up from the current working directory. A `--config` flag provides an explicit override.
```
lmolkova: `weaver registry generate` currently takes `--config` for template config. should we just stop supporting it there since "`template_config.yaml` is read from the given template directory and nowhere else"

Maybe we can figure out how to do it without breaking? E.g. with --future flag or when --v2 is used

docs/specs/package-vs-project-config/config-proposal.md
```diff
@@ -0,0 +1,129 @@
...
+# target = ""
+#
+# [emit]
+# endpoint = "http://localhost:4317"
+```
+
+## Why Two Files / Two Formats?
+
+- **YAML for packages** — template config is complex (nested structures, comment format definitions, glob patterns). YAML is already the established format for this.
+- **TOML for projects** — project config is flatter and simpler. TOML is the standard for project-level config in Rust ecosystems, and the dotfile convention (`.weaver.toml`) signals "project config, check this in."
```
lmolkova, line 151: I actually like a different format, it's also something only power-users would ever interact with

docs/specs/package-vs-project-config/config-proposal.md
```diff
@@ -0,0 +1,129 @@
...
+
+```toml
+# === Phase 1: Live-check finding modification ===
+
+[[live_check.finding_overrides]]
+id = ["not_stable"]
+level = "information"
+signal_type = "span"
+
+[[live_check.finding_filters]]
```
lmolkova, line 74: orthogonal to this proposal, but I don't understand the different between finding_overrides and finding_filters

docs/specs/package-vs-project-config/config-proposal.md
```diff
@@ -0,0 +1,129 @@
...
+exclude = ["deprecated"]
+min_level = "improvement"
+exclude_samples = ["trace.parent_id", "trace.span_id"]
+
+# === Phase 2: CLI-equivalent settings ===
+
+# Shared options (apply to all subcommands that accept them)
+# [registry]
+# path = "https://github.com/open-telemetry/semantic-conventions.git"
```
lmolkova, line 107: I think it's common to break down params into two groups: persistent ones and specific invocation.
Only persistent ones go into the config. 

E.g. clippy does not seem to allow to configure which path to lint, it only takes it in CLI.

Following this logic:
- path and include_unreferenced seem to be pre-invocation. 
- Follow_symlink is probably persistent
- policies could be persistent (like I want all my semconv to pass policies)
- skipping policies is per-invocation
....

docs/specs/package-vs-project-config/config-proposal.md
```diff
@@ -0,0 +1,129 @@
...
+The name `weaver.yaml` is too generic — it sounds like it configures Weaver itself, when it actually configures a template package. Renaming to `template_config.yaml` makes its purpose explicit: it is the configuration for the templates it sits alongside.
+
+`weaver.yaml` will continue to be loaded as a fallback during a deprecation period, with a warning directing package authors to rename to `template_config.yaml`.
+
+### 2. Load template config from the package directory only
+
+`template_config.yaml` is read from the given template directory and nowhere else. No directory-walking. No `$HOME/.weaver/` fallback. This is already the behavior for remote template packages and the update-markdown command — we are making local packages consistent with remote ones.
+
+A template package should be self-contained and predictable. The same package produces the same output regardless of what `weaver.yaml` files happen to exist in parent directories or the user's home directory.
+
```
lquerel, line 29: It's important to note that some users use the same `acronyms`, the same `whitespace_control`, and the same `template_syntax` across multiple template packages (e.g. one package for code gen language A and language B, or even another for doc gen using those same parameters). The current mechanism makes that easy, and it would be a shame to lose that. At the same time, I understand and agree with the rest of the argument.  

I suggest that the new `.weaver.toml` file should allow defining (or overriding) a limited set of safe, shared knobs (e.g. `acronyms`, `template_syntax`, `whitespace_control`, `params`). If we do that, then we can make the simplification described in this proposal.

docs/specs/package-vs-project-config/config-proposal.md
```diff
@@ -0,0 +1,129 @@
...
+### 3. Introduce `.weaver.toml` as project config
+
+A new `.weaver.toml` file belongs to the **project** — the codebase that uses Weaver. It configures everything that does not belong to a template package: settings that are CLI-configured today, plus new project-level settings that are cumbersome to provide on the command line.
```
lquerel: A **project** can be a simple repo or a monorepo made up of multiple projects/services, which reinforces the need for "walking up from the current working directory" as defined later. Maybe we should clarify a bit more what we mean by "project" to better explain the value of the mechanism that's been put in place?

docs/specs/package-vs-project-config/config-proposal.md
```diff
@@ -0,0 +1,129 @@
...
+# otlp_logs_endpoint = "http://localhost:4317"
+# otlp_logs_stdout = false
+#
+# [generate]
+# templates = "templates"
+# target = ""
+#
+# [emit]
+# endpoint = "http://localhost:4317"
+```
```
lquerel, line 146: As a follow up to my previous comments, we could have something like:

```toml
# Shared defaults applied to all template packages
[template_defaults]
acronyms = ["API", "HTTP", "SDK", "CLI", "URL", "JSON"]

[template_defaults.template_syntax]
block_start = "{%"
block_end = "%}"
variable_start = "{{"
variable_end = "}}"
comment_start = "{#"
comment_end = "#}"

[template_defaults.whitespace_control]
trim_blocks = true
lstrip_blocks = true
keep_trailing_newline = false

[template_defaults.params]
copyright_owner = "Acme"
include_experimental = false
year = 2026
```

and even an optional per-package overrides (if we want, could come later).

```toml
# Optional per-package overrides
[[template_overrides]]
acronyms = ["API", "HTTP", "SDK", "CLI", "OTLP"]

[template_overrides.template_syntax]
variable_start = "[["
variable_end = "]]"

[template_overrides.whitespace_control]
trim_blocks = false

[template_overrides.params]
language = "rust"
crate_name = "otel_semconv"

# TOML has no null so we could use explicit unsets
unset_params = ["include_experimental"]
```

docs/specs/package-vs-project-config/config-proposal.md
lquerel: I agree as well.

docs/specs/package-vs-project-config/config-proposal.md
```diff
@@ -0,0 +1,129 @@
...
+
+[[live_check.finding_filters]]
+exclude = ["deprecated"]
+min_level = "improvement"
+exclude_samples = ["trace.parent_id", "trace.span_id"]
+
+# === Phase 2: CLI-equivalent settings ===
+
+# Shared options (apply to all subcommands that accept them)
+# [registry]
```
lquerel, line 106: Should we rename this section `[defaults.registry]` ?

docs/specs/package-vs-project-config/config-proposal.md
```diff
@@ -0,0 +1,129 @@
...
+# path = "https://github.com/open-telemetry/semantic-conventions.git"
+# follow_symlinks = false
+# include_unreferenced = false
+# v2 = true
+#
+# [policy]
```
lquerel, line 112: Should we rename this section `[defaults.policy]` ?

docs/specs/package-vs-project-config/config-proposal.md
```diff
@@ -0,0 +1,129 @@
...
+# paths = ["./policies"]
+# skip = false
+#
+# [diagnostics]
```
lquerel, line 116: Should we rename this section `[defaults.diagnostics]` ?

docs/specs/package-vs-project-config/config-proposal.md
```diff
@@ -0,0 +1,129 @@
...
+# otlp_logs = false
+# otlp_logs_endpoint = "http://localhost:4317"
+# otlp_logs_stdout = false
+#
+# [generate]
+# templates = "templates"
+# target = ""
+#
+# [emit]
+# endpoint = "http://localhost:4317"
```
lquerel, line 145: Should we rename those sections with the following pattern `[commands.<command_in_cli>]` ? 
That will make things really clear in my opinion.

lquerel (commented):
I really like the direction this proposal is taking. I've added a number of suggestions that I think remain aligned with the overall idea.

docs/specs/package-vs-project-config/config-proposal.md
jerbly: That's not all settled yet... but, overrides allow you to change the level of a finding, e.g. if you want strict stability you could upgrade that finding to a violation level. Filters allow you to completely remove a finding. I could probably combine these (I did have an iteration with a single `finding_modifier` that had an `action` field in to which you could define exclude or a level change. It was starting to get complicated so I moved the "action" out and made the two configs.

docs/specs/package-vs-project-config/config-proposal.md
jerbly: Why restrict it? if my `.weaver.toml` is in a project dir I don't want to keep typing `weaver registry check --registry /path/to/my/model` - I can just type `weaver registry check` and override the default registry which is the semconv github project main branch.

docs/specs/package-vs-project-config/config-proposal.md
jerbly: This makes sense to me. You're actively choosing to change the behaviour of the packaged config within your project.

docs/specs/package-vs-project-config/config-proposal.md
jerbly: Yes, that makes sense. I also really want to take the opportunity to find more `defaults` where the same settings are used across two or more commands. e.g. `grpc*` is now also used by `infer`

docs/specs/package-vs-project-config/config-proposal.md
```diff
@@ -0,0 +1,129 @@
...
+## Problem
+
+Today `weaver.yaml` is the only configuration mechanism, and it's being asked to serve two roles. Beyond configuring the template package it sits in, it also supports directory-walking and a `$HOME/.weaver/weaver.yaml` fallback — features designed to let users customize behavior from outside the package. This conflation creates confusion:
+
+- A `weaver.yaml` in a parent directory or `$HOME` silently merges into every template package loaded below it. It's unclear where a setting originates.
+- New features like live-check finding modification don't belong in template package config, but there's nowhere else to put project-level concerns.
+
+## Proposal
+
+### 1. Rename `weaver.yaml` to `template_config.yaml`
```
jsuereth: Bikeshed: `weaver_template.yaml` ?

docs/specs/package-vs-project-config/config-proposal.md
```diff
@@ -0,0 +1,129 @@
...
+
+## Why Two Files / Two Formats?
+
+- **YAML for packages** — template config is complex (nested structures, comment format definitions, glob patterns). YAML is already the established format for this.
+- **TOML for projects** — project config is flatter and simpler. TOML is the standard for project-level config in Rust ecosystems, and the dotfile convention (`.weaver.toml`) signals "project config, check this in."
+- Two distinct files with distinct formats makes it impossible to confuse one for the other.
+
+## Implementation Phases
+
+**Phase 1:** Introduce `.weaver.toml` with `[live_check]` finding overrides and filters.
```
jsuereth, line 156: Can we add it for `check` and `live-check`? The stable-by-default / federated semconv OTEP would be WAY simpler if `check` also had config we could share.

docs/specs/package-vs-project-config/config-proposal.md
lmolkova: do we need overrides? I might be out-of-date or never pain enough attention, but I've never used level overrides in linters. If you want custom, disable default and write your own policy?

docs/specs/package-vs-project-config/config-proposal.md
lmolkova: back to what we chatted on the call: it's not typical and is surprising. 

If `weaver registry check` is provided without args, I imagine it defaults `-r` to `.`; if it's also in the config, `-r`  resolution becomes ambiguous - can you use config for a specific registry on a different registry provided in explicit option? is it valid? 

----
I think we're discussing this: is this config for a registry or for a weaver command.

I'm going to throw a crazy idea and run away.

What if it's a part of definition manifest? 
- My registry files are in this location
- I want this set of policies I want to run on it (this is almost independent of the tooling used to validate)
- I want this set of live-checks (? maybe, not 100% sure if it's part of definition)

Then weaver.toml becomes a list of setting specific to weaver, not specific to registry.
You pass  `-r [definition-]manifest.yaml` and this is where registry config lives.

docs/specs/package-vs-project-config/config-proposal.md
jerbly: We can talk more about that in #1256 - I may have just made up a use case where there isn't one.

docs/specs/package-vs-project-config/config-proposal.md
jerbly: > If weaver registry check is provided without args, I imagine it defaults -r to .

It defaults to semconv github main branch today.

Like cargo, I don't want to change dir to `target/debug` every time I need to build, test or run my debug code.

Maybe we establish that `model` is the default location under the project root to have your model and then you can change that (like you can with `.cargo/config.toml` for targets) to some other location. 

So, when you run a weaver command that uses a registry it would:

1. use what you specify in `-r`
2. look in CWD for a `manifest.yaml`
3. look in `model-dir` location as specified in `.weaver.toml` with `./model` as the default
4. error

docs/specs/package-vs-project-config/config-proposal.md
jerbly: phase 1 is just these new filtering settings for `live-check`. phase 2 will include `check`. We can prioritise `check`

docs/specs/package-vs-project-config/config-proposal.md
jerbly: Agreed, and updated the doc.

docs/specs/package-vs-project-config/config-proposal.md
```diff
@@ -0,0 +1,174 @@
...
+
+A template package should be self-contained and predictable. The same package produces the same output regardless of what `weaver.yaml` files happen to exist in parent directories or the user's home directory.
+
+The `--config` / `-c` flag on `generate` (which currently accepts additional `weaver.yaml` files for layering) is deprecated as part of this change. It was the explicit version of the same layering mechanism — with `.weaver.toml` taking over project-level overrides, it is no longer needed.
+
+### 3. Introduce `.weaver.toml` as project config
+
+A new `.weaver.toml` file belongs to the **project** — the codebase that uses Weaver. A project may be a single repository or a monorepo containing multiple services. It configures everything that does not belong to a template package: settings that are CLI-configured today, plus new project-level settings that are cumbersome to provide on the command line.
+
+**Discovery strategy — open question:** `.weaver.toml` is discovered by walking up from CWD through parent directories to root and then includes `$HOME/.weaver.toml`. Two approaches:
```
jerbly, line 36: @jsuereth @lquerel @lmolkova - I have this open question if you can give me your opinions here.

lmolkova: Is this needed right away or ever? I believe this was not used before and we can always build it if it becomes necessary.

docs/specs/package-vs-project-config/config-proposal.md
lmolkova: > It defaults to semconv github main branch today.

😮 another TIL 
Does it make sense on all commands though? Live check makes sense, but why everyone should generate, resolve, package, check, update markdown for otel semconv main branch?! 


I don't think we need to decide on this PR, but I think there will be properties that belong on the config and don't belong on it. Some properties are part of the manifest, some are in config and maybe there are some that are cli only. I don't think it changes general design here

lmolkova (approved):
this is great! I'd like to polish the details around properties (which config / manifest they go to), but that can be done during implementation

docs/specs/package-vs-project-config/config-proposal.md
lmolkova: update: even live-check does not make sense because of main branch - only semconv internals and prototypes should care about it.

docs/specs/package-vs-project-config/config-proposal.md
lquerel: I clearly prefer the second approach for the reasons mentioned in a previous comment (monorepo or multiple templates sharing configuration elements). Example in a monorepo context:
- I define acronyms at the global level (across microservices).
- I define parameters at the service level, for example to specify their names.

lmolkova: Is this a practical or a theoretical need? In my mental model one config per repo/registry is necessary and merging live check suppression rules across registries automatically does not make sense. 
If we need some inheritance and walking up, i would always prefer it to be explicit and only add it when 2-3 people ask for it

PR #1262, opened 2026-03-04
Add file_format field to JSON Schema v2

lmolkova:
JSON Schema v2 does not contain file_format resulting in benign but annoying VS code errors

image

This PR brings it there without breaking compatibility with `version: "2"` and other tricks we do in custom deserialization.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -64,7 +64,7 @@ jobs:
         # we specify bash to get pipefail; it guards against the `curl` command
         # failing. otherwise `sh` won't catch that `curl` returned non-0
         shell: bash
-        run: "curl --proto '=https' --tlsv1.2 -LsSf https://github.com/axodotdev/cargo-dist/releases/download/v0.30.4/cargo-dist-installer.sh | sh"
+        run: "curl --proto '=https' --tlsv1.2 -LsSf https://github.com/axodotdev/cargo-dist/releases/download/v0.31.0/cargo-dist-installer.sh | sh"
       - name: Cache dist
         uses: actions/upload-artifact@b7c566a772e6b6bfb58ed0dc250532a479d7789f
         with:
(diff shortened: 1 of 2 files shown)


crates/weaver_semconv/src/v2/mod.rs
```diff
@@ -50,10 +50,16 @@ pub struct CommonFields {
...
-/// A semconv file is a collection of semantic convention groups (i.e. [`GroupSpec`]).
+/// A semconv file is a collection of attributes, signals, groups,
+/// and imports.
 #[derive(Serialize, Deserialize, Debug, Clone, JsonSchema)]
 #[serde(deny_unknown_fields)]
 pub struct SemConvSpecV2 {
+    /// The file format version.
+    /// Must be "definition/2" for this version of the schema.
+    #[serde(default, skip_serializing_if = "String::is_empty")]
+    #[schemars(extend("const" = "definition/2"))]
```
lmolkova: try to do it with transforming schema itself, not the code

src/registry/json_schema.rs
```diff
@@ -59,7 +59,7 @@ pub(crate) fn command(args: &RegistryJsonSchemaArgs) -> Result<ExitDirectives, D
     let json_schema = match args.json_schema {
         JsonSchemaType::ResolvedRegistry => schema_for!(ResolvedRegistry),
         JsonSchemaType::SemconvGroup => schema_for!(Versioned),
```
jsuereth: Should we do this same trick for V1?

lmolkova: V1 is perfectly fine without file format - I don't want anyone to start using it, so I'd rather keep it intact

jsuereth: Right - but look at it returning `Versioned` here - which means it's both V1 and V2 in the same schema - This is the complaint you have on my PR which blindly updated the schema :)

I think we should probably just changed `SemconvGroup` to be `schema_for!(SemconvGroupV1)` or whatever here... then I can merge in your pr before submitting mine.

PR #1263, opened 2026-03-04
Move infer logic to its own crate

ArthurSens:
It's pretty much just changing code from one file to another, nothing else

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -123,7 +123,7 @@ version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
 dependencies = [
- "windows-sys 0.61.2",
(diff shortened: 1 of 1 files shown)


crates/weaver_infer/README.md
```diff
@@ -0,0 +1,77 @@
+# Weaver Infer
```
ArthurSens, line 1: Ok, after opening the PR, I realized we have a check requiring README files for all crates.

I've added a README with the mindset that crates are libraries and aimed for developers. This is definitely not aimed at CLI users. Am I in the right mindset?

jsuereth: Yes.

This, e.g. lets us re-use functionality in the weaver MCP server, or via the API for our UI.

jerbly:
@ArthurSens - can you add some more tests to get the coverage up? 56 missing lines at the moment. (we're trying to keep coverage high as it helps both humans and AIs with feedback).

PR #1267, opened 2026-03-05
Support imports for all signal types and public attribute groups.

jsuereth:
Expand import support so we can cover all possible scenarios.

### DIFF
--- a/ui/pnpm-lock.yaml
+++ b/ui/pnpm-lock.yaml
@@ -16,10 +16,10 @@ importers:
         version: 1.162.8(@tanstack/react-router@1.162.8(react-dom@19.2.4(react@19.2.4))(react@19.2.4))(@tanstack/router-core@1.162.6)(csstype@3.2.3)(react-dom@19.2.4(react@19.2.4))(react@19.2.4)
       dompurify:
         specifier: ^3.3.1
-        version: 3.3.1
+        version: 3.3.2
       marked:
         specifier: ^17.0.1
-        version: 17.0.3
+        version: 17.0.4
       rapidoc:
         specifier: ^9.3.8
         version: 9.3.8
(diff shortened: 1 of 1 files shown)


schemas/semconv.schema.json
lmolkova, line 1: is this intentional or necessary? we probably have a wrong constant for it  (Versioned instead of v1) and the output is not right

schemas/semconv-syntax.md
```diff
@@ -46,10 +46,12 @@ groups ::= semconv [imports]
 
 semconv ::= id convtype brief [note] [extends] [stability] [deprecated] [display_name] [attributes]  [annotations] specificfields
 
-imports := [metrics] [events] [entities]
+imports := [metrics] [events] [entities] [spans] [attribute_groups]
```
lmolkova, line 49: I assume at this point it's not possible to not make change to v1?

crates/weaver_semconv/src/semconv.rs
```diff
@@ -538,6 +550,28 @@ mod tests {
                 .len(),
             1
         );
+        assert_eq!(
```
lmolkova, line 553: do we have some resolve/forge tests for imports? can we add spans and attribute groups to them?

schemas/semconv.schema.json
```diff
@@ -1,847 +1,2869 @@
...
+            {
+              "type": "null"
+            }
+          ]
+        }
+      },
+      "additionalProperties": false,
+      "required": [
+        "file_format"
+      ]
```
lquerel: Was `file_format` mandatory in v1? I thought it wasn’t.

schemas/semconv.schema.json
jsuereth: Asked this in slack - This is what following the directions of using `weaver registry json-schema` did because it was outputing the *versioned* instead of v1.  Should I revert this and wait for your PR to fix that?

schemas/semconv-syntax.md
jsuereth: It's shared, so this SHOULD work for both.  I just didn't update the docs.

crates/weaver_semconv/src/semconv.rs
jsuereth: I have resolve unit tests, I can make an integration test for it though, too.

schemas/semconv.schema.json
jsuereth: NO - I think this is due to a change in what `weaver registry json-schema` outputs - so you can't copy-paste here anymore.  See the thread above with @lmolkova

lmolkova: you can for schema v1 - we don't validate it :)

schemas/semconv.schema.json
lmolkova: let's not update v1 schema here - nothing will complain because v1 is not validated (there are more discrepancies there than versioned 😨 ) or if you really want to update it in this PR, let's change json-schema output to v1

jsuereth: Reverted!

schemas/semconv.schema.json
jsuereth: Reverted the whole thing

crates/weaver_semconv/src/semconv.rs
jsuereth: Ok, added more integration tests.

I haven't tested forge specifically, but you can see the resolved registry has the imports in it, and so Forge should be working here.

crates/weaver_resolver/data/registry-test-published-2/registry/main.yaml
```diff
@@ -0,0 +1,32 @@
+version: '2'
```
lmolkova, line 1:
```suggestion
file_format: definition/2
```

PR #1270, opened 2026-03-08
Add json schema for policy findings, make context optional

lmolkova:
noticed that context is required on PolicyFinding while working on https://github.com/open-telemetry/opentelemetry-weaver-packages/pull/24, but it does not need to be, so fixing that and also adding json schema for it

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -64,7 +64,7 @@ jobs:
         # we specify bash to get pipefail; it guards against the `curl` command
         # failing. otherwise `sh` won't catch that `curl` returned non-0
         shell: bash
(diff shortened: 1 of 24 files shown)


crates/weaver_live_check/src/advice/mod.rs
```diff
@@ -40,7 +40,7 @@ pub trait Advisor {
 /// Fluent builder for creating PolicyFinding instances with automatic emission
 pub struct FindingBuilder {
     id: String,
-    context: JsonValue,
+    context: Option<JsonValue>,
```
jsuereth, line 43: Ah - I noticed this but forgot to make a note/issue to fix later.  Thanks for fixing this!

PR #1271, opened 2026-03-09
Fix package command not producing output due to warnings

lmolkova:
and add test

### DIFF
--- a/crates/weaver_resolver/src/lib.rs
+++ b/crates/weaver_resolver/src/lib.rs
@@ -194,6 +194,10 @@ mod tests {
                         assert!(group.is_some());
                         let group = resolved_registry.group("event.session.end");
                         assert!(group.is_some());
+                        let group = resolved_registry.group("custom.group");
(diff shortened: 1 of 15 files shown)


crates/weaver_resolver/src/lib.rs
```diff
@@ -124,6 +124,7 @@ impl SchemaResolver {
             include_unreferenced,
         )
         .map(move |resolved_registry| {
+            let root_attributes = attr_catalog.drain_root_attributes();
             let catalog = Catalog::from_attributes(attr_catalog.drain_attributes());
```
jsuereth, line 127: I think if you call `drain_attributes` twice, this should be broken and lead to empty catalogs.

Maybe that's fine on this path for some reason? but it reads oddly, as this line theoretically isn't safe to call.

lmolkova: I pushed wrong branch, this is not related

PR #1277, opened 2026-03-10
Initial implementatoin of V2 lineage.

jsuereth:
An initial implementation of lineage for V2.

- Tracks whether refinements override or re-use brief/note/etc. from the signal.
- Tracks what aspects of attributes of overriden/changed.
- Updates `weaver_resolver` test suite to allow testing resolved schema
- Updates resolved schema for V2 to use `BTree*` data structures for ordering.

## TODOs

- [ ] Figure out multi-registry provenance - today it uses the name.  We want the ability to know if something was imported so we can resolve conflicts when we move to multi-import allowance.
- [ ] Don't track overrides that MUST happen in signals (e.g. `requirement_level`, `sampling_relevant`) unless they are actually overriden.

### DIFF
--- a/crates/weaver_resolved_schema/src/v2/lineage.rs
+++ b/crates/weaver_resolved_schema/src/v2/lineage.rs
@@ -46,6 +46,7 @@ pub struct AttributeLineage {
 
 impl AttributeLineage {
     /// Creates a new attribute lineage.
+    #[must_use]
     pub fn new(source: AttributeSource) -> Self {
         Self {
             source,
@@ -75,6 +76,7 @@ pub struct SignalLineage {
 
 impl SignalLineage {
     /// Creates a new SignalLineage with the given provenance.
+    #[must_use]
     pub fn new(provenance: Provenance) -> Self {
         Self {
             provenance,
@@ -127,6 +129,7 @@ pub struct RefinementLineage {
 
 impl RefinementLineage {
     /// Creates a new RefinementLineage.
+    #[must_use]
     pub fn new(provenance: Provenance, refines: String) -> Self {
         Self {
             provenance,
--- a/crates/weaver_resolved_schema/src/v2/mod.rs
+++ b/crates/weaver_resolved_schema/src/v2/mod.rs
@@ -1,6 +1,6 @@
 //! Version 2 of semantic convention schema.
 
-use std::collections::{HashMap, HashSet};
+use std::collections::HashMap;
 
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
--- a/crates/weaver_resolver/src/attribute.rs
+++ b/crates/weaver_resolver/src/attribute.rs
@@ -326,7 +326,7 @@ impl AttributeLookup for V1Schema {
                 })
                 .map(|a| AttributeWithGroupId {
                     attribute: a,
-                    group_id: g.id.to_owned(),
+                    group_id: g.id.clone(),
                 })
         });
         result
--- a/crates/weaver_resolver/src/loader.rs
+++ b/crates/weaver_resolver/src/loader.rs
@@ -13,7 +13,7 @@ use walkdir::DirEntry;
 use weaver_common::result::WResult;
 use weaver_resolved_schema::v2::ResolvedTelemetrySchema as V2Schema;
 use weaver_resolved_schema::ResolvedTelemetrySchema as V1Schema;
-use weaver_semconv::registry_repo::{RegistryRepo, LEGACY_REGISTRY_MANIFEST, REGISTRY_MANIFEST};
+use weaver_semconv::registry_repo::{RegistryRepo, REGISTRY_MANIFEST};
 use weaver_semconv::{group::ImportsWithProvenance, semconv::SemConvSpecWithProvenance};
 
 use crate::Error;
@@ -295,7 +295,7 @@ fn load_definition_repository(
             && (extension == "yaml" || extension == "yml")
             && file_name != "schema-next.yaml"
             && file_name != REGISTRY_MANIFEST
-            && file_name != LEGACY_REGISTRY_MANIFEST
+            && file_name != "registry.yaml"
     }
     let local_path = registry_repo.path().to_path_buf();
     let registry_path_repr = registry_repo.registry_path_repr();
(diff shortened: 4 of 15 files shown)


crates/weaver_resolver/data/registry-test-lineage-v2-1-base-signals/expected-schema.json
```diff
@@ -0,0 +1,288 @@
...
+    "attribute_groups": [
+      {
+        "id": "test.attributes",
+        "attributes": [
+          0,
+          1
+        ],
+        "brief": "A test attribute group",
+        "stability": "stable",
+        "lineage": {
```
lmolkova, line 35: do we want lineage in resolve schema? it's going to really bloat it and it does not seem to be essential for telemetry consumers. TBH I struggle to understand why is needed in any cases but debugging. I think it makes sense to provide it in the forge one.
Even better: it seems to belong in the debug log when resolving a registry.

The dependency resolution is a separate story, but it would not need that much.

crates/weaver_resolver/data/registry-test-lineage-v2-3-multi-registry/registry/app.yaml
```diff
@@ -0,0 +1,19 @@
...
+      - ref: custom.attr1
+        requirement_level: required
+
+span_refinements:
+  - id: span.test.base.app.refinement
+    ref: test.base
+    brief: 'A test span refinement.'
+    note: 'This note overrides the base span imported from source.'
+    attributes:
+      - ref: custom.attr2
```
lmolkova, line 18: why does it work? there are no dependencies in the manifest

crates/weaver_resolver/data/registry-test-lineage-v2-3-multi-registry/expected-schema.json
```diff
@@ -0,0 +1,207 @@
...
+            "brief",
+            "note"
+          ],
+          "inherited_fields": [
+            "stability"
+          ],
+          "attributes": {
+            "custom.attr2": {
+              "source": {
+                "raw_attribute": "span.test.base"
```
lmolkova, line 142: `span.test.base` is not even included in this resolved schema file (why?), but also the most important information is which registry this attr came from and there is no indication of it.

crates/weaver_resolver/data/registry-test-lineage-v2-3-multi-registry/expected-schema.json
```diff
@@ -0,0 +1,207 @@
...
+      "brief": "Attribute 2",
+      "stability": "stable"
+    }
+  ],
+  "registry": {
+    "attributes": [],
+    "attribute_groups": [],
+    "spans": [
+      {
+        "type": "test.base.app.refinement",
```
lmolkova, line 29: why refinement is in the span type?

crates/weaver_resolver/data/registry-test-lineage-v2-1-base-signals/expected-schema.json
```diff
@@ -0,0 +1,288 @@
...
+                "brief",
+                "examples",
+                "note",
+                "requirement_level",
+                "stability"
+              ]
+            },
+            "custom.attr2": {
+              "source": {
+                "attribute_group": "registry.data.registry-test-lineage-v2-1-base-signals.registry.base"
```
lmolkova, line 54: shouldn't it be `raw_attribute` ? it seems it's coming through v1 and becomes irrelevant to the test case (definition v2)

crates/weaver_resolver/data/registry-test-lineage-v2-3-multi-registry/expected-schema.json
```diff
@@ -0,0 +1,207 @@
...
+        "stability": "stable",
+        "lineage": {
+          "provenance": {
+            "registry_id": "opentelemetry.io/schemas",
+            "path": "data/registry-test-lineage-v2-3-multi-registry/registry/app.yaml"
+          },
+          "attributes": {
+            "custom.attr2": {
+              "source": {
+                "raw_attribute": "span.test.base"
```
lmolkova, line 51: why `raw_attribute` ? I don't quite understand what it means. If it's

> /// The attribute is defined as a free-floating attribute not tied to any group.

then why does it have `span.test.base` as the source?

crates/weaver_resolver/data/registry-test-lineage-v2-2-refinements/expected-schema.json
```diff
@@ -0,0 +1,200 @@
...
+        ],
+        "brief": "A base metric",
+        "stability": "stable",
+        "lineage": {
+          "provenance": {
+            "registry_id": "default",
+            "path": "data/registry-test-lineage-v2-2-refinements/registry/refinements.yaml"
+          },
+          "attributes": {
+            "custom.attr1": {
```
lmolkova, line 51: why it's an attribute key and not an index in attr catalog?
also why attrs in catalog have no lineage - they are definitions and refinements and if we need lineage, it should be on the attr  in the catalog and then there will be less duplication on signals

crates/weaver_resolver/data/registry-test-lineage-v2-1-base-signals/expected-schema.json
jsuereth: I think this is a crucial dependency for us handling multi-imports in the future.  For V1 - lineage was *crucial* for making sure we could transition to v2 and we would not be where we are today without it.

So I do disagree here. I understand the concern, and it's possible we need to go with a far more lightweight lineage tracking design here, but as i was working on allowing multiple "parent" registries to be defined, lineage was necessary to resolve ambiguities and import-issues across registires.

crates/weaver_resolver/data/registry-test-lineage-v2-2-refinements/expected-schema.json
jsuereth: Right - I did the most naive implementation for now - This is kind of what I wanted to investigate.

I'm still considering the actual lineage tracking design, and the important pieces of it.

For the *most* part, I need to know when a signal or attribute (or attirbute group) came from a specific schema, so if we have multiple imports I can resolve that these are actually the same, and deal with version conflicts appropriately.  This is still draft so we can collectively consider that use case and what we need.

PR #1280, opened 2026-03-11
Fix resolution for dependencies

lmolkova:
Both v1 and v2 resolution has a problem when using dependencies - it returns random attribute from attribute catalog instead of definition resulting in wrong briefs/notes, etc.

E.g. in this yaml https://github.com/lmolkova/semconv-scale23x-demo/blob/965f9f0a6094e1d21150d4a0441ede7aafbb4a0b/conventions/src/registry.yaml#L36 (` - ref: server.address` from semconv), the `server.address` resolves to "Name of the database host." which comes from DB refinement.

https://github.com/lmolkova/semconv-scale23x-demo/blob/2d31c5bf2437f5936d8740a338845bcb06ced456/docs/storage/metrics.md#attributes-1

This PR fixes those. It does not solve multi-dependency/from-transitive-dependency resolution.

### DIFF
--- a/.typos.toml
+++ b/.typos.toml
@@ -1,5 +1,5 @@
 [files]
-extend-exclude = ["*.cast", "**/*.yaml", "**/*.svg", "**/allowed-external-types.toml"]
+extend-exclude = ["*.cast", "**/*.yaml", "**/*.json", "**/*.svg", "**/allowed-external-types.toml"]
 
 [default.extend-words]
 ratatui = "ratatui"


crates/weaver_resolved_schema/src/catalog.rs
```diff
@@ -15,13 +14,17 @@ use weaver_semconv::stability::Stability;
...
-    #[serde(default, skip_serializing_if = "Vec::is_empty")]
-    pub(crate) attributes: Vec<Attribute>,
+    /// Catalog of attribute definitions and refinements used in the schema.
+    /// Contains elements with the same attribute key.
+    /// Use root_attributes for original attribute definitions.
+    attributes: Vec<Attribute>,
+    /// Attribute definitions available in this registry (including those
+    /// from dependencies). Used for cross-registry attribute lookup.
+    /// Not serialized — populated only for freshly resolved schemas.
+    root_attributes: HashMap<String, (Attribute, String)>,
```
lmolkova, line 27: this is the real change for v1, let's populate root_attributes directly instead of trying to use lineage to infer if something in the catalog is original definition.

The rest in this file is cleanup - this struct should not allow to mutate attrs, does not need to be (de)serializable, many of its features were only used in tests

crates/weaver_resolver/data/multi-registry/otel_registry/otel_registry.yaml
```diff
@@ -9,6 +9,13 @@ groups:
         stability: stable
         requirement_level: recommended
         examples: "Ex"
+  - id: db.client.metrics
```
lmolkova, line 12: this is repro for the v1 issue

crates/weaver_resolver/data/registry-test-1-single-attr-ref/expected-attribute-catalog.json
```diff
@@ -1,30 +1,28 @@
...
-      "type": "string",
-      "brief": "The message destination name",
-      "examples": [
-        "MyQueue",
-        "MyTopic"
-      ],
-      "requirement_level": "recommended",
-      "note": "Destination name SHOULD uniquely identify a specific queue, topic or other entity within the broker. If\nthe broker doesn't have such notion, the destination name SHOULD uniquely identify the broker.\n",
-      "stability": "stable"
+[
```
lmolkova, line 1: no change here except unwrapping list of attrs - this allows to remove serialization from Catalog

crates/weaver_resolver/data/registry-test-v2-dep/app_registry/manifest.yaml
```diff
@@ -0,0 +1,8 @@
+description: App registry that depends on the consumer V2 package.
+schema_url: https://app.example.com/schemas/1.0.0
+dependencies:
+  - schema_url: https://consumer.example.com/schemas/1.0.0
+    # TODO: this only works with definition registry, but not with
+    # resolved one, because resolved does not know how to get
+    # attributes from transitive dependencies.
+    registry_path: data/registry-test-v2-dep/consumer_registry
```
lmolkova, line 8: fun around attrs from transitive dependencies:
- if you use definition as a dependency, you can use attributes in transitive one
- if you use resolved as dependency, you can't - they are not there

Let's make it consistent (regardless of the approach we take)

crates/weaver_resolver/src/attribute.rs
```diff
@@ -284,45 +300,20 @@ impl AttributeLookup for ResolvedDependency {
...
-                            }
-                        })
-                })
-                .map(|a| AttributeWithGroupId {
-                    attribute: a,
-                    group_id: g.id.to_owned(),
-                })
-        });
-        result
+        self.catalog
```
lmolkova, line 303: this is part of solution for v1 - use root attributes instead of refs, the logic here didn't work for dependencies

crates/weaver_resolver/src/attribute.rs
```diff
@@ -284,45 +300,20 @@ impl AttributeLookup for ResolvedDependency {
...
+            })
     }
 }
 
 impl AttributeLookup for V2Schema {
     fn lookup_attribute(&self, key: &str) -> Option<AttributeWithGroupId> {
         let fake_group_id = format!("v2_dependency.{}", self.schema_url.name());
-        self.attribute_catalog.iter().find_map(|attr| {
+        self.registry.attributes.iter().find_map(|attr_ref| {
+            let attr = self.attribute_catalog.get(attr_ref.0 as usize)?;
```
lmolkova, line 316: this is solution for v2 - attribute_catalog contains definitions and refinements. given lookup by key, we can only return the original definition from registry.attributes

crates/weaver_resolver/data/registry-test-v2-dep/app_registry/manifest.yaml
lmolkova: also more fun: since only one dependency is allowed, you can't really make 

A -> resolved B -> C work today, unless you can modify B to import things

crates/weaver_resolved_schema/src/lib.rs
```diff
@@ -61,22 +59,18 @@ pub struct ResolvedTelemetrySchema {
     /// and signals.
     pub catalog: Catalog,
     /// Resource definition (only for application).
-    #[serde(skip_serializing_if = "Option::is_none")]
```
jsuereth, line 64: Is this intentional?

lmolkova: yep, it does not need json schema - we don't (and don't need to) support it for v1. And it's not (de)serializable either

lmolkova: I'm trying to remove serialization from things that are not intended to be read or written by weaver. This helps to understand what's internal detail and what's a real contract

crates/weaver_resolver/data/multi-registry/otel_registry/otel_registry.yaml
jsuereth: Ah - I gotcha, we don't use `extends` here, so all the assumptions fail, and the catalog is ambiguous.  I thought it'd be safe to remove the hardcoded `registry.` paths for V1, but looks like not, thanks for catching this.

crates/weaver_resolver/data/registry-test-1-single-attr-ref/expected-attribute-catalog.json
jsuereth: OOF, yeah you can blame me for that change and I regret doing it.

I wonder if maybe serializing this to YAML or TOML or something would make this more readable / easier to diff in the future.  WDYT?

crates/weaver_resolver/src/attribute.rs
jsuereth: Doh! Yeah - I can't believe I missed this the first time.

PR #1282, opened 2026-03-12
Fix stability/ordering issues in v2 publishing

jsuereth:
- Add to the test infrastructure so we can test end-to-end v2 resolution in `weaver_resolver`
- Add a test case that USED to churn/fail due to inconsistent ordering.
- Fix ordering to force us to sort attributes once, and remove duplicates from the `attributes` registry.

### DIFF
--- a/crates/weaver_resolved_schema/src/v2/mod.rs
+++ b/crates/weaver_resolved_schema/src/v2/mod.rs
@@ -181,9 +181,7 @@ pub fn convert_v1_to_v2(
         })
         .collect();
 
-    let mut sorted_attributes = attributes.into_iter().collect::<Vec<_>>();
-    sorted_attributes.sort_by(|a, b| a.key.cmp(&b.key));
-    let v2_catalog = Catalog::from_attributes(sorted_attributes);
+    let v2_catalog = Catalog::from_attributes(attributes.into_iter().collect());
 
     // Create a lookup so we can check inheritance.
     let mut group_type_lookup = HashMap::new();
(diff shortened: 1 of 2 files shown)


crates/weaver_resolved_schema/src/v2/mod.rs
```diff
@@ -165,26 +165,25 @@ pub fn convert_v1_to_v2(
...
+                deprecated: a.deprecated,
+                annotations: a.annotations.unwrap_or_default(),
+            },
         })
         .collect();
 
-    let v2_catalog = Catalog::from_attributes(attributes.into_iter().collect());
+    let mut sorted_attributes = attributes.into_iter().collect::<Vec<_>>();
+    sorted_attributes.sort_by(|a, b| a.key.cmp(&b.key));
+    let v2_catalog = Catalog::from_attributes(sorted_attributes);
```
lmolkova: nit: maybe catalog should sort?

crates/weaver_resolver/data/registry-test-v2-5-published/expected-schema.yaml
```diff
@@ -0,0 +1,76 @@
+file_format: resolved/2.0.0
```
lmolkova, line 1: 🚀

crates/weaver_resolved_schema/src/v2/mod.rs
jsuereth: Done! much cleaner

PR #1298, opened 2026-03-17
BREAKING: Use `schema_url` to track registries consistently in lineage / provenance

jsuereth:
This stops using a `registry_id` and instead uses SchemaURL to track registires.  Existing registries are given a dummy SchemaURL based on their registry id, if needed.

This modifies *V1* schema, as we need to make this change to start supporting V2 and the migration to V2 syntax.

I do *not* believe this is a significant breaking change, but there are corner cases that may need to be tweaked.  Would appreciate review and consideration on other options.

### DIFF
--- a/crates/weaver_semconv/src/schema_url.rs
+++ b/crates/weaver_semconv/src/schema_url.rs
@@ -44,12 +44,12 @@ impl SchemaUrl {
             .find(full_path)
             .map(|i| i + scheme_end)
             .ok_or_else(|| "The schema URL has an invalid path structure.".to_owned())?;
-        
+
         let trimmed_path = full_path.trim_end_matches('/');
         let last_slash = trimmed_path
             .rfind('/')
             .ok_or_else(|| "The schema URL has an invalid path structure.".to_owned())?;
-        
+
         let name_range = scheme_end..(path_start + last_slash);
         let version_range = (path_start + last_slash + 1)..(path_start + trimmed_path.len());
 


crates/weaver_semconv/src/provenance.rs
```diff
@@ -38,7 +44,7 @@ impl Provenance {
     #[must_use]
     pub fn undefined() -> Self {
```
lmolkova: nit, I think it's only used as a fallback in

https://github.com/open-telemetry/weaver/blob/1754691dd6e05aa0bc0d722eb3f4b1a3d227773f/crates/weaver_resolved_schema/src/registry.rs#L459-L464

and can be constructed right there or, even better, Provenance should be optional. No provenance if better than `{"schema_url": "unknown/unknown", "path": "undefined"}`

crates/weaver_semconv/src/provenance.rs
```diff
@@ -2,34 +2,40 @@
...
 pub struct Provenance {
-    /// The registry id containing the specification file.
-    /// A registry id is an identifier defined in the `registry_manifest.yaml` file.
-    pub registry_id: Arc<str>,
+    /// The schema URL where this was specified.
+    ///
+    /// The Schema url contains the registry id and the version of the schema.
+    /// It can be used to detect conflicts or resolve multiple "ids" existing across
+    /// dependency chains but being the same thing, conceptually.
+    pub schema_url: SchemaUrl,
```
lmolkova, line 20: can you please add a changelog ?

crates/weaver_semconv/src/provenance.rs
jsuereth: Fair - I'll fix that up

crates/weaver_semconv/src/provenance.rs
jsuereth: Done

PR #1299, opened 2026-03-17
Add a 'filter playground'

jsuereth:
This adds a new tab for `weaver serve` that leverages the previous `filter` API endpoint I had.  This will let you write JQ filters and see the resulting JSON/YAML that is returned, so you can do a first step of debugging your templates.

I'd like to take this further, but I think this is a good first step to help.

- Updates `weaver_forge` errors on JQ to use the *span* and report this semantically so we can get it to the UI.
- Update `ui` to depend on Monaco editor, as we can do all sorts of cool stuff with it (may be too large, let me know).
- Wire in new tab for `Filter (jq) playground`, that has a text entry box, a selection for "yaml" or "json" and lets you see the result of your filter *or* the error from JQ, with highlighted areas.

The UI portion was vibe-coded, but I reviewed the typescript and basic structure and stand behind it.  The actual HTML/css design is no longer my strong suit, so feel free to perform extra review there.

### DIFF
--- a/crates/weaver_forge/src/jq.rs
+++ b/crates/weaver_forge/src/jq.rs
@@ -336,7 +336,10 @@ mod tests {
     #[test]
     fn test_report_io_error() {
         let detail = super::report_io(("test_file.jq", "permission denied".to_owned()));
-        assert_eq!(detail.error, "could not load file test_file.jq: permission denied");
+        assert_eq!(
+            detail.error,
+            "could not load file test_file.jq: permission denied"
+        );
         assert!(detail.source.is_none());
     }
 }


ui/src/routes/filter_playground.tsx
```diff
@@ -0,0 +1,157 @@
...
+
+  const handleExecute = async () => {
+    setLoading(true)
+    setErrorDetails(null)
+    try {
+      const data = await filterRegistry(filter)
+      setResult(data)
+    } catch (err: unknown) {
+      if (err instanceof ApiError) {
+        setError(err.message)
```
lmolkova, line 67: I think we never unset error or result, only override it.

E.g. try this

1. write bogus jq
2. execute
3. see error
4. fix jq
5. execute
6. error and output are both still there

Image

jsuereth: Fixed - Still new to react, so forgot how weirdly stateful everything is, while also not being stateful...

Issue #1300, opened 2026-03-17
Prebuilt binaries for  aarch64-unknown-linux-gnu


anomalizer:
**Is your feature request related to a problem? Please describe.**
I use an ARM64 based linux machine and would like weaver to be able to run on it. There is no pre-built binary generated for this hardware architecture. The download script confirms the same
```
➜  /tmp curl --proto '=https' --tlsv1.2 -LsSf https://github.com/open-telemetry/weaver/releases/download/v0.22.1/weaver-installer.sh | sh
ERROR: there isn't a download for your platform aarch64-unknown-linux-gnu
```

**Describe the solution you'd like**
Weaver release process to add support for publishing pre-built binaries

**Describe alternatives you've considered**
I built weaver from source on the target architecture and it is works. No modification was needed to the source tree. Commands used were as follows

```
cargo build --release
cd ui
pnpm install
pnpm build
cd ..
cargo build --release
```
Result
```
➜  ~/oss/weaver git:(main)  target/release/weaver
Manage semantic convention registry and telemetry schema workflows (OpenTelemetry Project)

Usage: weaver [OPTIONS] <COMMAND>

Commands:
  registry    Manage Semantic Convention Registry
  diagnostic  Manage Diagnostic Messages
  completion  Generate shell completions
  serve       Start the API server (Experimental)
  help        Print this message or the help of the given subcommand(s)

Options:
      --debug...  Turn debugging information on. Use twice (--debug --debug) for trace-level logs
      --quiet     Turn the quiet mode on (i.e., minimal output)
      --future    Enable the most recent validation rules for the semconv registry. It is recommended to enable this flag when checking a new registry. Note: `semantic_conventions` main branch should always enable this flag
  -h, --help      Print help
  -V, --version   Print version
➜  ~/oss/weaver git:(main)  target/release/weaver
Manage semantic convention registry and telemetry schema workflows (OpenTelemetry Project)

Usage: weaver [OPTIONS] <COMMAND>

Commands:
  registry    Manage Semantic Convention Registry
  diagnostic  Manage Diagnostic Messages
  completion  Generate shell completions
  serve       Start the API server (Experimental)
  help        Print this message or the help of the given subcommand(s)

Options:
      --debug...  Turn debugging information on. Use twice (--debug --debug) for trace-level logs
      --quiet     Turn the quiet mode on (i.e., minimal output)
[...]

PR #1301, opened 2026-03-18
support for arm64 linux with glibc

anomalizer:
fix for #1300


jsuereth:
Ok - verified the release builds succesfully with https://github.com/open-telemetry/weaver/actions/runs/24475977856/job/71826416949?pr=1363

PR #1302, opened 2026-03-19
fix: switch gix TLS backend from native-tls to rustls

jmacd:
Replace blocking-http-transport-reqwest-native-tls with blocking-http-transport-reqwest-rust-tls in weaver_common to avoid TLS conflicts for downstream consumers that use reqwest with rustls.

Remove the now-unnecessary openssl workspace dependency.

This helps with an otel-arrow PR attempting the weaver update translated into broken tests because of TLS mismatches [open-telemetry/otel-arrow#2368](https://github.com/open-telemetry/otel-arrow/pull/2368).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -690,16 +690,6 @@ dependencies = [
 
 [[package]]
 name = "core-foundation"
-version = "0.9.4"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


michaelvanstraten:
I checked this and it seamed to work (at least the package command with one git dependency).

jmacd:
I will open a new PR. I have refined my understanding of the requirements.

PR #1306, opened 2026-03-21
Support cloning private repositories

jerbly:
Added `--allow-git-credentials` global flag to use default git open options instead of isolated mode when cloning repositories, enabling system credential helpers (e.g. `osxkeychain`, `manager`, `libsecret`) for private repository access.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -259,7 +259,7 @@ jobs:
           | sarif-fmt
         continue-on-error: true
       - name: Upload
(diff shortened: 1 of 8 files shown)


crates/weaver_common/src/vdir.rs
```diff
@@ -392,7 +392,7 @@ impl VirtualDirectory {
                 destination_must_be_empty: true,
                 fs_capabilities: None,
             },
-            open::Options::isolated(),
+            open::Options::default(),
```
jsuereth: Should this be a flag in weaver?

PR #1313, opened 2026-03-23
Add provenance tracking for V2

jsuereth:
This adds the provenance tracking we had discussed during our list SIG.

- Weaver Resolved schema will track the following:
  - The whole schema has a list of dependencies, which are `schema_url`s
  - Every attribute in the catalog has a `Provenance` which will include a reference to the index of the dependency that defined it.
  - Every signal / attribute group will have the same Provenance.
  - Any attribute / signal defined *locally* to the schema (i.e. not imported) will NOT have a Provenance.
- This does NOT impact "forge" / template schema yet, that will be follow-on work and we may want to discuss what belongs in forge.

### DIFF
--- a/crates/xtask/src/schema_compat.rs
+++ b/crates/xtask/src/schema_compat.rs
@@ -28,11 +28,11 @@ const CHECKS: &[(&str, &str, &str)] = &[
         "schemas/publication-manifest.v2.json",
         "PublicationManifestV2",
     ),
-//     (
-//         "resolved-registry-v2",
-//         "schemas/semconv.resolved.v2.json",
-//         "ResolvedRegistryV2",
-//     ),
+    //     (
+    //         "resolved-registry-v2",
+    //         "schemas/semconv.resolved.v2.json",
+    //         "ResolvedRegistryV2",
+    //     ),
 ];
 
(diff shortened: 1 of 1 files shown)


crates/weaver_resolver/src/attribute.rs
```diff
@@ -56,6 +56,25 @@ impl AttributeCatalog {
             .or_insert_with(|| AttributeRef(next_id))
     }
 
+    /// Returns an attribute ref from the catalog and also registers the attribute
+    /// in `root_attributes` so its provenance is recorded.
+    pub fn attribute_ref_with_group(
+        &mut self,
+        attr: attribute::Attribute,
+        group_id: String,
```
lmolkova: it seems we only use this method to record provenance and not group id, would it make sense to rename it?

Also potentially we can add dependency schema url to AttributeWithGroupId and this way avoid using `v2_dependency` prefix as indication of attribute coming from dep?

jsuereth: Unfortunately, we need to fully go into V1 to come back as V2.  We could if I further break V1 schema - but I figure this is temporary.

Did some cleanup and added a big comment explaining.

PR #1334, opened 2026-04-05
Use semantic conventions v2 for `weaver registry infer`

ArthurSens:
Continuing the work we agreed on, this PR migrates the `weaver registry infer` command to the schema file's second version.

There are a few open questions from my side, which I'd love some input from maintainers:

* From an OTLP message with Resource Attributes, but no EntityRef fields populated, should we be creating a single entity with all attributes as identifying? Should we not create Entities at all? In v1, we created a single resource group, but I'm not sure what to do in v2.
* For Spans, I'm using the span name for both `span.type` and `span.note`. Should I be using something else for them?
* I know span events will be deprecated, but for now, I'm joining span events and log events with the same name into the same Event record. Is that ok?

### DIFF
--- a/crates/weaver_infer/src/lib.rs
+++ b/crates/weaver_infer/src/lib.rs
@@ -253,7 +253,9 @@ impl AccumulatedSamples {
                     .keys()
                     .map(|name| span_attribute_ref(name))
                     .collect::<Vec<_>>();
-                attributes.sort_by(|left, right| span_attribute_ref_name(left).cmp(span_attribute_ref_name(right)));
+                attributes.sort_by(|left, right| {
+                    span_attribute_ref_name(left).cmp(span_attribute_ref_name(right))
+                });
 
                 Span {
                     r#type: SignalId::from(span.name.clone()),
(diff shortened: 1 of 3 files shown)


jsuereth:
> From an OTLP message with Resource Attributes, but no EntityRef fields populated, should we be creating a single entity with all attributes as identifying? Should we not create Entities at all? In v1, we created a single resource group, but I'm not sure what to do in v2.

You can either create one entity with everything or none. I think I'd prefer none until we get Entity populated, but I can see having *something* inferred better than nothing.  We'll also need to deal with "loose" attributes for some time, so we should probably figure out the right behavior here.

> For Spans, I'm using the span name for both span.type and span.note. Should I be using something else for them?

We don't have a formal `span.type` in  OTLP yet, so that behavior is probably the right way to go for now, and we can evolve the `span.type` over time.

> I know span events will be deprecated, but for now, I'm joining span events and log events with the same name into the same Event record. Is that ok?

YES! I think that should even help us with the migration from span-event to just event.

crates/weaver_infer/src/lib.rs
```diff
@@ -205,110 +214,163 @@ impl AccumulatedSamples {
...
-///
-/// Note: We use this wrapper instead of `SemConvSpecV1` directly because
-/// `SemConvSpecV1.groups` is `pub(crate)` in `weaver_semconv`.
-/// This wrapper produces the same YAML structure as a valid semconv file.
+/// Wrapper for serializing an inferred v2 semconv registry file.
 #[derive(Serialize)]
-pub struct InferredRegistry {
-    /// Inferred semantic convention groups.
-    pub groups: Vec<GroupSpec>,
+pub struct InferredRegistryV2 {
```
jsuereth: Is there a reason this is different than the actual struct/type we use for V2?

This seems to mostly by copy-paste from  here - https://github.com/open-telemetry/weaver/blob/main/crates/weaver_semconv/src/v2/mod.rs#L54

You should be able to directly use that struct and wrap it in `Versioned::V2(...struct...)` when you need the manifest version.

ArthurSens: Mostly because it was all private in that crate. I've implemented constructors and accessors for it now :)

ArthurSens:
Comments addressed!

ArthurSens:
Anything else needed here? :)

jsuereth (approved):
This looks good - but probbaly needs a ton of merge conflicts.  Sorry I dropped the re-review @ArthurSens

PR #1355, opened 2026-04-13
chore: remove unused openssl dependency

michaelvanstraten:
The dependency does not looked used anywhere.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3308,15 +3308,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7c87def4c32ab89d880effc9e097653c8da5d6ef28e6b539d313baaacfbafcbe"
 
 [[package]]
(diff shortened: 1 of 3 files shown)


jsuereth:
@michaelvanstraten It's used for our docker distribution.  We're using a minimal container, and we don't have access to an SSL from the platform, so we bundle one.

michaelvanstraten:
Okay, it was messing up my build pipeline because I did not have the build tools installed to compile openssl, maybe I make a PR to enable it optionally for the docker distribution.

jsuereth:
Related - https://github.com/open-telemetry/weaver/pull/1302

I'd prefer NOT to use openssl at all if we can - Just didn't have time to fix the docker image myself.  I'll throw tokens/agent at it at some point to see if it can figure it out, but it hasn't bubbled to the top of our list.

PR #1356, opened 2026-04-14
Pull registries from HTTP(S) URLs, with private-source auth

jerbly:
Adds HTTP Bearer token authentication for downloading remote registries from private HTTP(S) sources (closes #1344). Primary use case: pulling published registry artifacts (manifest.yaml + resolved.yaml) from private GitHub release assets.

- New RemoteFile source type for VirtualDirectoryPath - supports non-archive, non-git HTTP(S) URLs.
- HTTP auth is configured per-URL via `[[auth]]` entries in `.weaver.toml`. Each entry pairs a `url_prefix` (longest match wins) with a token source: a literal `token`, a `token_env` variable name, or a `token_command` helper whose first stdout line is the token (e.g. `["gh", "auth", "token"]`). Matched requests get `Authorization: Bearer <token>` and a `User-Agent: weaver` header.
- GitHub URL normalization - browser-style URLs like https://github.com/org/repo/releases/download/v1.0.0/manifest.yaml are transparently resolved to GitHub API asset URLs (which accept Bearer auth). Release metadata is cached per-release to avoid redundant API calls.
- Shared ureq Agent with RedirectAuthHeaders::SameHost so auth headers survive GitHub's same-host redirects but are stripped on cross-origin redirects.

Breaking change: A bare HTTP(S) URL with no .git/.zip/.tar.gz suffix and no refspec or [sub_folder] now parses as RemoteFile instead of GitRepo. Add .git to such URLs to keep the old behaviour.


crates/weaver_common/src/vdir.rs
```diff
@@ -76,6 +108,174 @@ pub fn is_git_credentials_enabled() -> bool {
     ALLOW_GIT_CREDENTIALS.load(std::sync::atomic::Ordering::Relaxed)
 }
 
+/// Optional Bearer token for authenticating HTTP requests when downloading
+/// remote archives and files. Set via `set_http_auth_token()`, typically from
+/// the `WEAVER_HTTP_AUTH_TOKEN` or `GITHUB_TOKEN` environment variable at startup.
+static HTTP_AUTH_TOKEN: Mutex<Option<String>> = Mutex::new(None);
```
jsuereth, line 114: Does this work if we start having multi-threaded servers using the weaver engine?

Could we wind up with conflicts for using both HTTP and GITHUB urls at the same time?

jerbly: I don't think I follow. You can only have one token with this implementation, do you mean that we may need different tokens to fetch different dependencies?

Regarding URLs, the type is detected for each download. 
```
//! # Disambiguating HTTP(S) URLs
//!
//! An HTTP(S) `source` is classified as follows (in order):
//! 1. `.zip` or `.tar.gz` suffix → remote archive (may carry a `[sub_folder]`).
//! 2. `.git` suffix, or presence of `@refspec` or `[sub_folder]` → Git repo. Once
//!    archives are ruled out, a `@refspec` or `[sub_folder]` is a reliable signal
//!    of a Git repo, so the `.git` suffix is not required.
//! 3. Otherwise → remote file.
```

jsuereth: Yes - the idea we need different for different dependencies, or, e.g. where we're pulling in `schema_url` dynamically in live-check

jerbly: Oof! That will be really ugly - is that realistic? In my case it's company reg depends on otel reg: company needs token, otel does not. This would work today.

We could have company->company->company...->otel too, the same token would be valid in all those private locations.

We'd have to have env vars associated with each private location. I guess that's doable if we think it's needed?? Perhaps defer to another issue/pr later?

jsuereth: See - https://doc.rust-lang.org/cargo/reference/config.html#credentials or https://maven.apache.org/settings.html#Servers

I think once your `weaver.toml` stuff lands this would be a good spot for it.

jsuereth (approved):
I believe there *is* a non-breaking way to support multiple credentials per-domain you resolve from *and* this flag at the same time.

Specifically I think we can have a url - token config, similar to cargo, in the future that would "win" over this broad credential config.

Approve-ing but want to *confirm* you agree with me first, @jerbly. WDYT?

jerbly:
@jsuereth - I think this is good now. Requested re-review since it's changed a lot since you looked last. This is what I demoed to you a couple of weeks ago.

jsuereth (approved):

PR #1359, opened 2026-04-14
fix(resolved-schema): Stable sort when converting from v1 to v2

michaelvanstraten:
When converting from a v1 catalog to a v2 catalog we collect the attributes in a `HashSet` and then call `iter()` on that hashset. We then sort the resulting vector by the attribute key. The issue is that multiple attributes with the same key are ordered randomly because of the `RandomState` of the `std::collection::HashSet`.

This patch addresses this issue by adding a secondary sorting key based on the hash of the attribute.


jsuereth (approved):

michaelvanstraten:
@jsuereth could you merge this.

jsuereth:
@michaelvanstraten Yes - was just waiting for tests to re-pass after conflict resolution.

michaelvanstraten:
Thank you, and thank you for the many quick reviews the last few days.

PR #1365, opened 2026-04-16
Tolerate unknown properties in resolved schema / published manifest

lmolkova:
Older weaver can now read files written by a newer minor version. 
                                                                                                                                                                                                                                                               
- Newer minor (resolved/2.X on a 2.0 build): loads, unknowns ignored, warning that newer version is available is logged.                                                                                                                                                                       
- Major mismatch (e.g. resolved/3.0): not supported, fatal error at start
- Same/older minor with unknown field: fatal error listing unexpected fields (path). e.g. registry.spans[0].unknown_field

> [!WARNING]
> 
> It introduces regression for definition side (we reuse quite a few types between definition and resolved).
> Since deny_unknown_fields is relaxed, we'll tolerate errors in definitions that we failed on before.
> 
> Definitions PR: https://github.com/open-telemetry/weaver/pull/1422

### DIFF
--- a/crates/weaver_common/src/file_format.rs
+++ b/crates/weaver_common/src/file_format.rs
@@ -1,0 +1,212 @@
+// SPDX-License-Identifier: Apache-2.0
+
+//! File-format version identifiers of the form `"prefix/MAJOR.MINOR"`,
+//! shared by manifest and resolved-schema files.
+
+use crate::Error;
+use schemars::{json_schema, JsonSchema, Schema, SchemaGenerator};
+use serde::{de, Deserialize, Deserializer, Serialize, Serializer};
+use std::borrow::Cow;
+use std::str::FromStr;
+
+/// A file-format identifier of the form `"prefix/MAJOR.MINOR"`, e.g. `"manifest/2.0"`.
+///
+/// Serializes and deserializes as the string form (e.g. `"resolved/2.0"`).
+#[derive(Clone, PartialEq, Eq, Hash)]
+pub struct FileFormat {
+    /// The format prefix (e.g. `"manifest"`, `"resolved"`).
+    pub prefix: Cow<'static, str>,
+    /// The major version.
+    pub major: u32,
+    /// The minor version.
+    pub minor: u32,
+}
+
+impl FileFormat {
+    /// Construct a [`FileFormat`] from its parts.
+    #[must_use]
+    pub const fn new(prefix: &'static str, major: u32, minor: u32) -> Self {
+        Self {
+            prefix: Cow::Borrowed(prefix),
+            major,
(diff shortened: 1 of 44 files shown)


crates/weaver_resolved_schema/src/v2/version.rs
```diff
@@ -0,0 +1,345 @@
+// SPDX-License-Identifier: Apache-2.0
+
+//! Version parsing and forward-compatibility validation for resolved schemas.
+
+use crate::v2::ResolvedTelemetrySchema;
+
+/// Parsed file format version.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct FileFormatVersion {
```
lmolkova: should we replace string constants with it?

crates/weaver_semconv/src/manifest.rs
```diff
@@ -229,6 +242,87 @@ impl RawManifestFields {
             }))
         }
     }
+
+    /// Check if a file_format string identifies a publication manifest (prefix is "manifest").
+    fn is_publication_manifest(file_format: &str) -> bool {
```
lmolkova: this should be a method on file_format and it should not be a string?

crates/weaver_common/src/file_format.rs
```diff
@@ -0,0 +1,212 @@
...
+use schemars::{json_schema, JsonSchema, Schema, SchemaGenerator};
+use serde::{de, Deserialize, Deserializer, Serialize, Serializer};
+use std::borrow::Cow;
+use std::str::FromStr;
+
+/// A file-format identifier of the form `"prefix/MAJOR.MINOR"`, e.g. `"manifest/2.0"`.
+///
+/// Serializes and deserializes as the string form (e.g. `"resolved/2.0"`).
+#[derive(Clone, PartialEq, Eq, Hash)]
+pub struct FileFormat {
```
jsuereth, line 16: @lmolkova I approve of most of this PR - but wanted to check in on your plans based on the discussion today.

I'm fine leaving this here with some flexibility we discussed, or addressing file-format expansion in a later PR.

Should I approve this PR as-is?

lmolkova: this PR does not touch anything related to definition, I was going to address it in separate PR (within this release). So I think this one can go as is

lmolkova: here's the definition part https://github.com/open-telemetry/weaver/pull/1422

jsuereth (approved):
@lmolkova FYI - I forgot to approve this oh so long ago - Approving now if we still consider this a blocker - hopefully it's easy to bring this back up to date and merge if we want it.

PR #1368, opened 2026-04-16
Allow multiple dependencies to be defined. 

### LINKED ISSUE #1358: Registry dependency limit
https://github.com/open-telemetry/weaver/pull/1254 added limit on dependencies in semantic convention registry: https://github.com/open-telemetry/weaver/blob/be6d4a754c0ba54db77433c8579b614ddbcd7007/crates/weaver_resolver/src/loader.rs#L211-L213

There is a lot of truly great work being done to support multi-registries, and from my basic testing I can see that multiple dependencies are merged correctly if limit is removed. In my option having multi dependency registries makes semantic conventions and Weaver truly powerful tool.

So my question is: do dependency limit of 1 still stands or is still needed? If yes, maybe you can help understand reasoning about it and required work to be done to remove limit?

jsuereth:
Does NOT resolve ref conflicts between dependencies or resolve version conflicts.

This sort of fixes #1358

What this does:

- Allow multiple dependencies to be defined.
- uses `schema_url` to figure out if something is a dependency.

What this does NOT do

- Use local repo's `schema_url` in any dependency resolution (i.e. we COULD have a full circle and that'd be bad).
- Attempt to resolve `ref` attributes or signals *between* dependencies with syntax - this just declares that an error and gives up.
- Allow *compatible* dependencies to come from multiple paths - I.e. if both B and C depend on A at different version numbers, that would still be an error for now.

This is just starting to make progress and evaluate a smaller scope.  I think there may be a more significant refactoring to solve the above, so leaving this as draft while I do some more testing.


jsuereth:
So there are a lot of shortcuts in this PR i had to fix in https://github.com/open-telemetry/weaver/pull/1377.  Debating closing this for part 2 directly.  Will check in with maintainers.

jsuereth:
discussed with @jerbly - in the interest of smaller PRs, going to push this one for review first before the layer on top of it.

crates/weaver_resolver/data/compatible-version-conflict/main/registry_manifest.yaml
```diff
@@ -0,0 +1,9 @@
+name: main
```
lmolkova, line 1: nit, could you switch tests to v2? probably it's not important for any logic, but helps to see and validate v2 syntax, plus build enough examples with v2 that's useful for docs, agents, etc

lmolkova (approved):

crates/weaver_resolver/data/compatible-version-conflict/main/registry_manifest.yaml
jsuereth: As discussed in the SIG call - we'll merge this PR and fix up test cases in part 2.

PR #1377, opened 2026-04-20
Wip allow multiple dependencies part 2 - Use semver to resolve *some* dependency conflicts

jsuereth:
With multiple dependencies, if two depend on the same parent, e.g. two sub-registries depend on semconv core, then we can resolve discrepencies by picking the "latest" version of semconv, assuming we do NOT have a major version conflict.

Note: Until semconv goes 1.x this will *conflict for every releease*.

cc @open-telemetry/specs-semconv-maintainers for awareness.


crates/weaver_resolver/src/loader.rs
```diff
@@ -190,31 +191,63 @@ fn load_semconv_repository_recursive(
...
             registry_name: registry_name.clone(),
             chain: chain_str,
         });
     }
-    // Add current registry to visited set and dependency chain
-    let _ = visited_registries.insert(registry_name.clone());
+
+    // Check for conflict across the graph
+    if let Some(prev_schema_url) = visited_registries.get(&registry_name) {
+        if prev_schema_url != &schema_url {
```
jsuereth, line 246: This should pick latest version make sure visited registries has the *latest* version in it.

jsuereth: Fixed

jsuereth:
Now stale - replaced by another PR

PR #1380, opened 2026-04-22
Remove vendored openssl dependency

sapatrjv:
The OTel Dataflow Engine (aka otel-arrow) has been using Weaver. In weaver due to dependency on openssl vendored feature it causes openssl build everytime. This change is made to avoid such overhead. With this change weaver will rely on windoows native tls implementation. 

Part of #1382.

Part of https://github.com/open-telemetry/otel-arrow/issues/2697.

### DIFF
--- a/crates/weaver_common/Cargo.toml
+++ b/crates/weaver_common/Cargo.toml
@@ -34,10 +34,15 @@ gix = { version = "0.81.0", default-features = false, features = [
     "worktree-mutation",
     "sha1",
 ] }
-openssl.workspace = true
 flate2 = "1.1.4"
 tar = "0.4.43"
 zip.workspace = true
+
+# Use vendored OpenSSL on non-Windows platforms to avoid system OpenSSL deps
+# This enables the "vendored" feature for the openssl crate when building on
+# non-Windows targets.
+[target.'cfg(not(windows))'.dependencies]
+openssl.workspace = true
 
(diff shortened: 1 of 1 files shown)


crates/weaver_common/Cargo.toml
```diff
@@ -34,11 +34,16 @@ gix = { version = "0.81.0", default-features = false, features = [
...
 ] }
-openssl.workspace = true
 flate2 = "1.1.4"
 tar = "0.4.43"
 zip.workspace = true
 
+# Use vendored OpenSSL on non-Windows platforms to avoid system OpenSSL deps
+# This enables the "vendored" feature for the openssl crate when building on
+# non-Windows targets.
+[target.'cfg(not(windows))'.dependencies]
```
jsuereth, line 44: Could we change this so we only vendor on MUSL deployments? I think that's the only platform where we ran into trouble and needed it...

jsuereth (approved):
Thanks for issuing a fix here!  I'm wondering if we can more aggressive at turning off the OpenSSL requirement

jsuereth:
We discussed in semconv tooling SIG - We're going to work on extracting the entire SSL embedding as a `feature` of weaver over time, and dramatically clean up our build so things we need for a docker container don't bleed into decisions for platform-specific distributions.


This is good to merge for now.

PR #1381, opened 2026-04-22
Add metric requirement level

### LINKED ISSUE #406: Add metric requirement level to yaml
This stems from https://github.com/open-telemetry/opentelemetry-specification/issues/2972 and https://github.com/open-telemetry/opentelemetry-specification/pull/3158/files 

> "The [HTTP metrics semantic convention spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/http-metrics.md) should provide clarity "what is considered a compliant implementation", e.g. "an implementation is considered compliant if it has implemented all of the metrics described by this spec"

The work required in build-tools is: 
* Add a field to the metrics semconv yaml (should this also be for traces?)
* Decide where this will be rendered in markdown (most likely the metrics table makes sense) + write it into the markdown renderer

joaopgrassi:
Fixes #406

This adds the metric requirement level to the model/schema. This will enable us to auto-generate the entire metric markdown pages, without having to enter the metric name + requirement level by hand, how it is today (which is prone to getting out-of-date as seen [here](https://github.com/open-telemetry/semantic-conventions/commit/819c69b5ee0387f5e872ecc6ec544a07ec2f4d7e))

I used AI (Sonnet) to help me with getting the main changes in and added the tests myself. Of course fully reviewed it before opening this :)


crates/weaver_semconv/src/v2/metric.rs
```diff
@@ -43,6 +44,9 @@ pub struct Metric {
     #[serde(default)]
     #[serde(skip_serializing_if = "Vec::is_empty")]
     pub entity_associations: Vec<String>,
+    /// The requirement level of the metric. Defaults to 'recommended' when omitted.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub metric_requirement_level: Option<BasicRequirementLevelSpec>,
```
jsuereth: Nit: We should just call this `requirement_level` here - We're in a specific `metrics:` section, so it's mroe obvious that there should be one.

jsuereth (commented):
Nice fixes!!!

A few thoughts --

I'm less concerned with naming for V1 - and that change is ok, we can ignore features that aren't coming to V2 syntax.

For V2 - I have a few questions:

1. Should this just be `requirement_level` on the `metric` section?  Given there's no ambiguity in V2 now, I'd prefer to keep it simple/consistent.
2. If I'm refining/extending a metric, should I be able to change requirement level?  My gut reaction is 'no', but wanted to check.

crates/weaver_semconv/src/v2/metric.rs
joaopgrassi: Done [2a1be84](https://github.com/open-telemetry/weaver/pull/1381/commits/2a1be843ed98ccacbdbfe5bd196d52ec2c430bf6). I think I changed all places where necessary.

crates/weaver_semconv/src/v2/metric.rs
```diff
@@ -43,6 +44,9 @@ pub struct Metric {
     #[serde(default)]
     #[serde(skip_serializing_if = "Vec::is_empty")]
     pub entity_associations: Vec<String>,
+    /// The requirement level of the metric. Defaults to 'recommended' when omitted.
+    #[serde(skip_serializing_if = "Option::is_none")]
```
jsuereth, line 48: Would like others thoughts here - given we made requirement_level *required* in other settings, should we make it *required* to exist in V2 when defining metrics?

cc @open-telemetry/weaver-maintainers

joaopgrassi: I took your suggestion about making it required if the `--future` flag is passed, to avoid breaking everyone using weaver but still make it fail for semconv. [f1a4cdf](https://github.com/open-telemetry/weaver/pull/1381/commits/f1a4cdf3bdc8e7528519d8d6fc1b6ccaba25e548). Let me know if I found the correct way to do it.

jsuereth (approved):

crates/weaver_semconv/src/v2/metric.rs
jsuereth: Absolutely right! Thanks

PR #1386, opened 2026-04-23
fix(ui): support slash-containing attribute routes in Weaver UI

rbagd:
I am using Weaver registry to document telemetry data we currently have in our company using Weaver. Not all of the attributes are "compliant" yet with semconv, e.g. today our Filebeat Kubernetes ingestion adds labels with slashes which is generally fine for Elastic.

```
kubernetes.labels.app_kubernetes_io/name
kubernetes.labels.app_kubernetes_io/version
kubernetes.labels.app_kubernetes_io/managed-by
...
```

Unfortunately, slashes don't act friendly with the current UI routing. This PR is a just proposal to fix this minor inconvenience while keeping the rest unchanged.

**Before**

image

---

**After**

image


ui/src/routes/attribute.$key.tsx
```diff
@@ -74,7 +74,7 @@ function AttributeDetail() {
                 <div className="text-sm">{data.deprecated.note || 'This attribute is deprecated.'}</div>
                 {data.deprecated.renamed_to && (
                   <div className="text-sm mt-1">
-                    Use <Link to="/attribute/$key" params={{ key: data.deprecated.renamed_to }} className="link">{data.deprecated.renamed_to}</Link> instead.
```
jsuereth, line 77: Would it instead make sense to *escape* the `/` in Weaver UI here?

I understand this can fix a single `foo/bar`, but won't we have the same issue with `foo/bar/baz` ?

rbagd: `$` in `/attribute/$` now [gets treated as a wildcard](https://tanstack.com/router/latest/docs/routing/routing-concepts#splat--catch-all-routes), and so `_splat` should capture all the remainder. Just ran a sanity check and it still seems to be OK.

image

jsuereth: Ah nice - I'm still new to some of these UI frameworks we're using.

Running tests - thanks for the patch!

jerbly (approved):
LGTM

PR #1393, opened 2026-04-28
Gix update from 0.81 to 0.83 to get rid of dependency on winnow

sapatrjv:
Due to indirect multi-version dependency on winnow package through gix-config & gix-actor packages, gix version 0.81 consumption in otel-arrow project causing build failure.

Gix version 0.83 has removed dependency on winnow.  

image


lquerel (approved):

sapatrjv:
Testvfailure is due to issue with gix 0.83 unable to parse tags. It is fixed in https://github.com/GitoxideLabs/gitoxide/pull/2555.

Once latest bits available I will update the gix appropriately.

jsuereth:
Note: Accepted a separate PR that bumped gix *and* fixed the failing tests.

I'm going to close this as I think this accomplishes the goal, thanks for the contribution!

PR #1400, opened 2026-05-01
fix(weaver_resolved_schema): Missing `ToSchema` derive on `Provenance`

michaelvanstraten:
We apparently don't do a `cargo check --all-features -p weaver_resolved_schema` on CI so this was not checked. I will add a PR next week to check all crate features in CI.


jerbly (approved):
LGTM

Issue #1296, opened 2026-03-16
Ergonomic way of reusing entities


mmanciop:
**Is your feature request related to a problem? Please describe.**
At Dash0 we publish our own [semantic conventions](https://github.com/dash0hq/dash0-semantic-conventions). We encountered a limitation when modelling log events that require specific resource attributes. We want the log events of `event_name=dash0.deployment` to carry both the service identity (`service.name` + optionally `service.namespace` and `service.version`) _and_ `deployment.name`, but to do so, we need to maintain a separate entity, rather than saying that the resource of the log event needs to fulfil both. (Also: the `entity_associations` key sounded like "all must match" instead of the actual "any must match", and that was a fair bit of debugging to figure it out.

**Describe the solution you'd like**
I would like to be able to specify the expected location of an attribute in groups ("we expect `service.name` in the resource attributes"). It might also be helpful for OTel secomvs: I came across a lot of practitioners, especially those coming from Jaeger and Prometheus backgrounds, that had no idea that some attributes go into a *specific* attribute bag (or many!) in OpenTelemetry.

**Describe alternatives you've considered**
One option would be to allow entities to inherit each other, but I suspect that would play badly with OTel Entities sent by SDKs, which likely won't have such an inheritance mechanism. 

**Additional context**
I like Weaver :-)

PR #1426, opened 2026-05-10
Live check entity associations

### LINKED ISSUE #1297: Weaver Live Check seems not to validate required resource attributes
**Describe the bug**
Live check does not seem to complain when required attributes in entities are missing.

**To Reproduce**
1. Clone https://github.com/dash0hq/dash0-semantic-conventions/tree/main/tests
2. Mark the `deployment.name` attribute in `model/deployment/entities.yaml` as required
3. Run `make test` and see it NOT FAIL the `tests/valid/deployment/minimal-deployment-event.json` test

**Expected behavior**
Live check screams at me

**Additional context**
I hope I am not holding it wrong.

jerbly:
Fixes: #1297

Adds entity live-checks. 
image


crates/weaver_live_check/model/live_check.yaml
```diff
@@ -63,6 +63,23 @@ attributes:
           value: "conditionally_required_attribute_not_present"
           brief: "A conditionally required attribute is absent from the sample"
           stability: development
+        # Entity requirement-level findings (resource attributes checked against entity associations)
```
jsuereth, line 66: Yay dogfooding!!!

crates/weaver_live_check/model/live_check.yaml
```diff
@@ -63,6 +63,23 @@ attributes:
...
+          brief: "A recommended entity attribute is absent from the resource"
+          stability: development
+        - id: entity_opt_in_attribute_not_present
+          value: "entity_opt_in_attribute_not_present"
+          brief: "An opt-in entity attribute is absent from the resource"
+          stability: development
+        - id: entity_conditionally_required_attribute_not_present
+          value: "entity_conditionally_required_attribute_not_present"
+          brief: "A conditionally required entity attribute is absent from the resource"
+          stability: development
```
jsuereth, line 82: Side question - How are planning to do stability/release for these?  You can answer offline.

crates/weaver_live_check/src/sample_log.rs
```diff
@@ -87,6 +90,35 @@ impl LiveCheckRunner for SampleLog {
...
+        let resource_attributes: &[SampleAttribute] = parent_signal
+            .resource()
+            .map(|r| r.attributes.as_slice())
+            .unwrap_or(&[]);
+        let entity_associations: &[String] = match semconv_event.as_deref() {
+            Some(VersionedSignal::Group(g)) => &g.entity_associations,
+            Some(VersionedSignal::Event(e)) => &e.entity_associations,
+            _ => &[],
+        };
+        for entity_type in entity_associations {
```
jsuereth, line 103: So we only require *one of* the entities in entity associations to match.


I think as a first draft this is good - as we don't have any entities with multiple associations yet.

However the model *allows* any-of matching, so e.g. we wouldn't want to emit violations of one entity association if another one was "clean" and we don't track entity type in OTLP yet (It's there, but it's not used).

I think this is OK for now - but we should document this limitation in weaver or somehow denote it in the error message.

jsuereth (approved):

crates/weaver_live_check/src/sample_log.rs
jerbly: Oh I see. I did not realise that. I just assumed the entities were ANDED. So does "clean" mean that all the required attributes from an entity are satisfied? We could have logic:

Given entities A,B,C - if A is clean, report findings for non-required attributes on A only. If A is dirty, look at B and so on. If all are dirty report findings for all.

Would that work? Must there be at least one required attribute on an entity?

jsuereth: So entities have *identity* attribute and and *description* attributes.

You can detect presense if all the identity attribute are there.  *implicilty* (in v2) *all* identity attributes are required.

For description - SOME may be required (although that hasn't happened yet), bu tshes can be warnings.

So for your logic - I'd use the identity attribute to determine if "A" is present, e.g. and if you find none, warn that you didn't find an identity for one of the given entities.

Also - at least for semconv - we're keeping identity attributes *distinct* so there should be no overlap where one attribute could be the id of two entities in *semconv*, you can decide if we allow that in weaver - I'm on the fence myself.  Today it's basically allowed because it's not disallowed (we'd have to detect and error)

jerbly:
Just noting - merging this support now but it will change when #1296 is resolved before the next release.

PR #1436, opened 2026-05-11
test: use dynamic port for serve UI test

### LINKED ISSUE #1435: Stabilize some tests by using portpicker
We're not using portpicker in every test and occasionally we are getting port collisions. These collisions _could_ be caused by other tests.

nanookclaw:
## Summary

Closes #1435.

This updates the UI serve integration test to pick an available local port at runtime instead of always binding to 127.0.0.1:9080. The test already has portpicker available as a dev dependency, so this keeps the change limited to the flaky test setup while preserving the same server and HTML assertions.

## Testing

- pnpm install --frozen-lockfile && pnpm build (in ui/)
- cargo test --test serve_ui --no-default-features
- cargo fmt --check
- git diff --check

### DIFF
--- a/tests/registry_emit.rs
+++ b/tests/registry_emit.rs
@@ -8,6 +8,8 @@ use std::thread::sleep;
 use std::time::Duration;
 use tempfile::tempdir;
 
+use portpicker::pick_unused_port;
+
 /// This test verifies the roundtrip functionality of registry live check and emit commands.
 /// This test doesn't count for the coverage report as it runs separate processes.
 #[test]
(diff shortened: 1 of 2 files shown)


jerbly (commented):
Thanks for this but the issue was not just about the serve ui test. There are other tests that have hardcoded ports e.g. tests/registry_emit.rs. To close the issue we need to find them all.

nanookclaw:
Expanded this to cover `tests/registry_emit.rs` as well in f7466cc. The live-check/emit tests now allocate OTLP and admin ports with `portpicker`, including the v2 and resource-attributes cases, so the PR no longer only handles `tests/serve_ui.rs`.

Local checks run:
- `cargo fmt --check`
- `git diff --check`
- `cargo check --test registry_emit`

jerbly (approved):
LGTM - Thanks!

PR #1441, opened 2026-05-14
Scoped permissions in github workflow actions

jerbly:
Attempting to fix the scorecard token permission issues.

Sadly one of these is in the dist generated release.yml. This PR therefore has a script to correct the yml after a `dist generate`. Also added to the justfile.

Another downside is that we now need to bypass dist's generate --check so we can keep the hand-patched permissions in release.yml 

Is this trade-off worth it to close a high severity scorecard issue?


jsuereth:
I think permissions are a very real threat - I think it's probably worth it here - we could check to see if DIST has a fix to lock down permissions yet...

scripts/fix_release_permissions.py
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env python3
```
jsuereth: This is my only complaint - do I need to go install python to work on weaver now?

jsuereth (approved):
Approving - with the assumption we fix this up as we go.  Your question is good - let's secure down the build then evolve.

scripts/fix_release_permissions.py
jerbly: I guess if you're on windows you might not have it... I remade it as a cargo xtask.

PR #1442, opened 2026-05-14
Update weaver so that it ONLY imports, even when using the flag.

jsuereth:


jerbly:
@jsuereth I tried this locally and it's consistent. So, good to go once you've fixed clippy.

crates/weaver_resolver/src/dependency.rs
lmolkova, line 1: can you please add a changelog record?

lmolkova (approved):
Tried out https://github.com/open-telemetry/opentelemetry-python-genai/pull/55 as an alternative to --include-unreferenced and it also works nicely. 

Even without config, this PR does not break old behavior, just changes how it's implemented, so we should be good to go!

crates/weaver_resolver/src/dependency.rs
jsuereth: Done

PR #1444, opened 2026-05-16
fix: default enum member value to id

### LINKED ISSUE #733: Semconv schema: enum types - `value` is optional, inherit it from the `id` when not present
We've made a couple of similar mistakes in the past when defining/renaming enum members:

(still there now)

https://github.com/open-telemetry/semantic-conventions/blob/991d6caa40f5b25352ac3f772981662870e96faa/model/gen-ai/registry.yaml#L174-L176

and 

(fixed on main)

https://github.com/open-telemetry/semantic-conventions/blob/19f2ba7d0c0b0adc1b9efb8ab189c5300e038e4d/model/registry/messaging.yaml#L144-L146
      

Where we have different values between `id` and `value`. It creates confusion:
- codegen uses ID as a constant name and value as a value, but usually those are the same
- we enforce back-compat for id and not for the value, but changing value would still be a breaking change
- value is not subject to `a-z0-9._` charset rule, while id is. In some cases enum value becomes part of the attribute name (`db.system=foo` and `db.foo.property=bar`)  and then needs to be `a-z0-9._` - it's not enforced in any way

----
Proposal:

- Option 1: stop using `value`.

  - don't allow values that's different from `id` - don't support non `a-z0-9._`  chars in values. 
  - weaver would need to populate value from id for codegen back-compat.
  - in the future we can combine it with open-telemetry/semantic-conventions#1462 to support extended chars in the value.
 
- Option 2: make `value` optional.

  - default to `id` when not specified
  - enforce back-compat (you can't remove or change value)
  - use `id` to create attribute name when doing (`db.system=foo` and `db.foo.property=bar`)

I'd prefer Option 1 combined with open-telemetry/semantic-conventions#1462.  Both options result in the same outcome, but with Option 1 we start from being strict and relax it once open-telemetry/semantic-conventions#1462 comes.

nanookclaw:
## What changed

Enum member deserialization now accepts entries without an explicit `value` field and fills the public `EnumEntriesSpec.value` with the member `id`. Explicit `value` fields are still preserved.

This keeps downstream code working with a concrete `ValueSpec` while matching the semantic convention schema behavior described in #733.

Closes #733.

## Verification

- `cargo test -p weaver_semconv test_enum_member_value_defaults_to_id`
- `cargo test -p weaver_semconv`

### DIFF
--- a/crates/weaver_semconv/src/attribute.rs
+++ b/crates/weaver_semconv/src/attribute.rs
@@ -418,19 +418,34 @@ pub struct EnumEntriesSpec {
     pub annotations: Option<BTreeMap<String, YamlValue>>,
 }
 
+/// Deserialization helper for [`EnumEntriesSpec`] that allows `value` to be
+/// omitted and defaulted from `id`. The fields mirror [`EnumEntriesSpec`]
+/// (including doc comments) so the generated JSON schema retains field
+/// descriptions when `#[serde(from = "...")]` redirects schema derivation here.
 #[derive(Deserialize, JsonSchema)]
 #[serde(deny_unknown_fields)]
 struct EnumEntriesSpecDeserialize {
+    /// String that uniquely identifies the enum entry.
     id: String,
+    /// String, int, or boolean; value of the enum entry.
+    /// If omitted, defaults to the value of `id`.
     value: Option<ValueSpec>,
+    /// Brief description of the enum entry value.
+    /// It defaults to the value of id.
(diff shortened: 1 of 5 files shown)


jsuereth (approved):
This generally looks good to me, but want to confirm with @lmolkova or other @open-telemetry/specs-semconv-maintainers

lquerel (approved):
LGTM

lmolkova (approved):

PR #1445, opened 2026-05-16
Fix Scorecard pin dependencies issues

jerbly:
There are 5 pin dep issues that this should clear. A couple of them are download-then-run issues so a verify step against the sha is added.

This does add more maintenance on our plates - renovate can't cover the custom sha verification steps.


lquerel (approved):
LGTM

PR #1447, opened 2026-05-16
Add ClusterFuzzLite fuzzing

jerbly:
Nightly continuous fuzzing via ClusterFuzzLite — resolves the OpenSSF Scorecard "Fuzzing" finding.

9 cargo-fuzz targets covering weaver's file-format parsers: JSON/text live-check ingesters, YAML semconv + manifest + forge config, TOML weaver config, Rego policy, jq filter, Jinja template.

Run locally with just fuzz [target] [seconds] or just fuzz-all. Requires nightly Rust + cargo install cargo-fuzz.

In CI, .github/workflows/cflite_batch.yml runs nightly. Crashes surface as Code scanning alerts in the Security tab; corpus + crash inputs are stored as workflow artifacts.

Finding: forge_jq already discovered a reachable panic — the 4-byte input 00.0 unwraps a parse error in jaq-json 1.1.3. Left in place so it exercises the SARIF pipeline on the first nightly run.

Out of scope: OTLP and serve HTTP surfaces live in the binary crate and can't be depended on from fuzz/. This will require a deeper code refactor in another PR.


lquerel (approved):
Nice. I’ll probably take inspiration from this for  OTel-Arrow￼. Thanks!

Issue #879, opened 2025-08-13
Re-usable github action


ArthurSens:
**Is your feature request related to a problem? Please describe.**
I'd love to start using Weaver as part of my CI workflows in a couple of repositories. Unfortunately, not everybody knows Rust and how to build Rust applications. Running containers in CI also isn't the best experience since we need to spin up things like QEMU and it makes the whole workflow slower.

**Describe the solution you'd like**

It would be awesome if Weaver could be offered as a standalone GitHub Action. We probably don't need all the commands available, but `generate` and `validate` would be awesome already.

Some ideas:

```yaml
    steps:
      - uses: opentelemetry/weaver
        with:
          command: validate
          registry: xxx
     - uses: opentelemetry/weaver
        with:
          command: generate
          registry: xxx
```

or 

```yaml
    steps:
      - uses: opentelemetry/weaver-validate
        with:
          registry: xxx
     - uses: opentelemetry/weaver-generate
        with:
          registry: xxx
```

jerbly:
This would be a nice feature. FYI, you don't need to run weaver in a container in CI. In my weaver-example project I run live-check in CI - weaver is installed using our installer. https://github.com/jerbly/weaver-example/blob/main/.github/workflows/release-and-test.yml

kuklyy:
I'm happy to work on it, but I'd like to propose a solution that keeps things simple and follows GitHub Actions best practices.
Instead of creating command-specific actions, I propose implementing a setup action similar to setup-go, setup-node, and setup-python. This action would:

  1. Install the Weaver CLI using our existing cargo-dist installer
  2. Add it to PATH
  3. Cache the binary for faster subsequent runs

 Users would then call Weaver commands directly in their workflow steps, just like they would locally.

  Example Usage

```yaml
  name: Validate Semantic Conventions

  on:
    pull_request:
    push:
      branches: [main]

  jobs:
    validate:
      runs-on: ubuntu-latest
      steps:
        - uses: actions/checkout@v4

        - name: Setup Weaver
          uses: open-telemetry/weaver/.github/actions/setup-weaver@v1
          # Optionally pin version:
          # with:
          #   version: '0.18.0'

        - name: Validate registry
          run: weaver registry check -r ./model -p ./policies --diagnostic-format gh_workflow_command

        - name: Generate artifacts
          run: weaver registry generate -r ./model -t ./templates rust ./output
```

Why This Approach?
Solves the original problem:

[...]

PR #1448, opened 2026-05-17
Add weaver-live-check-{start,stop} composite actions

cijothomas:
Adds two reusable composite actions that wrap the `weaver registry live-check` lifecycle for CI integration, mirroring the `setup-weaver` pattern:

- **`weaver-live-check-start`** — start the OTLP listener in the background, wait for `/health`, expose endpoints as outputs.
- **`weaver-live-check-stop`** — `POST /stop` to retrieve the report, render a step summary, gate the job on `fail-on` (`violation` | `improvement` | `information` | `none`).

Each action has a README and is exercised by a new `test-weaver-live-check-actions.yml` workflow (parity with `test-setup-weaver-action.yml`). Linux runners only in v1.

Closes the live-check half of #879 (which produced `setup-weaver`).

### End-to-end prototypes

I wired the actions up in two SDK repos as throw-away prototypes, purely to validate the flow end-to-end. These are *not* PRs to those repos — just personal forks running the actions against real instrumentation telemetry:

| Repo (fork) | Subject | Run |
|---|---|---|
| opentelemetry-rust-contrib | tower middleware | [green](https://github.com/cijothomas/opentelemetry-rust-contrib/actions/runs/26458898928) |
| opentelemetry-dotnet-contrib | AspNetCore | [green](https://github.com/cijothomas/opentelemetry-dotnet-contrib/actions/runs/26458900647) |

The prototypes confirmed the actions work uniformly across two languages and SDK shapes; each consumer workflow ended up ~130 lines, of which only ~50 are project-specific (build, run, drive traffic).

### Examples (follow-up)

Per discussion with @jsuereth, @lmolkova, and @jerbly, examples will be added post-merge to the [opentelemetry-weaver-examples](https://github.com/open-telemetry/opentelemetry-weaver-examples) repo (consistent with how `setup-weaver` examples are hosted there today). A follow-up PR will add a `live_check_ci/` example pinning these actions to a tagged weaver ref once this PR lands.


cijothomas:
otel-arrow just started using weaver live-check in CI: open-telemetry/otel-arrow#2986. I think that is among the first
 usage of weaver live check in CI.
 
 Once the actions proposed in this PR land, those workflows can be modified to leverage this and reduce boilerplates and ensure better consistency.

cijothomas:
Marking ready for review. cc @jsuereth @lmolkova @jerbly — per our offline discussion, examples will follow in a separate PR to `opentelemetry-weaver-examples` once this lands. Happy to address feedback in either action.

.github/actions/weaver-live-check-start/action.yml
```diff
@@ -0,0 +1,190 @@
...
+      Seconds the listener will stay idle (no telemetry) before exiting on
+      its own. A safety net; the stop action normally stops it sooner.
+    required: false
+    default: '120'
+  diagnostic-format:
+    description: >-
+      Format passed to `weaver registry live-check --diagnostic-format`.
+      `gh_workflow_command` surfaces findings as inline GitHub Actions
+      annotations.
+    required: false
```
jerbly, line 56: It might be worth clarifying in this description that diagnostics are not the live-check findings, but other problems found during the process, e.g. warnings during registry loading.

.github/actions/weaver-live-check-stop/action.yml
```diff
@@ -0,0 +1,128 @@
...
+  on semantic-convention findings. Does more than just stop — also
+  collects, summarizes, and grades the report. Pair with
+  `weaver-live-check-start`. Linux runners only (v1).
+author: 'OpenTelemetry'
+
+branding:
+  icon: 'check-square'
+  color: 'blue'
+
+inputs:
```
jerbly, line 14: In the stats in the report there is a `coverage` output. It would be good to offer a failure condition for that too. This allows you to check that your registry is being fully exercised by your tests.

.github/actions/weaver-live-check-start/action.yml
```diff
@@ -0,0 +1,190 @@
...
+  in the background and expose its endpoints as outputs. Pair with the
+  `weaver-live-check-stop` action after driving telemetry through it.
+  Linux runners only (v1).
+author: 'OpenTelemetry'
+
+branding:
+  icon: 'play'
+  color: 'blue'
+
+inputs:
```
jerbly, line 13: Could we also have `--config` here? This allows you to give weaver a toml file of the weaver-config schema. Within that you can configure all the cli parameters and add extra items like:
```
[[live-check.finding_filters]]
exclude_samples = [
     "service.name",
     "telemetry.sdk.language",
     "telemetry.sdk.name",
     "telemetry.sdk.version",
]
```

.github/actions/weaver-live-check-stop/parse-report.py
```diff
@@ -0,0 +1,123 @@
...
+        "",
+        "| Level | Count |",
+        "| --- | ---: |",
+        f"| violation | {violations} |",
+        f"| improvement | {improvements} |",
+        f"| information | {informations} |",
+        "",
+    ]
+    if msg_counts:
+        summary_lines += [
```
jerbly, line 86: The summary is nice but it needs more context info to be useful. e.g. the summary produced here: https://github.com/open-telemetry/weaver/actions/runs/26181926973 

Is there a way to get the full report? As a user my next step is going to be to dig into exactly what failed.

(Maybe weaver needs to have a mode where it produces the text output for the log AND json on http??)

.github/actions/weaver-live-check-stop/action.yml
```diff
@@ -0,0 +1,128 @@
...
+  fail-on:
+    description: >-
+      Lowest finding level that should fail the job: `violation` (default,
+      strict CI gate), `improvement` (also fail on improvement-level
+      advisories), `information` (also fail on info-level), or `none`
+      (never fail; report-only). Start with `none` when first adopting,
+      then tighten once existing findings are addressed.
+    required: false
+    default: 'violation'
```
jerbly, line 23: Perhaps this should be lifted up into weaver. live-check could return exit code 1 for the failure level selected, not just `violation` which is the default.

jerbly (commented):
This is really nice, thank you!

I have added a few comments, mostly for future improvements. I do think we need a way to see more detail when the run fails though. Let's improve that part a bit and then this will be a great first version.

.github/actions/weaver-live-check-start/action.yml
cijothomas: make sense. clarifying it now.

.github/actions/weaver-live-check-start/action.yml
cijothomas: agree

.github/actions/weaver-live-check-stop/parse-report.py
cijothomas: Good point. The full json report is uploaded as an artifact now. Summary also slightly improved (and points to the full report)

.github/actions/weaver-live-check-stop/action.yml
cijothomas: For this PR, I modified the script to print the `coverage`. Failure based on coverage - I think its better to add in follow up to keep scope minimal in this PR.

.github/actions/weaver-live-check-stop/action.yml
cijothomas: Great idea! 
I'll file an issue to track adding this to weaver itself, and then leverage it here.

.github/actions/weaver-live-check-stop/action.yml
cijothomas: https://github.com/cijothomas/opentelemetry-rust-contrib/actions/runs/26458898928 tried this in OTel Rust instrumentation. Looks better!

cijothomas: I'll open an issue to better track coverage, right from weaver tool itself. We need it namespaced or something, else it'll show just 1-2% for most standalone instrumentations!

.github/actions/weaver-live-check-stop/action.yml
cijothomas: https://github.com/open-telemetry/weaver/issues/1473 opened.

.github/actions/weaver-live-check-stop/action.yml
cijothomas: https://github.com/open-telemetry/weaver/issues/1474 opened this for tracking improvements to this.

cijothomas:
@jerbly Thanks for the feedback, addressed them, and this is ready for another round of review. Also opened couple of issues for follow up things separately.

.github/actions/weaver-live-check-stop/parse-report.py
```diff
@@ -0,0 +1,139 @@
...
+
+    stats = report.get("statistics") or {}
+    level_counts = stats.get("advice_level_counts") or {}
+    highest_counts = stats.get("highest_advice_level_counts") or {}
+    violations = int(level_counts.get("violation", 0))
+    improvements = int(level_counts.get("improvement", 0))
+    informations = int(level_counts.get("information", 0))
+    samples = len(report.get("samples") or [])
+    total = stats.get("total_entities", 0) or 0
+    by_type = stats.get("total_entities_by_type") or {}
```
jerbly: There's an unfortunate naming clash here which I haven't fixed yet. `total_entities` is the total number of samples and likewise for by_type. I made this before I knew about the `entity` type in semantic conventions :( ... to make this clear for now you could remove your `samples` count and use `total_entities` as `samples` instead. The main fix here though is in the labels in the output below.

.github/actions/weaver-live-check-stop/parse-report.py
```diff
@@ -0,0 +1,139 @@
...
+        status = "PASS"
+
+    summary_lines: list[str] = [
+        f"## Weaver live-check: {status}",
+        "",
+        f"- Samples received: **{samples}**",
+    ]
+    if by_type:
+        parts = ", ".join(f"{v} {k}" for k, v in sorted(by_type.items()))
+        summary_lines.append(f"- Entities checked: **{total}** ({parts})")
```
jerbly:
```suggestion
        summary_lines.append(f"- Samples checked: **{total}** ({parts})")
```

.github/actions/weaver-live-check-stop/parse-report.py
```diff
@@ -0,0 +1,139 @@
...
+    if coverage is not None:
+        try:
+            summary_lines.append(f"- Registry coverage: **{float(coverage) * 100:.1f}%**")
+        except (TypeError, ValueError):
+            pass
+    summary_lines += [
+        "",
+        "| Level | Findings | Worst-level entities |",
```
jerbly:
```suggestion
        "| Level | Findings | Worst-level samples |",
```

jerbly (commented):
Let's just fix the confusion caused by `samples` versus `entities` and then this is good to go.

.github/actions/weaver-live-check-stop/action.yml
lquerel: This is really cool. Thanks Cijo for these Weaver integrations!

lquerel (approved):
LGTM. Thanks Cijo.

cijothomas:
@jerbly Could you re-review?

jerbly (approved):
LGTM

PR #1463, opened 2026-05-22
Fix Flaky test port reservation

jerbly:
We have flaky tests due to port collisions despite using port picker. There are race hazards due to parallel runs.

Solved with our own implementation of port reservation which also separates port selection for multiple binaries running at the same time.


lquerel (approved):
LGTM

PR #1482, opened 2026-06-03
fix(live-check-stop): group step-summary findings by severity

cijothomas:
Current summary sorts findings by count, so a 1x violation can sit below 99x informational rows — the actual reason CI failed is buried.

Group advisories by `level` and render one table per severity (🚫 Violations, ⚠️ Improvements, ℹ️ Information), each top 15 by count. No input/output/exit-code changes. Falls back to the old count-sorted view if `samples` is missing.

Follow-up to #1448.

### DIFF
--- a/.github/actions/weaver-live-check-stop/README.md
+++ b/.github/actions/weaver-live-check-stop/README.md
@@ -42,7 +42,9 @@ Linux runners only in v1.
 - Parses the report with `parse-report.py` (a Python script bundled
   alongside this action) and writes a markdown summary to
   `$GITHUB_STEP_SUMMARY` — including counts per level, registry
-  coverage, and a top-findings table.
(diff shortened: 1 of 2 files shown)


jerbly (approved):
Nice... Given your experience making these actions and the python script, there's probably more improvements we can make upstream in weaver to simplify this.

cijothomas:
> Nice... Given your experience making these actions and the python script, there's probably more improvements we can make upstream in weaver to simplify this.

Totally! I am incorporating this to OTel Arrow and OTel Rust repos in parallel, and once done, I'll have a solid idea of what to do inside weaver itself vs the GH Action vs custom in each repo.

(My end goal is to have weaver live check *enforced* in **every** OTel language repo for the sdk internal metrics. All of this is moving one step closer to that goal.)

PR #1485, opened 2026-06-04
Prevent the fuzzing found panic in minijinja marking all PRs with a red cross

jerbly:
The fuzzer consistently finds a panic in minijinja (an i64 to f64 comparison). Issue raised: https://github.com/mitsuhiko/minijinja/issues/904

To prevent seeing red crosses for every PR, this change will not fuzz minijinja temporarily at PR time until there's a fix. We will still fuzz minijinja in the nightly batch fuzzer.

Also, included a fix to batch mode to hopefully avoid the config error reported in the security tab.

(I had also accidentally excluded the Cargo.lock - this PR adds it back)


lquerel (approved):
LGTM

PR #1504, opened 2026-06-11
Refactor weaver_resolver to use a cache

jsuereth:
Weaver resolver becomes a struct that remembers/caches previously loaded `schema_url`s

- New WeaverResolver struct that has cache of resolved schemas by schema url
- update all code in weaver to use this struct
- Create "visitor" pattern mechanism to allow policy + resolver to not have direct dependnecies on each other but allow the "weaver" type to call policies on before_resolution still
- Configuration where we can specifying `schema_url` -> `vdir` mappings via config toml, allowing you to resolve by schema_url, but create proxies, e.g.

This does *not* leverage any of the advanced new capabilities yet.  This just adds in the core interface and features.  We can expand live-check and CLI independently from here.

Had to fix a a flaky issue with the port-grabbing lib.


crates/weaver_resolver/src/lib.rs
```diff
@@ -138,450 +302,555 @@ impl SchemaResolver {
...
+            match RegistryRepo::try_new_dependency_with_auth(&dep, &mut nfes, &self.config.auth) {
+                Ok(r) => r,
+                Err(e) => return WResult::FatalErr(Error::FailToResolveDefinition(e)),
+            }
+        };
+
+        let res = self.load_and_resolve_schema(repo, DefaultSchemaVisitor);
+        match res {
+            WResult::Ok(_) => {
+                let arc = self.cache.get(schema_url).expect("Just cached").clone();
```
jerbly: I think this will panic if the override target resolves to a different schema_url than the override key

crates/weaver_test_support/src/lib.rs
```diff
@@ -4,37 +4,99 @@
...
+struct PortLock {
+    lock_path: std::path::PathBuf,
+}
+
+impl PortLock {
+    fn acquire() -> Self {
+        let temp_dir = std::env::temp_dir();
+        let lock_path = temp_dir.join("weaver_test_port_allocator.lock");
+
+        loop {
```
jerbly, line 29: This can loop forever: an orphaned lock file (left by a hard-killed process) is never reclaimed, so every retry hits AlreadyExists.

crates/weaver_resolver/src/lib.rs
```diff
@@ -21,75 +27,233 @@ mod loader;
...
+            LoadedSemconvRegistry::Resolved(s) => {
+                match SchemaUrl::try_from(s.schema_url.as_str()) {
+                    Ok(url) => url,
+                    Err(_) => return WResult::FatalErr(Error::FailToResolveSchemaUrl {}),
+                }
+            }
+            LoadedSemconvRegistry::ResolvedV2(s) => s.schema_url.clone(),
+        };
+
+        if let Some(cached) = self.cache.get(&schema_url) {
```
lmolkova, line 203: it does not look like we normalize schema - minor differences like `/` at the end would result in different schemas. Looks like an existing bug / follow-up

crates/weaver_resolver/src/lib.rs
```diff
@@ -21,75 +27,233 @@ mod loader;
...
+    pub fn new(config: WeaverResolverConfig) -> Self {
+        Self {
+            cache: LruCache::new(config.cache_capacity),
+            config,
+        }
+    }
+
+    /// Configures an explicit location override for a specific SchemaUrl.
+    /// When Weaver resolves this SchemaUrl, it will redirect the fetch to the target VirtualDirectoryPath.
+    pub fn add_schema_url_override(
```
lmolkova, line 152: should we replace dependency.registry_path with the config way?

crates/weaver_resolver/src/lib.rs
```diff
@@ -21,75 +27,233 @@ mod loader;
...
+        if let Some(cached) = self.cache.get(&schema_url) {
+            match &**cached {
+                WeaverResolvedSchema::V1(s) => return WResult::Ok(s.clone().into()),
+                WeaverResolvedSchema::V2(s) => return WResult::Ok(s.clone().into()),
             }
-            LoadedSemconvRegistry::ResolvedV2(_) => {
-                todo!("Converting V2 schema back into V1 is unsupported")
+        }
+
+        let loaded = if let Some(override_path) = self.config.schema_url_overrides.get(&schema_url)
```
lmolkova, line 210: what happens if if I have dependency with registry_path and overrides configured at the same time?

lmolkova (commented):
Nice refactoring!

crates/weaver_test_support/src/lib.rs
jsuereth: Good point - WDYT on next steps - should I have this wait for N seconds before killing the lock and progressing?

Note: This consistently fails on my dev machine without the fix

crates/weaver_resolver/src/lib.rs
jsuereth: Fair - I don't think that can actually happen, but I'll be more defensive here.

crates/weaver_resolver/src/lib.rs
jsuereth: I can try to tackle that now if you'd like.

crates/weaver_resolver/src/lib.rs
jsuereth: This is here so we can add that in a follow-up PR.

crates/weaver_resolver/src/lib.rs
jsuereth: I wanted override path to win, always.  I can verify that's the case - but curious if you have a different suggestion on the goal there.

crates/weaver_resolver/src/lib.rs
jsuereth: This now issues an error when you try to resolve a `schema_url` but get a *different* `schema_url` from the location we resolve.

crates/weaver_test_support/src/lib.rs
jsuereth: Fixed this to have hard timeout

crates/weaver_resolver/src/lib.rs
jsuereth: Added minimum of normalization

lmolkova (approved):

PR #1515, opened 2026-06-17
Replace metric requirement level with generic (for all signals) one

### LINKED ISSUE #1484: Requirement levels for all signals
We just added `metric_requirement_level` in the https://github.com/open-telemetry/weaver/pull/1381, but the problem is generic across signals:

- Instrumentation can emit some required spans and some less important ones (e.g. logical operation is required and internal physical requests are optional)
- Events might also be optional and required regardless of severity. 
- A similar argument can also be made for entities and events. Some specific detector must report certain entities but others are not that important / too verbose

The requirement level is a guidance for instrumentation authors on 
- what must be implemented to make instrumentation generally useful
- what should be opt-in

When doing pure live-check (without knowing which set of instrumentations is being tested), it's not really possible to enforce requirement levels, but instrumentation-specific tests can add extra validation based on live-check report to validate that specific signal was / was not emitted.

lmolkova:
Fix #1484

We didn't release weaver after metric_requirement_level was introduced, so it's ok to rename it without any back-compat.

This is only part of v2 definition / resolved / forge schema now.


lquerel (approved):
LGTM

jsuereth (approved):

jsuereth:
This needs docs regenerated

PR #1517, opened 2026-06-18
feat(live-check): add --fail-on severity gate

### LINKED ISSUE #1473: live-check: --fail-on flag for severity-based exit code
Today `weaver registry live-check` only exits non-zero on `violation`-level findings. Every CI consumer that wants a different threshold (improvement, information, none) ends up re-implementing the gating in a wrapper — e.g. the composite actions added in #1448 do this in a Python helper.

Add a `--fail-on <violation|improvement|information|none>` flag (and matching `.weaver.toml` field) so the CLI returns the right exit code itself.

Discussion: https://github.com/open-telemetry/weaver/pull/1448#discussion_r3287761501

cijothomas:
Closes #1473.

Adds `--fail-on <violation|improvement|information|none>` to `weaver registry live-check`. Default is `violation`, preserving prior behavior.

## Usage

CLI:
```sh
weaver registry live-check -r ./model --fail-on improvement
```

Configuration file (`.weaver.toml`):
```toml
[live_check]
fail_on = "improvement"
```

CLI flags override `.weaver.toml` values.

## Levels

| `--fail-on` | Exits non-zero when findings include |
|---|---|
| `violation` (default) | violations |
| `improvement` | improvements or violations |
| `information` | any finding |
| `none` | never (always exits 0) |

`--no-stats` still exits 0 as before (statistics are unavailable, so the severity gate cannot be evaluated); a warning is logged to make the no-op visible.


src/registry/live_check.rs
```diff
@@ -229,6 +236,17 @@ pub(crate) fn command(
     let registry_args = cmd_config.registry;
     let policy_args = cmd_config.policy;
 
+    // --no-stats disables the statistics accumulator, so --fail-on cannot be
+    // evaluated. Warn the user so the silent no-op is visible.
+    if config.no_stats && config.fail_on != FailOnLevel::None {
+        log_warn(format!(
+            "--no-stats disables statistics; --fail-on={} cannot be enforced. \
```
lmolkova, line 248: is this right that these two are related? 

In my mind --no-stats controls if stats are rendered in the output, not that they are not calculated

crates/weaver_config/src/live_check.rs
```diff
@@ -38,6 +102,10 @@ pub struct LiveCheckConfig {
     /// Disable statistics accumulation. Useful for long-running live-check sessions.
     pub no_stats: bool,
 
+    /// Severity threshold that causes a non-zero exit code. Findings at this
+    /// level or higher fail the run. Use `none` to never fail.
+    pub fail_on: FailOnLevel,
```
lmolkova, line 107: not blocking: throwing in some naming alternatives that sound a bit more clear to me

--fail-level
--fail-severity
--min-severity

src/registry/live_check.rs
cijothomas: I see. I was assuming no-stats means we dont calculate the stats at all, so no way to apply fail-on! I'll recheck this and get back.

jerbly: Currently --no-stats stops them from being collected. This was by design to save some processing time. However, we could change the internals so it just captures the highest level so that --fail-on will work... I think I left a TODO in the code there about this.

crates/weaver_config/src/live_check.rs
cijothomas: `--fail-on` follows [pylint's naming](https://pylint.readthedocs.io/en/latest/user_guide/configuration/all-options.html) for the same concept. Reads naturally: `--fail-on violation`.

src/registry/live_check.rs
cijothomas: Agreed — tracking just the highest severity seen (a single enum) would make `--fail-on` work independently of `--no-stats` at negligible cost. Will address in a follow-up PR to keep this one focused.

crates/weaver_config/src/live_check.rs
cijothomas: Resolving, as I think the current name can be used as-is. If any strong pushback, will revisit.

src/registry/live_check.rs
cijothomas: Resolving, as this would be in follow up PR, not this one. I can create a tracking issue

crates/weaver_live_check/src/stats.rs
```diff
@@ -211,6 +211,18 @@ impl CumulativeStatistics {
             .contains_key(&FindingLevel::Violation)
     }
 
+    /// Returns the most severe level currently recorded, or `None` if no
+    /// findings have been accumulated. Severity ordering is defined by
+    /// [`FindingLevel`]'s derived `Ord`: Information < Improvement < Violation.
+    pub(crate) fn worst_level(&self) -> Option<FindingLevel> {
```
lmolkova: maybe

```suggestion
    pub(crate) fn max_level(&self) -> Option<FindingLevel> {
```

?

crates/weaver_live_check/src/stats.rs
```diff
@@ -211,6 +211,18 @@ impl CumulativeStatistics {
...
+        self.highest_advice_level_counts.keys().copied().max()
+    }
+
+    /// Returns true if any recorded finding is at or above `threshold`.
+    pub(crate) fn fails_at(&self, threshold: FindingLevel) -> bool {
```
lmolkova: nit, can callers compare levels themselves? 

The name is a bit confusing, I'd suggest `should_fail` or something descriptive

src/registry/live_check.rs
```diff
@@ -234,6 +241,17 @@ pub(crate) fn command(
     let registry_args = cmd_config.registry;
     let policy_args = cmd_config.policy;
 
+    // --no-stats disables the statistics accumulator, so --fail-on cannot be
+    // evaluated. Warn the user so the silent no-op is visible.
+    if config.no_stats && config.fail_on != FailOnLevel::None {
+        log_warn(format!(
```
lmolkova, line 247: is this still needed?

src/registry/live_check.rs
```diff
@@ -394,9 +412,13 @@ pub(crate) fn command(
...
 
     stats.finalize();
-    // Set the exit_code to a non-zero code if there are any violations
-    if stats.has_violations() {
-        exit_code = 1;
+    // Set exit_code based on the configured --fail-on threshold. `None`
+    // threshold means "never fail". `fails_at` returns false for disabled
+    // stats; the startup check above rejects --no-stats + non-`none` gates.
+    if let Some(threshold) = config.fail_on.as_finding_threshold() {
+        if stats.fails_at(threshold) {
```
lmolkova: To the prev comment on naming:

```suggestion
        if stats.should_fail(threshold) {
```

or  `stats.max_level < threshold`  (with rust dances)

would be more readable, no?

lmolkova (commented):
Looks good, just some cosmetic comments

src/registry/live_check.rs
cijothomas: yes. https://github.com/open-telemetry/weaver/pull/1517#discussion_r3522405941

src/registry/live_check.rs
```diff
@@ -234,6 +241,17 @@ pub(crate) fn command(
     let registry_args = cmd_config.registry;
     let policy_args = cmd_config.policy;
 
+    // --no-stats disables the statistics accumulator, so --fail-on cannot be
+    // evaluated. Warn the user so the silent no-op is visible.
+    if config.no_stats && config.fail_on != FailOnLevel::None {
```
jerbly, line 246: Since the default is `violation` this warning will always show now for anyone using `--no-stats` even if they don't care about the exit code.

However, this will go away in a follow up PR when we make `--no-stats` capture the highest finding level for a run. We need to do that before we release IMO.

crates/weaver_live_check/src/stats.rs
```diff
@@ -326,6 +338,18 @@ impl LiveCheckStatistics {
         }
     }
 
+    /// Returns true if any recorded finding is at or above `threshold`.
+    /// Always returns `false` for [`LiveCheckStatistics::Disabled`]; the
+    /// caller is responsible for warning about `--no-stats` + a
+    /// stats-dependent gate at startup.
+    #[must_use]
+    pub fn should_fail(&self, threshold: FindingLevel) -> bool {
```
jerbly, line 331: I think this function now makes `has_violations()` redundant. We should drop it.

crates/weaver_config/src/live_check.rs
```diff
@@ -324,4 +395,76 @@ min_level = "violation"
...
+        assert_eq!(
+            FailOnLevel::from_str("improvement").unwrap(),
+            FailOnLevel::Improvement
+        );
+        assert_eq!(
+            FailOnLevel::from_str("information").unwrap(),
+            FailOnLevel::Information
+        );
+        assert_eq!(FailOnLevel::from_str("none").unwrap(), FailOnLevel::None);
+        let err = FailOnLevel::from_str("bogus").unwrap_err();
```
jerbly: Nit: We're trying not to use `unwrap()` in tests, `expect()` is preferred to give some context.

jerbly (approved):
Looks good. Just a couple clean-ups / nits.

crates/weaver_live_check/src/stats.rs
cijothomas: agree

crates/weaver_config/src/live_check.rs
cijothomas: done in latest commit.

cijothomas:
https://github.com/open-telemetry/weaver/issues/1577 is opened to fix the no-stats issue, will get it after this is merged.

src/registry/live_check.rs
cijothomas: Agrees. https://github.com/open-telemetry/weaver/issues/1577 tracking.

jerbly (approved):

Issue #1362, opened 2026-04-15
live-check: custom policies


lmolkova:
1. Custom policies in rego replace backed-in ones, we need to have ways to keep both
2. maybe related: allow to add json to policies dir (it should be translated to `data`)

lmolkova:
WRT

> Custom policies in rego replace backed-in ones, we need to have ways to keep both

What if we did this:

1. Have custom policies to be composable (maybe we already do)
2. Have shared ones published on https://github.com/open-telemetry/opentelemetry-weaver-packages
3. Let users pick a set they want to run

Maybe also we should revisit what we want to validate at live-check time in shared policies?

The [current one](https://github.com/open-telemetry/weaver/tree/main/defaults/policies/live_check_advice) covers naming pattern - this seems to be `weaver check` responsibility.

Some ideas for shared (but not required for everyone) policies:
- schema_url presence 
- schema_url reachability 
- presence of `service*` entities

jerbly:
Regarding composable policies. We had an idea somewhere (I can't find it) of being able to reference builtin policies (and other things) with a URI like: `builtin://otel.rego` - then you could just include it alongside any files or remote files as just another source.

Regarding "The [current one](https://github.com/open-telemetry/weaver/tree/main/defaults/policies/live_check_advice) covers naming pattern - this seems to be `weaver check` responsibility." -- If you run `live-check` against your OTLP output or via MCP scraped from your code, or by pulling dataset columns from your observability back-end, you might get a bunch of things that aren't in your model. It's useful to know right there that those things are going to need work if you want to model them properly. So, this is a use-case for having this builtin optional - e.g. in your model-driven case you would not require it.

PR #1539, opened 2026-06-25
Live-check custom policies: support additional data 

lmolkova:
Related to #1362

It fixes the  
> 2. maybe related: allow to add json to policies dir (it should be translated to data)

Allows to provide additional data (via --advice-data or weaver toml) to live check policies. They are loaded into `data` and available for custom policies. 

E.g. we use it to validate complex attributes against schema in genai. This PR would allow to remove most of the hacks in https://github.com/open-telemetry/opentelemetry-python-genai/blob/27d84260b43c972bad41f615ca43cdd3f44a6f8e/util/opentelemetry-test-util-genai/src/opentelemetry/test_util_genai/_setup_weaver.py#L237

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -445,7 +445,7 @@ jobs:
       - name: Install cargo-tarpaulin
         run: cargo install cargo-tarpaulin
       - name: Gather coverage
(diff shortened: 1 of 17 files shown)


jerbly (approved):
Nice! We should also fix part 1 of https://github.com/open-telemetry/weaver/issues/1362 before we release the next version of weaver.... another PR.

PR #1561, opened 2026-07-03
Make weaver.yaml more customizeable via toml and params

lmolkova:
To support reusable templates here https://github.com/open-telemetry/opentelemetry-weaver-packages/pull/38, we need to make templates + weaver.yaml more customizable.

Several things that should be configurable by semconv repos:
1. acronyms
2. text maps
3. set of sections to generate (e.g. metric / span /event registries are not rendered today in semconv or semconv-genai and I might even want to disable attribute registry in genai conventions)
4. params (supported via CLI, not yet via toml), JQ helper options can all be configured through params, no change in this PR: 
    - v2 - not configurable - templates are  
    - exclude_deprecated 
    - excluded_namespaces

This PR
- adds acronyms and text maps to weaver.toml and merges them between toml and yaml giving toml priority (yaml is reusable, toml is project-specific)
- adds `when` clause into weaver yaml `template` section that contains JQ expression that controls if this template is on. 

E.g.

```yaml
# weaver.yaml
  - template: span_namespace.md.j2
    when: $generate_span_registry
    filter: |-
      semconv_grouped_spans({v2: true, exclude_deprecated: $exclude_deprecated, stable_only: $stable_only, exclude_root_namespace: $exclude_root_namespace})
    application_mode: each
    file_name: "{{ ctx.root_namespace | kebab_case }}/spans.md"
```

Would only be triggered when  `$generate_span_registry` is true

and

```yaml
  - template: registry_readme.md.j2
    when: $generate_attribute_registry or $generate_entity_registry or $generate_span_registry or $generate_metric_registry or $generate_event_registry

```

would be triggered when any of `generate_*_registry` params is true.

Reusing JQ conditions in `when` to match `filter`.

https://github.com/open-telemetry/opentelemetry-weaver-packages/pull/38 is demonstrating this version and https://github.com/open-telemetry/semantic-conventions-genai/compare/main...lmolkova:semantic-conventions-genai:reusable-templates-demo shows e2e.


jerbly:
I haven't had a detailed look at this yet but how does it fit in with the spec we agreed on here: https://github.com/open-telemetry/weaver/blob/main/docs/specs/package-vs-project-config/config-proposal.md ?

I partially completed phase 2 in the implementation plan. But did not tackle things to do with templates, this bullet in the doc for example:
> - Template defaults — shared settings like acronyms, template_syntax, whitespace_control, and params that apply across all template packages used by the project, overriding the package's own weaver_template.yaml defaults

Phase 3 renames the weaver.yaml to weaver_template.yaml to disambiguate too. Perhaps this PR covers the missing parts of Phase 2 and should include Phase 3?

crates/weaver_forge/src/lib.rs
```diff
@@ -451,6 +456,31 @@ impl TemplateEngine {
...
+        template: &TemplateConfig,
+        context: &serde_json::Value,
+        params: &BTreeMap<String, serde_json::Value>,
+    ) -> Result<bool, Error> {
+        let Some(when) = template.when.as_deref() else {
+            return Ok(true);
+        };
+
+        let result = Filter::new(when).apply(context.clone(), params)?;
+        Ok(!matches!(
```
jerbly: Question on the truthiness here: since `execute_jq` only returns a scalar for a single-output stream and otherwise gives back an array (empty `[]` for zero outputs), does this mean a `when` that produces an empty stream — e.g. `.groups[] | select(...)` matching nothing — comes back as `[]` and gets treated as truthy, so the template renders anyway? That'd be the opposite of the `filter` path just below, which treats an empty array as "nothing." Wondering if `when` should do the same, or only accept a single boolean?

crates/weaver_forge/src/config.rs
```diff
@@ -669,6 +676,44 @@ impl WeaverConfig {
...
+    ///
+    /// On a name conflict the incoming (toml) map replaces the yaml.
+    pub fn merge_text_maps(&mut self, text_maps: Option<HashMap<String, HashMap<String, String>>>) {
+        let Some(incoming) = text_maps else {
+            return;
+        };
+        let merged = self.text_maps.get_or_insert_with(HashMap::new);
+        for (name, map) in incoming {
+            // Top-level override: replace the whole named map, don't merge values.
+            let _ = merged.insert(name, map);
```
jerbly: Just to check: when a project `.weaver.toml` sets only a couple of keys under a named map, does `insert` replace the package's whole map and drop its other entries? Since `merge_acronyms` merges per item, is the difference intended, or should this merge per key too?

crates/weaver_forge/src/config.rs
```diff
@@ -669,6 +676,44 @@ impl WeaverConfig {
...
+    /// Merge additional `text_maps` from a higher-precedence source, such as the
+    /// project-level `.weaver.toml` `[template]` section, into the template
+    /// package's own `text_maps` (from `weaver.yaml`).
+    ///
+    /// On a name conflict the incoming (toml) map replaces the yaml.
+    pub fn merge_text_maps(&mut self, text_maps: Option<HashMap<String, HashMap<String, String>>>) {
+        let Some(incoming) = text_maps else {
+            return;
+        };
+        let merged = self.text_maps.get_or_insert_with(HashMap::new);
```
jerbly: Nit: could this loop just be `merged.extend(incoming)`? Same behavior, one line.

crates/weaver_forge/src/config.rs
```diff
@@ -669,6 +676,44 @@ impl WeaverConfig {
...
+    /// On a case-insensitive collision the toml file wins.
+    pub fn merge_acronyms(&mut self, acronyms: Option<Vec<String>>) {
+        let Some(incoming) = acronyms else {
+            return;
+        };
+        let mut merged = self.acronyms.take().unwrap_or_default();
+        for acronym in incoming {
+            match merged
+                .iter_mut()
+                .find(|existing| existing.to_lowercase() == acronym.to_lowercase())
```
jerbly: Nit: `acronym.to_lowercase()` runs for every existing entry inside the `find`. Could we compute it once per incoming acronym instead of once per comparison?

crates/weaver_forge/src/lib.rs
```diff
@@ -451,6 +456,31 @@ impl TemplateEngine {
...
+        &self,
+        template: &TemplateConfig,
+        context: &serde_json::Value,
+        params: &BTreeMap<String, serde_json::Value>,
+    ) -> Result<bool, Error> {
+        let Some(when) = template.when.as_deref() else {
+            return Ok(true);
+        };
+
+        let result = Filter::new(when).apply(context.clone(), params)?;
```
jerbly: Nit: since `when` only depends on the registry context and params (not the file), does this clone the whole context and run an extra JQ pass once per matching file? Could it run once per template, or be part of the existing filter pass, so it doesn't repeat on large registries?

crates/weaver_forge/src/config.rs
lmolkova: acronyms are an array, they are unioned, the only conflict is casing
text_maps are more complicated. The toml value replaces yaml value if they have the same keys. It's hard to tell what would consistency mean here. I'd argue that replacing value for conflicting key is the same beahvior for both. 

Most importantly, I don't see a good use case for merging pairs inside map, but it adds a bit of complexity. So I'd rather not add it until someone actually wants it.

crates/weaver_forge/src/lib.rs
lmolkova: great point, added a requirement for `when` expression to evaluate to a boolean and hard-fail when it doesn't

crates/weaver_forge/src/lib.rs
lmolkova: I optimized it a bit (by not cloning the context). we're still evaluating another jq filter independently.

A couple of reasons:
- folding into `filter` is possible, but then it easier just to ask users to make condition part of the `filter` itself. Not a bad alternative, just hard to do right. If we still provide `when`, we'd just make it harder (and come back to truthy problem among others)
- there are usually just a few (even if it's 100) templates per registry and docs/code generation is never on a hot path. 
- We can optimize execute_jq by caching jaq Ctx across all templates (all of them share the same registry and params). Given context is huge and JQ programs are small (usually), re-compiling JQ would probably be less expensive. Created https://github.com/open-telemetry/weaver/issues/1565 to track it

lmolkova:
> I haven't had a detailed look at this yet but how does it fit in with the spec we agreed on here: https://github.com/open-telemetry/weaver/blob/main/docs/specs/package-vs-project-config/config-proposal.md ?

> Phase 3 renames the weaver.yaml to weaver_template.yaml to disambiguate too. Perhaps this PR covers the missing parts of Phase 2 and should include Phase 3?

The acronyms and maps are based exactly on the proposal with a caveat that I used `[templates]` and not `[template_defaults]` because `defaults` is relative and confusing (weaver.yaml in reusable templates would have defaults, toml effectively overrides them).

It's part of the [phase 2](https://github.com/open-telemetry/weaver/blob/main/docs/specs/package-vs-project-config/config-proposal.md#implementation-phases) and doesn't do anything from the phase 3

jerbly (approved):
Nice! Probably just needs a changelog entry and maybe something in the docs to finish.

PR #1564, opened 2026-07-04
docs: expand GitHub Actions section with live-check actions

cijothomas:
The GitHub Actions section in the README only mentioned `setup-weaver`. This adds the `weaver-live-check-start` and `weaver-live-check-stop` composite actions to the table, and links to real-world CI workflow examples from opentelemetry-rust-contrib and otel-arrow.


jerbly (approved):
This is nice, thanks. Could we add or improve the examples in opentelemetry-weaver-examples? The basic example runs live check inside a test - this is still nice to show one way of using it. But, it would be nice if the CI used the new actions. https://github.com/open-telemetry/opentelemetry-weaver-examples/blob/main/basic/tests/live-check.rs

cijothomas:
> This is nice, thanks. Could we add or improve the examples in opentelemetry-weaver-examples? The basic example runs live check inside a test - this is still nice to show one way of using it. But, it would be nice if the CI used the new actions. https://github.com/open-telemetry/opentelemetry-weaver-examples/blob/main/basic/tests/live-check.rs

I added links to actual OTel repos leveraging this (more repos are onboarding to CI weaver check, so would have more links soon)  - Do you think we should maintain the examples in long term, if we point users to real CI itself, across OTel?

jerbly:
> I added links to actual OTel repos leveraging this (more repos are onboarding to CI weaver check, so would have more links soon) - Do you think we should maintain the examples in long term, if we point users to real CI itself, across OTel?

It's fantastic to have real use cases, but I think it's also important to have a "Hello World" to help disentangle the usage of weaver from the complexity of the codebase it's being used in.

PR #1570, opened 2026-07-06
Add stats dashboard to the UI

jerbly:
Adds a rich stats page to the `serve` UI

`/api/v1/registry/stats` now returns the full `weaver registry stats` output, rendered as charts:

image

### DIFF
--- a/src/serve/handlers.rs
+++ b/src/serve/handlers.rs
@@ -17,7 +17,7 @@ use weaver_forge::run_filter_raw;
 use crate::serve::types::FilterParams;
 
 use super::server::AppState;
-use super::types::{RegistryCounts, RegistryStats, SearchParams, SearchResponse};
+use super::types::{SearchParams, SearchResponse};
 
 /// Health check.
 #[utoipa::path(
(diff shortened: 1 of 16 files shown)


src/serve/handlers.rs
```diff
@@ -78,30 +78,25 @@ pub async fn get_schema(Path(name): Path<String>) -> impl IntoResponse {
 }
 
 /// Get statistics for the registry.
+///
+/// Returns the full `weaver registry stats` output: counts, and type/stability/
+/// instrument/unit/span-kind breakdowns plus deprecation and documentation coverage.
 #[utoipa::path(
     get,
     path = "/api/v1/registry/stats",
```
jsuereth, line 86: Nit: At some point we'll want to allow multiple registries here :)

I'm not sure how we want our API to look at that point.

src/serve/mod.rs
```diff
@@ -88,8 +88,21 @@ fn run_serve(
...
+    // into the template (forge) representation.
+    let stats = resolved_v2.resolved_schema().stats();
     let forge_registry = resolved_v2.into_template_schema();
 
+    let stats_response = types::RegistryStatsResponse {
+        schema_url: forge_registry.schema_url.to_string(),
+        version: "v2",
+        stats,
+    };
+    let stats_json = serde_json::to_string(&stats_response)
```
jsuereth, line 102: This is ok for now - I think maybe longer term, e.g. when we start using the resolved-registry-cache and can lookup resolved schema by schema_url - we should use that instead of just caching the response.

jsuereth (approved):

src/serve/handlers.rs
jerbly: `serve` will change significantly when we make this move, and the UI too, you'll need to select a registry somehow: provide URLs etc. So I imagine it'll be a breaking change.

PR #1573, opened 2026-07-06
Expand dependency conflict resolution to allow different versions if major-version is compatible

jsuereth:
- Fix forge tests
- Add new tests for dependencies
- Implement very simple "latest version in major-series wins" conflict resolution algorithm


crates/weaver_resolver/src/attribute.rs
```diff
@@ -70,23 +72,134 @@ impl AttributeCatalog {
...
+    pub(crate) fn upgrade_attribute_with_source<C: crate::SchemaCacheLookup>(
+        mut attr_with_source: AttributeWithSource,
+        cache_lookup: &C,
+    ) -> Result<AttributeWithSource, Error> {
+        if let AttributeSource::Dependency { schema_url } = &attr_with_source.source {
+            let reg_name = schema_url.name();
+            if let Some(chosen_url) = cache_lookup.chosen_version(reg_name) {
+                if chosen_url != schema_url {
+                    if let Ok(winning_url) =
+                        UseLatestMajorVersion.resolve_conflict(schema_url, chosen_url)
```
lmolkova: do we need to return error otherwise?

crates/weaver_resolver/src/attribute.rs
```diff
@@ -70,23 +72,134 @@ impl AttributeCatalog {
...
 
+/// Resolves conflicts between two attributes based on provenance.
+fn resolve_conflict(
+    key: &str,
+    m: AttributeWithSource,
+    existing: AttributeWithSource,
+) -> Result<AttributeWithSource, Error> {
+    match (&m.source, &existing.source) {
+        (AttributeSource::Local { .. }, AttributeSource::Local { .. }) => Ok(existing),
+        (AttributeSource::Local { .. }, AttributeSource::Dependency { .. }) => Ok(existing),
```
lmolkova, line 188: is this right that between local and dependency attribute we pick one from dependency?

crates/weaver_resolver/src/attribute.rs
```diff
@@ -70,23 +72,134 @@ impl AttributeCatalog {
...
+    ) -> Result<AttributeWithSource, Error> {
+        if let AttributeSource::Dependency { schema_url } = &attr_with_source.source {
+            let reg_name = schema_url.name();
+            if let Some(chosen_url) = cache_lookup.chosen_version(reg_name) {
+                if chosen_url != schema_url {
+                    if let Ok(winning_url) =
+                        UseLatestMajorVersion.resolve_conflict(schema_url, chosen_url)
+                    {
+                        if winning_url == *chosen_url {
+                            if let Some(upgraded_schema) = cache_lookup.lookup_schema(chosen_url) {
```
lmolkova: do we need any fallback when lookups fail or it's not possible here?

crates/weaver_resolver/src/attribute.rs
```diff
@@ -70,23 +72,134 @@ impl AttributeCatalog {
...
+        }
+        Ok(attr_with_source)
+    }
+
+    /// Returns an attribute ref from the catalog and also registers the attribute
+    /// in `root_attributes` so its provenance is recorded.
+    ///
+    /// Note: This may "upgrade" an attribute reference to use a different version
+    /// when dealing with transitive dependencies and version conflicts.
+    pub fn attribute_ref_with_provenance<C: crate::SchemaCacheLookup>(
```
lmolkova: nit, does it need to be public?

crates/weaver_resolver/src/conflict_strategy.rs
```diff
@@ -0,0 +1,90 @@
...
+        let v1 = url1.semver()?;
+        let v2 = url2.semver()?;
+        if v1.major == v2.major {
+            if v1 > v2 {
+                Ok(url1.clone())
+            } else {
+                Ok(url2.clone())
+            }
+        } else {
+            Err(Error::DuplicateDependency {
```
lmolkova: nit, should it be somethings like VersionMismatchError

crates/weaver_resolver/src/dependency.rs
```diff
@@ -182,48 +194,160 @@ impl ImportableDependency for V1Schema {
...
-                .is_some_and(|name| explicit_entities_matcher.is_match(name.as_str())),
-            GroupType::Scope => false,
-            GroupType::Undefined => false,
+        let matches_explicitly = move |g: &Group| {
+            if g.is_v2 {
+                match g.r#type {
+                    GroupType::AttributeGroup => {
+                        explicit_attribute_groups_matcher.is_match(&g.id)
+                            || g.id
+                                .strip_prefix("registry.")
```
lmolkova, line 203: 😢 

can't wait till we remove v1 compat

crates/weaver_resolver/src/loader.rs
```diff
@@ -177,17 +178,50 @@ pub(crate) fn load_semconv_repository_with_cache(
...
         auth,
-    )
+    );
+
+    if let WResult::Ok(_) | WResult::OkWithNFEs(_, _) = result {
+        let mut warnings = vec![];
+        for (name, main_url) in main_dependencies.iter() {
+            if let Some(selected_url) = chosen_versions.get(name) {
+                if selected_url != main_url {
+                    warnings.push(Error::ConversionError {
```
lmolkova: is this a warning/NFE ? ConversionError seems to be regular error

crates/weaver_resolver/src/lib.rs
```diff
@@ -239,7 +241,92 @@ impl WeaverResolver {
             WResult::FatalErr(e) => WResult::FatalErr(e),
         }
     }
+}
+
+/// Abstraction for querying chosen dependency versions and resolved schemas from cache during resolution.
+pub trait SchemaCacheLookup {
```
lmolkova: nit, does it need to be public?

crates/weaver_resolver/src/registry.rs
```diff
@@ -378,17 +380,35 @@ fn resolve_prefix_on_attributes(ureg: &mut UnresolvedRegistry) -> Result<(), Err
...
+                if let Ok(winning_url) =
+                    UseLatestMajorVersion.resolve_conflict(&prov_url, chosen_url)
+                {
+                    prov_url = winning_url;
+                }
+            }
+        }
+        let provenance = Some(Provenance {
+            schema_url: prov_url.clone(),
+            path: "".to_owned(),
```
lmolkova: can we use the prov.path (when present) ?

jsuereth: Let me think about that one.

We need to erase `path` *prior* to publishing a resolved schema because the paths will be all wrong / uesless.  As such - I was consistently always dropping paths on dependencies as the files may not be accessible (e.g. in `/tmp`) so the error message wasn't better.

If you think it's useful I can add it in here, but that's the rationale on that one.

crates/weaver_resolver/src/attribute.rs
jsuereth: Yes - documented why.  TL;DR: in that case its' a remote dependency imported locally, - we need to track the remote.

crates/weaver_resolver/src/attribute.rs
jsuereth: Yes, fixed

crates/weaver_resolver/src/attribute.rs
jsuereth: Not possible to fix, updated warning.

crates/weaver_resolver/src/attribute.rs
jsuereth: Fixed

crates/weaver_resolver/src/conflict_strategy.rs
jsuereth: Fixed! good catch

lmolkova (approved):

PR #1576, opened 2026-07-07
Warn on invalid markdown snippet markers

ahfoysal:
Refs #501

This warns when `registry update-markdown` sees an HTML comment that looks like a semconv/weaver snippet marker but does not parse as one.

Marker validation now runs during the generator's existing line-by-line Markdown update pass and returns warnings as non-fatal `WResult` diagnostics, avoiding a second file read and parse. Existing whitespace-tolerant marker parsing remains unchanged.

Tests:
- `cargo fmt --all --check`
- `cargo test -p weaver_semconv_gen recognizes_invalid_snippet_markers`
- `cargo test -p weaver test_registry_update_markdown_warns_on_invalid_snippet_marker`
- `cargo clippy -p weaver_semconv_gen --all-targets --no-deps`

### DIFF
--- a/crates/weaver_semconv_gen/src/lib.rs
+++ b/crates/weaver_semconv_gen/src/lib.rs
@@ -21,28 +21,25 @@ pub use v2::SnippetGenerator as SnipperGeneratorV2;
 
 use crate::parser::{GenerateMarkdownArgs, WeaverGenerateMarkdownArgs};
 
-/// A suspicious markdown snippet marker found in a markdown document.
-#[derive(Debug, Clone, PartialEq, Eq)]
-pub struct InvalidSnippetMarker {
-    /// One-based line number.
-    pub line: usize,
-    /// The marker line.
-    pub marker: String,
-}
-
-/// Returns snippet-looking HTML comments that do not parse as supported markers.
+/// Returns warnings for snippet-looking HTML comments that do not parse as supported markers.
 #[must_use]
-pub fn invalid_snippet_markers(contents: &str) -> Vec<InvalidSnippetMarker> {
-    contents
+pub fn validate_markers(path: &str, contents: &str) -> DiagnosticMessages {
+    let marker_warnings = contents
         .lines()
         .enumerate()
         .filter_map(|(index, line)| {
-            parser::is_invalid_snippet_marker(line).then(|| InvalidSnippetMarker {
-                line: index + 1,
-                marker: line.to_owned(),
-            })
+            if parser::looks_like_snippet_marker(line) && !parser::is_valid_snippet_marker(line) {
+                Some(DiagnosticMessage::new(Error::InvalidSnippetMarker {
+                    path: path.to_owned(),
+                    line: index + 1,
+                    marker: line.to_owned(),
+                }))
+            } else {
+                None
+            }
         })
-        .collect()
+        .collect();
+    DiagnosticMessages::new(marker_warnings)
 }
 
 /// Errors emitted by this crate.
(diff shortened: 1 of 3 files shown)


src/registry/update_markdown.rs
```diff
@@ -179,11 +197,21 @@ pub(crate) fn command(
...
         log_info(format!("{}: ${}", operation, entry.path().display()));
-        if let Err(error) = generator.update_markdown(
-            &entry.path().display().to_string(),
-            dry_run,
-            attribute_registry_base_url.as_deref(),
-        ) {
+        let path = entry.path().display().to_string();
+        if let Ok(contents) = fs::read_to_string(entry.path()) {
+            for marker in invalid_snippet_markers(&contents) {
+                marker_warnings.extend_from_vec(vec![DiagnosticMessage::new(
```
lmolkova: we should be able to create DiagnosticMessages directly from `invalid_snippet_markers` method and then rename it to something like `validate_markers` (that returns marker_warnings array)

crates/weaver_semconv_gen/src/parser.rs
```diff
@@ -258,6 +258,29 @@ fn parse_semconv_trailer(input: &str) -> IResult<&str, ()> {
     }
 }
 
+/// Returns true if the line is an HTML comment that looks like a snippet marker,
+/// but does not parse as a supported marker.
+pub fn is_invalid_snippet_marker(line: &str) -> bool {
```
lmolkova: just from readability perspective, `is_invalid` is harder to read / understand

```suggestion
pub fn is_valid_snippet_marker(line: &str) -> bool {
```

ahfoysal:
Addressed the review comments: marker validation now returns diagnostics directly, the parser helper uses positive marker validation, and the bool::then path is gone.

crates/weaver_semconv_gen/src/parser.rs
```diff
@@ -159,7 +159,7 @@ fn parse_html_comment(input: &str) -> IResult<&str, &str> {
 fn parse_semconv_snippet_directive(input: &str) -> IResult<&str, GenerateMarkdownArgs> {
     let (input, _) = multispace0(input)?;
     let (input, _) = tag(SEMCONV_HEADER)(input)?;
-    let (input, _) = multispace0(input)?;
```
jsuereth, line 162: Why are you requiring spaces now?

src/registry/update_markdown.rs
```diff
@@ -179,11 +182,13 @@ pub(crate) fn command(
...
-        if let Err(error) = generator.update_markdown(
-            &entry.path().display().to_string(),
-            dry_run,
-            attribute_registry_base_url.as_deref(),
-        ) {
+        let path = entry.path().display().to_string();
+        if let Ok(contents) = fs::read_to_string(entry.path()) {
+            marker_warnings.extend(validate_markers(&path, &contents));
+        }
+        if let Err(error) =
```
jsuereth: I'd ideally like to NOT parse the file twice to get these warnings.

Can you adjust `generate.update_markdown` or some other helper method that would issue warnings on the fly, similar to how we use `WResult` in other places, rather than iterating over all the lines in the file twice?

While we read synchronously strings today - we MAY move to a streaming line-by-line processor over time - so I'd rather not design components that can't be used in that manner if possible.

What's there now is still better than nothing - just curious if it'd be a simple change to use the existing line-by-line parsing we're already doing in generator.

crates/weaver_semconv_gen/src/lib.rs
```diff
@@ -152,21 +132,47 @@ pub trait MarkdownSnippetGenerator {
...
+            Ok(contents) => contents.replace("\r\n", "\n"),
+            Err(error) => return WResult::FatalErr(Error::StdIoError(error.to_string())),
+        };
+        let (updated_markdown, marker_warnings) = match self
+            .update_markdown_contents_with_diagnostics(
+                file,
+                &original_markdown,
+                attribute_registry_base_url,
+            )
+            .into_result_with_non_fatal()
```
jsuereth, line 158: This may be dropping your warnings here.

You should be able to match directly on the `WResult` from `update_markdown_contents_with_diagnostics` to pull out the `Ok` vs. `OkWithNfes(result,warnings)`, e.g.

crates/weaver_semconv_gen/src/lib.rs
jsuereth, line 184: It MAY be tedious - and you can add a TODO for this if so -

But we can just remove this method *or* update its signature to take WResult directly.  I'd rather do that than have two methods here which do the same thing.

jsuereth: Can you remove this method now?

It looks like it's just a passthrough in all cases.

PR #1583, opened 2026-07-10
Use OS-native cert store to validate TLS connections

jerbly:
When downloading a registry package via HTTPS the agent `ureq` was configured just to use its default bundled root CA. Changed this to use `RootCerts::PlatformVerifier`, so certificate validation now goes through the operating system's native trust store.

This meant that on a corporate network with proxy, access to github artifacts failed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ All notable changes to this project will be documented in this file.
 
 # Unreleased
 
+- Use the OS-native certificate store (via ureq's `platform-verifier` feature) to validate TLS connections for remote registry downloads, instead of a fixed bundled root CA list. ([#TODO](https://github.com/open-telemetry/weaver/pull/TODO) by @jerbly)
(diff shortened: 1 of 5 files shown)


jsuereth:
@jerbly did you test this on docker?  IIRC that was one reason we didn't use native - but I could be mis-remembering

jsuereth (approved):

jerbly:
> @jerbly did you test this on docker? IIRC that was one reason we didn't use native - but I could be mis-remembering

@jsuereth - tested on Docker and added a guide for how to set a custom CA for the container to use.

On the native vs rustls, it's a bit of a mess that needs tidying up. We're actually using both at the moment and two different http libs. For files and archives we're using `ureq` with `rustls`. For git, `gix` is setup to use `native` via `reqwest`. 🙃 - we moved gix to native because of some linker issue in Alpine a while ago.

PR #1587, opened 2026-07-12
Allow cross-registry refinements

lmolkova:
When refining something from dependency, weaver fails with 

```
  × The following `extends` clause reference is not resolved for the group
  │ 'refined.request'.
  │ `extends` clause reference: span.server.request
  │ Provenance: Some(Provenance { schema_url: SchemaUrl { url: "https://span-
  │ consumer.example.com/schemas/1.0.0", name_range: 8..41, version_range:
  │ 42..47 }, path: "data/registry-test-v2-dep/span_registry/registry/
  │ registry.yaml" })
```

fixing it


crates/weaver_resolved_schema/src/v2/mod.rs
```diff
@@ -419,9 +410,7 @@ pub fn convert_v1_to_v2(
                 let is_refinement = g
                     .lineage
                     .as_ref()
-                    .and_then(|l| l.extends_group.as_ref())
-                    .and_then(|parent| group_type_lookup.get(parent))
-                    .map(|t| *t == GroupType::Metric)
+                    .map(|l| l.extends_group_type == Some(GroupType::Metric))
                     .unwrap_or(false);
```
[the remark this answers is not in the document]
lmolkova: there is no v1 resolved schema

jsuereth (approved):
Nice cleanup / extension!

crates/weaver_resolver/src/registry.rs
```diff
@@ -579,7 +579,7 @@ fn resolve_extends_references(ureg: &mut UnresolvedRegistry) -> Result<(), Error
                     unresolved_group.group.lineage.as_mut(),
                 );
                 if let Some(lineage) = unresolved_group.group.lineage.as_mut() {
-                    lineage.extends(extends);
+                    lineage.extends(extends, parent_summary.r#type.clone());
```
jsuereth, line 608: I find it amusing *this* is the key fix here :)

lmolkova: right?!

PR #1598, opened 2026-07-17
Fail fast with an actionable error when a local registry, dependency, template, or policy path does not exist

ANcpLua:
## Problem

A relative `registry_path` in a registry manifest dependency resolves against the process working directory. Running weaver from any other directory fails late in resolution with a bare

```
× The following error occurred during the processing of semantic convention
│ file: IO error for operation on ./dep-model: No such file or directory (os error 2)
```

— no mention of which base directory the path was resolved against, or that the CWD is what matters. Hit in practice consuming `semantic-conventions-genai` (multi-registry manifest with `registry_path: ./.build/sc-upstream-filtered`) from an external codegen pipeline: the main registry resolves fine via `-r`, then the dependency dies with the message above.

## Change

`VirtualDirectory` now rejects a nonexistent `LocalFolder` eagerly. The error names the path as written, the absolute path it resolved to against the current working directory, and explains the resolution semantics:

```
× Virtual directory `./dep-model` is invalid: local path does not exist
│ (resolved to `/home/user/somewhere/dep-model` against the current working
│ directory). Relative paths — including `registry_path` entries in a
│ registry manifest — resolve against the current working directory, not the
│ location of the file that declares them; run weaver from the directory the
│ path is relative to, or use an absolute path
```

Absolute paths get a plain "local path does not exist". No behavior change for existing paths. Because the check lives in the shared `VirtualDirectory` layer, it also covers the main `-r` registry path, template and policy paths, and the case where a mistyped URL parses as a `LocalFolder`.

## Non-goal

This does not change resolution semantics. Whether a dependency `registry_path` should (also) resolve relative to the manifest location is a separate design discussion — happy to open an issue laying out the options (resolution fallback order, opt-in field, breaking switch) if there is interest.

## Testing

- New unit test covering both branches (relative path with CWD hint, absolute path without).
- Full workspace test suite passes.
- Manually verified end-to-end: a minimal registry with a `./dep-model` dependency fails with the new message from the wrong CWD and resolves unchanged from the correct one.


jsuereth:
Generally fix looks good  - would approve, but I think Co-pilot found the bug in the windows build.  Please fix tests and then I think we can merge.

ANcpLua:
> Generally fix looks good - would approve, but I think Co-pilot found the bug in the windows build. Please fix tests and then I think we can merge.

Windows failure fixed: the test hard-coded `/definitely/does/not/exist` as the "absolute" case, but a drive-less path isn't absolute on Windows. Went with the `TempDir` approach suggested in the review: absolute and guaranteed nonexistent on every platform. Also switched `exists()` –> `try_exists()` per the other comment (permission/IO errors now report as "could not be checked" instead of "does not exist", with a unix test covering that arm), and merged main to pick up the regenerated `weaver-config.json` that was tripping the docs/schemas check.

PR #1600, opened 2026-07-17
Fix live-check HTTP response delivery during shutdown

trask:
When `weaver registry live-check --output http` receives `POST /stop`, the live-check command can exit after sending the report to the admin handler but before the detached Axum server finishes writing the response. With HTTP/2, terminating the process during that window resets the connection and leaves the client without the promised report.

This change coordinates shutdown between the live-check command and the admin server. After handing the report to `/stop`, the command now waits for Axum's graceful shutdown to complete, bounded by the existing 60-second admin request timeout.

The regression test uses HTTP/2, produces a response larger than typical socket buffers, and pauses after receiving the response headers before consuming the body. Without the lifecycle wait, the test fails while decoding the body with `ConnectionReset`; with the fix, the complete report is delivered and Weaver exits normally.

This was observed downstream in the OpenTelemetry Java instrumentation JMX tests: https://github.com/open-telemetry/opentelemetry-java-instrumentation/actions/runs/29557928989/job/87829898889

## Testing

- `cargo fmt --check`
- `cargo test --bin weaver registry::otlp::tests::test_http_stop_endpoint_with_report -- --nocapture`
- `cargo test -p weaver_live_check --test livecheck_emit test_livecheck_emit_roundtrip -- --nocapture`
- `cargo clippy --bin weaver --no-deps`
- `cargo clippy -p weaver_live_check --test livecheck_emit --no-deps`

### DIFF
--- a/crates/weaver_live_check/tests/livecheck_emit.rs
+++ b/crates/weaver_live_check/tests/livecheck_emit.rs
@@ -10,6 +10,8 @@ use std::thread::sleep;
 use std::time::Duration;
 use weaver_test_support::reserve_test_port;
 
+const RESPONSE_PADDING_SIZE: usize = 2 * 1024 * 1024;
+
 /// Guard that kills the child process on drop (e.g., on panic) to prevent orphaned processes.
 struct ChildGuard(Option<Child>);
 
@@ -59,13 +61,22 @@ fn wait_for_health(port: u16) {
     }
 }
 
-/// POST /stop and return the response body.
-fn stop_and_collect_report(port: u16) -> String {
+/// POST /stop over HTTP/2 and return the response body.
+async fn stop_and_collect_report(port: u16) -> String {
     let url = format!("http://127.0.0.1:{port}/stop");
-    let response = ureq::post(&url).send("").expect("POST /stop failed");
+    let response = reqwest::Client::builder()
+        .http2_prior_knowledge()
+        .build()
+        .expect("Failed to build HTTP/2 client")
+        .post(url)
+        .send()
+        .await
+        .expect("POST /stop failed");
+    // Leave the response body unread long enough to expose premature process shutdown.
+    tokio::time::sleep(Duration::from_millis(500)).await;
     response
-        .into_body()
-        .read_to_string()
+        .text()
+        .await
         .expect("Failed to read /stop response body")
 }
 
@@ -242,7 +253,8 @@ async fn test_livecheck_emit_roundtrip() {
             json!({
                 "attribute_key": "db.system",
                 "attribute_value": "postgresql",
-                "expected": "postgres"
+                "expected": "postgres",
+                "padding": "x".repeat(RESPONSE_PADDING_SIZE)
             }),
         );
         emitter.emit_finding(&finding, &sample_ref, &parent);
@@ -336,7 +348,11 @@ async fn test_livecheck_emit_roundtrip() {
     sleep(Duration::from_millis(500));
 
     // 6. Collect report via POST /stop
-    let report_body = stop_and_collect_report(admin_port);
+    let report_body = stop_and_collect_report(admin_port).await;
+    assert!(
+        report_body.len() > RESPONSE_PADDING_SIZE,
+        "Expected a report large enough to exercise asynchronous response delivery"
+    );
 
     // 7. Wait for weaver to exit
     //    Exit code may be non-zero if there are violations — we check that separately below.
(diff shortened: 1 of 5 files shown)


crates/weaver_live_check/tests/livecheck_emit.rs
```diff
@@ -59,13 +61,25 @@ fn wait_for_health(port: u16) {
...
+        .build()
+        .expect("Failed to build HTTP/2 client")
+        .post(url)
+        .send()
+        .await
+        .expect("POST /stop failed")
+        .error_for_status()
+        .expect("POST /stop returned an error status");
+    // Leave the response body unread long enough to expose premature process shutdown.
+    tokio::time::sleep(Duration::from_millis(500)).await;
```
lmolkova, line 79: any chance to make it deterministic without relying on sleep?

lmolkova (approved):

crates/weaver_live_check/tests/livecheck_emit.rs
trask: I couldn't find a good way

src/registry/otlp/mod.rs
```diff
@@ -50,6 +52,22 @@ pub struct AdminReportSender {
     pub sender: Arc<Mutex<Option<oneshot::Sender<(String, String)>>>>,
     /// Set to `true` to have `/stop` wait and return the report as its response body.
     pub expect_report: Arc<AtomicBool>,
+    server_shutdown: Arc<(Mutex<bool>, Condvar)>,
```
jerbly, line 55: Nit: could this be a plain mpsc::channel instead of a Condvar, matching the ready_tx/shutdown_tx pattern already in this file?

src/registry/live_check.rs
```diff
@@ -470,6 +470,7 @@ pub(crate) fn command(
                     .take()
                 {
                     let _ = sender.send((content_type, body));
+                    report.wait_for_server_shutdown();
```
jerbly, line 473: Nit: this new call runs while report.sender's lock is still held. This could stall the whole admin runtime for up to 60s if anything else needs that lock. Probably should drop the guard before waiting.

jerbly (approved):
Worth a changelog entry for this I'd say.

Thanks for the fix. A couple of nits, I think the channel approach might come out neater and with fewer lines needed.

crates/weaver_live_check/tests/livecheck_emit.rs
```diff
@@ -59,13 +61,25 @@ fn wait_for_health(port: u16) {
...
+        .timeout(Duration::from_secs(65))
+        .build()
+        .expect("Failed to build HTTP/2 client")
+        .post(url)
+        .send()
+        .await
+        .expect("POST /stop failed")
+        .error_for_status()
+        .expect("POST /stop returned an error status");
+    // Leave the response body unread long enough to expose premature process shutdown.
```
jsuereth, line 78: I'm curious on this one - can you describe the behavior happening better here in this comment so we understand how to change this code going forward if we refactor?

Why would forcing a sleep here actually solve the issue vs. async await?

jsuereth: Note: I may just be asking for more PR description to be moved here so we have it when we do refactoring.

src/registry/otlp/mod.rs
```diff
@@ -50,6 +52,22 @@ pub struct AdminReportSender {
...
+}
+
+impl AdminReportSender {
+    /// Wait until the admin server has finished its graceful shutdown or times out.
+    pub fn wait_for_server_shutdown(&self) {
+        let (shutdown_lock, shutdown_complete) = &*self.server_shutdown;
+        let is_shutdown = shutdown_lock
+            .lock()
+            .expect("Admin server shutdown lock poisoned");
```
jsuereth, line 64: We should probably call out WHY this is happening - e.g. the thread that held the lock unexpected crashed / panic'd so the lock becomes poisoned.

Generally - this means the whole system is in an inconcsistent state and further panic is ok - perhaps we can say that in this error.

src/registry/otlp/mod.rs
jsuereth: I think the threading aspects of this may need a  little tuning.  Yes the tests pass - but it looks pretty suspect to me - like we're relying on accidental behavior of how we're calling Tokio vs. some principled concurrency guards here.

1. I suggest we have a server shutdown struct that captures all the "state" we need to track and you put that in this Arc.
2. We expose methods, cleanly, from this struct in all the places you use it.  So instead of `AdminReportSender` implementing *half* the locking logic, and the rest manually written, we encapsulate *all* concurrency/synchronization in the new struct, something like "ShutdownCoordinator".
3. We clearly define expectations/locking behavior for each method of the `struct ShutdownCoordinator`.

jsuereth (changes requested):
Sorry - the concurrency junky in me is worried about this one...  

I have a suggestion for cleanup that should make this way easier to review.

trask:
> Sorry - the concurrency junky in me is worried about this one...

No worries 😅, probably better for me to just open an issue for this one: #1626

PR #1604, opened 2026-07-19
refinements: ref_group attribute props win over inherited ones

lmolkova:
Example of what I want to do:

```yaml
spans:
  type: messaging.send.client
  ...
  attributes:
    ref: messaging.destination.name
    brief: some generic definition of a topic/queue name
    ref: messaging.destination.something_else
    brief: something else generic definition
  type: messaging.receive.client
    ...
    attributes:
      ref: messaging.destination.name
      ref: messaging.destination.something_else

attribute_groups:
  # describe how to populate this common group on kafka
  id: kafka.destination.attrs
  attributes:
    ref: messaging.destination.name
    brief: Kafka topic name
    ref: messaging.destination.something_else
    brief: <This is what is means for Kafka>

span_refinements:
  type: kafka.send.client
  attributes:
    # reuse as a group, don't repeat
    - ref_group: kafka.destination.attrs
  type: kafka.receive.client
  attributes:
    # reuse as a group, don't repeat
    - ref_group: kafka.destination.attrs
```

It's not possible today because priority on refinement is 
- explicit attr ref
- inherited attrs
- ref_groups

If ref_group adds new attributes - all good, but if it tries to refine parent ones, reg_group attr refinements are ignored.

fixing it with a small refactoring

### DIFF
--- a/schemas/semconv-syntax.v2.md
+++ b/schemas/semconv-syntax.v2.md
@@ -308,6 +308,11 @@ spans:
 > It's NOT RECOMMENDED to use `ref_groups` on other attribute groups
 > due to readability concerns.
 
+> [!NOTE]
+> When a refinement declares the same attribute through more than one source,
+> priority is (highest first): inline `- ref:`, then `ref_group`, then the
+> attribute inherited from the refined (parent) signal.
+
 ### `spans` definition
 
 The spans section contains a list of span definitions. A span definition consists of the following properties:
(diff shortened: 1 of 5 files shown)


schemas/semconv-syntax.v2.md
```diff
@@ -308,6 +308,11 @@ spans:
 > It's NOT RECOMMENDED to use `ref_groups` on other attribute groups
 > due to readability concerns.
 
+> [!NOTE]
+> When a refinement declares the same attribute through more than one source,
```
jsuereth, line 312: Maybe we shoudl call out the *order* you do this in the YAML does not matter, this precedence wins?

jsuereth (approved):

Issue #1360, opened 2026-04-14
live-check: schema url validation


lmolkova:
From https://cloud-native.slack.com/archives/C0697EXNTL3/p1776193648417459

We need to have means to validate schema  url and leverage it when doing live-checks.

1. We should validate telemetry against schema url provided in the actual telemetry if it's a v2 schema
2. If schema url is not included on telemetry we should issue info advise
    - at least in OTel instrumentation, we need to enforce all instrumentations to include it. We can have a custom otel-wide policy to make it a violation
3. We need to provide access to schema URL in custom policies:
     - via instrumentation scope as a sample? 
     - more importantly, we should have instrumentation scope to be included on the signal sample being validated (not attribute sample).
        - e.g. when I'm validating GenAI span, I should be able to check which schemaUrl was used, which instrumentation scope name was provided, etc (at least for 'good error message' reasons)

McGluut:
I tested a local implementation of the instrumentation-scope-on-signal part of this issue. It adds optional scope name, version, attributes, dropped-attribute count, and the enclosing scope schema URL to live-check span, metric, and log samples.

The OTLP ingester attaches that context to each child signal while retaining the existing standalone `Sample::Attribute` entries for scope attributes. Older JSON without the field remains valid, missing scope remains `None`, resource and scope stay separate, and custom Rego policies can inspect the scope fields. The local tests cover JSON compatibility/round-trip, OTLP propagation, same-named spans under different scopes, all three signal types, schema-only scope, resource separation, non-duplication of standalone scope attributes, and the real custom-policy path.

Before preparing a PR, could you confirm two scope choices?

1. Is span/metric/log parity the preferred review unit, or should the first PR cover a smaller signal surface?
2. Should scope attributes remain both attached as context and emitted through the existing standalone validation path, or should that behavior change as part of #1360?

PR #1605, opened 2026-07-19
live-check: preserve instrumentation scope on signal samples

McGluut:
## Summary

- preserve OTLP instrumentation-scope name, version, schema URL, attributes, and dropped-attribute count on live-check span, metric, and log samples through one shared `Rc`
- emit each instrumentation scope once as a normal root `Sample::InstrumentationScope`, ordered after its resource and before its child signals
- expose both `resource` and `instrumentation_scope` beside `sample` to custom Rego policies, while keeping the shared context out of each signal's serialized shape
- render instrumentation scope through the ordinary output/template path and document the custom-advisor input contract

## Why

Live-check currently loses the instrumentation scope that owns a signal. That makes same-named operations from different libraries indistinguishable and prevents custom policies from inspecting the scope schema URL or other scope metadata.

The OTLP container already supplies this information. This change preserves the ownership context without flattening it into resource metadata or repeating it on every signal. The root scope sample gives output and validation one normal carrier, matching the existing resource pattern.

## Compatibility

- legacy signal JSON without `instrumentation_scope` remains valid and retains its serialized shape
- shared resource and instrumentation-scope context remains skipped on individual signal serialization
- a missing scope plus an empty schema URL remains absent
- a schema URL is preserved and emitted once even when the OTLP scope message is absent
- scope attributes are validated exactly once under the root scope sample instead of being emitted as separate root attribute samples
- resource and instrumentation scope remain distinct; no positional scope association is added to JSON input

## Validation

Focused tests cover:

- one root scope carrier per OTLP scope, ordered `Resource -> InstrumentationScope -> signals`
- shared `Rc` identity for signals in one scope
- span, metric, and log propagation
- true absence versus schema-URL-only scope
- grouped scope-attribute validation without standalone duplication
- ordinary-path ANSI rendering
- root-versus-adjacent Rego context and scoped/unscoped `null` behavior
- legacy signal JSON compatibility

Local source gates on commit `721aaad`:

- Rust formatting check
- Cargo workspace metadata resolution
- YAML parse
- Git diff hygiene
- independent source/contract review

Fresh local Rust linking is unavailable in this Windows environment because the MSVC `link.exe`/import-library toolchain is not installed. GitHub Actions on the pushed commit are therefore the executable cross-platform gate.

## Review scope

This PR intentionally covers spans, metrics, and logs together because all three use the same OTLP resource/scope container semantics. The latest refactor follows review feedback by modeling scope as a normal sample carrier rather than adding output-specific template context.

Refs #1360

### DIFF
--- a/crates/weaver_live_check/src/advice/rego_advisor.rs
+++ b/crates/weaver_live_check/src/advice/rego_advisor.rs
@@ -9,9 +9,10 @@ use weaver_forge::jq;
 
 use super::{emit_findings, Advisor};
 use crate::{
-    live_checker::LiveChecker, otlp_logger::OtlpEmitter, Error, Sample, SampleRef,
-    VersionedAttribute, VersionedSignal, DEFAULT_LIVE_CHECK_JQ, DEFAULT_LIVE_CHECK_REGO,
-    DEFAULT_LIVE_CHECK_REGO_POLICY_PATH,
+    live_checker::LiveChecker, otlp_logger::OtlpEmitter,
+    sample_instrumentation_scope::SampleInstrumentationScope, sample_resource::SampleResource,
+    Error, Sample, SampleRef, VersionedAttribute, VersionedSignal, DEFAULT_LIVE_CHECK_JQ,
+    DEFAULT_LIVE_CHECK_REGO, DEFAULT_LIVE_CHECK_REGO_POLICY_PATH,
 };
 
 /// An advisor which runs a rego policy on the attribute
@@ -104,6 +105,8 @@ impl RegoAdvisor {
 #[derive(Serialize)]
 struct RegoInput<'a> {
     sample: SampleRef<'a>,
+    resource: Option<&'a SampleResource>,
+    instrumentation_scope: Option<&'a SampleInstrumentationScope>,
     registry_attribute: Option<Rc<VersionedAttribute>>,
     registry_group: Option<Rc<VersionedSignal>>,
 }
@@ -120,6 +123,8 @@ impl Advisor for RegoAdvisor {
         let mut findings = self.check(RegoInput {
             sample: sample.clone(),
             registry_attribute,
+            resource: signal.resource(),
+            instrumentation_scope: signal.instrumentation_scope(),
             registry_group,
         })?;
 
--- a/crates/weaver_live_check/src/sample_log.rs
+++ b/crates/weaver_live_check/src/sample_log.rs
@@ -13,6 +13,7 @@ use crate::{
     advice::{check_entity_associations, emit_findings, FindingBuilder},
     live_checker::LiveChecker,
     sample_attribute::SampleAttribute,
+    sample_instrumentation_scope::SampleInstrumentationScope,
     sample_resource::SampleResource,
     Error, FindingId, LiveCheckResult, LiveCheckRunner, LiveCheckStatistics, Sample, SampleRef,
     VersionedSignal,
@@ -36,6 +37,9 @@ pub struct SampleLog {
     pub trace_id: Option<String>,
     /// Span ID if the event is correlated with a span
     pub span_id: Option<String>,
+    /// Instrumentation scope that produced this log record.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub instrumentation_scope: Option<SampleInstrumentationScope>,
     /// Live check result
     pub live_check_result: Option<LiveCheckResult>,
     /// Reference to the parent resource (not serialized)
--- a/crates/weaver_live_check/src/sample_log.rs
+++ b/crates/weaver_live_check/src/sample_log.rs
@@ -37,9 +37,9 @@ pub struct SampleLog {
     pub trace_id: Option<String>,
     /// Span ID if the event is correlated with a span
     pub span_id: Option<String>,
-    /// Instrumentation scope that produced this log record.
-    #[serde(default, skip_serializing_if = "Option::is_none")]
-    pub instrumentation_scope: Option<SampleInstrumentationScope>,
+    /// Shared instrumentation scope that produced this log record (not serialized).
+    #[serde(skip)]
+    pub instrumentation_scope: Option<Rc<SampleInstrumentationScope>>,
     /// Live check result
     pub live_check_result: Option<LiveCheckResult>,
     /// Reference to the parent resource (not serialized)
--- a/defaults/live_check_templates/ansi/live_check_macros.j2
+++ b/defaults/live_check_templates/ansi/live_check_macros.j2
@@ -78,6 +78,19 @@
 {%- endfor %}
 {% endmacro %}
 
+{% macro display_instrumentation_scope(scope, indent=0) %}
+{{ " " * indent }}{{ ("Instrumentation scope") | ansi_bright_cyan | ansi_bold }} {{ scope.name }}{% if scope.version %} `{{ scope.version }}`{% endif %}
+{% if scope.schema_url %}
+{{ " " * indent }}  schema URL: {{ scope.schema_url }}
+{% endif %}
+{% if scope.dropped_attributes_count %}
+{{ " " * indent }}  dropped attributes: {{ scope.dropped_attributes_count }}
+{% endif %}
+{%- for attribute in scope.attributes %}
+{{ " " * (indent + 4) }}- {{ attribute.name }}{% if attribute.value %} = {{ attribute.value }}{% endif %}
+{%- endfor %}
+{% endmacro %}
+
 {% macro summarize_data_point_value(data_point) %}
 {%- if data_point.value is defined -%}
 {{ data_point.value }}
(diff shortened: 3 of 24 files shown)


crates/weaver_live_check/src/sample_log.rs
```diff
@@ -36,6 +37,9 @@ pub struct SampleLog {
     pub trace_id: Option<String>,
     /// Span ID if the event is correlated with a span
     pub span_id: Option<String>,
+    /// Instrumentation scope that produced this log record.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
```
jsuereth: @jerbly Wanted to check on this design choice here with you - should we skip serializing instrumentation scope?

I didn't see that in other options here and wasn't sure if this would be a breaking change to have it always serialzied or not.

jsuereth (approved):
Overall - this looks like a nice addition, thanks for contributing!

crates/weaver_live_check/src/sample_log.rs
McGluut: From my side, this was deliberate. `default` preserves deserialization of existing samples, while `skip_serializing_if = "Option::is_none"` preserves the existing serialized shape when no scope is provided, so legacy input does not gain `"instrumentation_scope": null` on round-trip.

This is covered by the compatibility test. I’m happy to align with Weaver’s convention if always serializing optional fields is preferred.

lmolkova: I don't believe there is any scenario when this repo would deserialize existing samples and from rego side it's easier to work with `instrumentation_scope: None` than with a property that can be absent. 
There should also be no concern with instrumentation_scope showing up in existing rego policies.

```suggestion
```

This would also be consistent with existing properties that show up no matter what.

lmolkova (approved):
@McGluut thanks for the contribution!

I'd like to start release sometime soon and would love to get this change in - it'd be really useful in some tests we have in otel. 

If you're around, could you please take a look at the open discussion. Otherwise I might just auto-apply the suggestion I left and merge the PR in a hope you won't mind it. Thanks!

src/registry/otlp/otlp_ingester.rs
```diff
@@ -68,16 +68,21 @@ impl OtlpIterator {
...
                                 self.buffer.push(Sample::Attribute(
-                                    sample_attribute_from_key_value(&attribute),
+                                    sample_attribute_from_key_value(attribute),
                                 ));
                             }
                         }
 
                         for log_record in scope_log.log_records {
                             let mut sample_log = otlp_log_record_to_sample_log(&log_record);
+                            sample_log.instrumentation_scope = instrumentation_scope.clone();
```
jerbly, line 83: We should match what we do with `resource` for `instrumentation_scope` using an `Rc` instead of cloning.

crates/weaver_live_check/src/sample_log.rs
jerbly: We should match what we do with `resource` (see below) - we intentionally do not clone and do not serialize the resource otherwise it repeats many times in memory and in the output. However, we should make both `resource` and `instrumentation_scope` available to rego somehow. (Probably by serializing them as something adjacent to the signal).

We should also better represent `instrumentation_scope` in the standard Text output on the console.

jerbly (changes requested):
Thank you - I think this is pretty close and useful. I would like it to be aligned with how we handle `resource` though. And in solving how to pass the `instrumentation_scope` to rego we can do the same for `resource`.

crates/weaver_live_check/src/sample_log.rs
McGluut: I checked this against the existing resource and JSON paths. The resource-aligned implementation would make three linked changes: store scope as `Option<Rc<SampleInstrumentationScope>>` with `#[serde(skip)]`, expose both `resource` and `instrumentation_scope` as top-level fields beside `sample` in `RegoInput`, and add scope to the standard text representation.

One contract question before I implement it: should JSON file/stdin input mirror resource through a standalone instrumentation-scope sample that applies to following signals, or should scope become OTLP-only context? Weaver currently deserializes `Vec<Sample>` from JSON and associates a preceding resource positionally, so skipping the signal field without a replacement carrier would drop scope on JSON round-trip.

Once that carrier is selected, I can update all three signal types and the focused tests together.

src/registry/otlp/otlp_ingester.rs
McGluut: Implemented in 8c15c26. Each OTLP scope is converted once into an `Rc<SampleInstrumentationScope>`, and that same allocation is shared by its signals. The focused `spans_from_the_same_otlp_scope_share_context` test asserts this with `Rc::ptr_eq`.

crates/weaver_live_check/src/sample_log.rs
McGluut: Implemented the resource-aligned refactor in 8c15c26: span, metric, and log hold `Option<Rc<SampleInstrumentationScope>>` with `serde(skip)`; Rego receives both `resource` and `instrumentation_scope` beside `sample` (with absent scope as `null`); and standard ANSI output renders scope context once. Scoped JSON input is intentionally not introduced here, while legacy scope-free JSON remains supported. I have left this thread open for maintainer verification.

McGluut:
Addressed the requested resource alignment in 8c15c26:

- instrumentation scope is allocated once per OTLP scope and shared with its signals through `Rc`;
- signal serialization no longer repeats the scope;
- Rego receives both `resource` and `instrumentation_scope` beside `sample`;
- standard ANSI output renders the scope context once.

Legacy scope-free JSON remains supported; scoped JSON input is intentionally not introduced by this refactor. Both inline threads have commit-linked replies and remain open for maintainer verification. The branch is rebased onto current `main`.

src/registry/live_check.rs
```diff
@@ -204,7 +206,55 @@ fn default_advisors() -> Vec<Box<dyn Advisor>> {
     ]
 }
 
-/// Generate output for a complete report - handles line-oriented special case
+/// Template-only view that keeps shared signal context adjacent to the sample.
```
jerbly: We shouldn't need this Template section at all. The instrumentation scope should appear in the output just like resource does today. It should not need special handling.

jerbly (commented):
We're very close. This is really good. Let's remove that Template section and make sure to update the docs: https://github.com/open-telemetry/weaver/tree/main/crates/weaver_live_check#custom-advisors

src/registry/live_check.rs
McGluut: Addressed in `721aaad`. Instrumentation scope is now emitted once as a normal root `Sample::InstrumentationScope`, after its resource and before its child signals. The `TemplateSample`/`TemplateLiveCheckReport` adapter and content-type special casing are removed; ANSI now renders the root variant through the ordinary `display_sample` path.

Signals still share one `Rc<SampleInstrumentationScope>` for Rego context without serializing it repeatedly. Scope attributes run through the grouped scope carrier exactly once, and the custom-advisor docs now describe both `input.resource` and `input.instrumentation_scope`. Focused tests cover ordering/count, all three signal types, schema-only versus absent scope, root-versus-adjacent Rego context, and ordinary-path ANSI rendering.

The workflows for this fork update are queued and currently require maintainer approval. I am leaving the thread open for maintainer verification.

McGluut:
Addressed the follow-up review in `721aaad`:

- replaced the template-only adapter with a normal once-per-scope sample carrier, matching the existing resource pattern;
- preserved shared `Rc` context on child signals for Rego without repeated serialization;
- grouped scope-attribute validation under that carrier;
- updated the custom-advisor documentation and generated sample-type model/docs;
- added focused coverage for ordering, absence/schema-only behavior, all three signal types, Rego context boundaries, and ordinary ANSI rendering.

The PR description now matches the revised design. The workflows for this fork update are queued and currently require maintainer approval; the review threads remain open for maintainer verification.

McGluut:
Following up on @lmolkova's review and the open discussion: I addressed the requested changes in `721aaad` and the subsequent CI-only statistics correction in `c344f74`.

The current head includes the resource-aligned instrumentation-scope carrier, the ordinary output path, documentation updates, and focused tests. No further author changes are pending; the current-head workflows are awaiting maintainer approval.

crates/weaver_live_check/src/advice/rego_advisor.rs
```diff
@@ -104,6 +105,8 @@ impl RegoAdvisor {
 #[derive(Serialize)]
 struct RegoInput<'a> {
     sample: SampleRef<'a>,
+    resource: Option<&'a SampleResource>,
+    instrumentation_scope: Option<&'a SampleInstrumentationScope>,
```
jerbly, line 109: Not for this PR, but there's a performance improvement opportunity here. Every sample: attribute, span, metric, ... will now cause the same resource and instrumentation_scope to be serialized to send to Rego every time. I think we should capture this into an issue to measure and then optimize. (Note to self).

defaults/live_check_templates/ansi/live_check_macros.j2
```diff
@@ -78,6 +78,20 @@
 {%- endfor %}
 {% endmacro %}
 
+{% macro display_instrumentation_scope(scope, indent=0) %}
+{{ " " * indent }}{{ ("Instrumentation scope") | ansi_bright_cyan | ansi_bold }} {{ scope.name }}{% if scope.version %} `{{ scope.version }}`{% endif %}
```
jerbly: Can we use `display_sample_header` here for the `scope.name`? This will mean the name is coloured according to the finding level.

jerbly (approved):
This is a good, in-depth update through live-check, not a small task! Please add a Changelog entry so you can take credit!

defaults/live_check_templates/ansi/live_check_macros.j2
McGluut: Addressed in `20b61a3`: the scope name now uses `display_sample_header(scope.live_check_result.highest_advice_level, scope.name)`, and the focused ANSI rendering test asserts the resulting finding-level colour.

McGluut:
Added the requested changelog credit in `20b61a3`, and rebased the branch onto current `main`. The same commit also applies `display_sample_header` to instrumentation scope names with a focused ANSI assertion.

crates/weaver_live_check/src/advice/rego_advisor.rs
McGluut: Agreed — keeping this out of #1605. I’d make the follow-up benchmark-first, separating shared-context serialization cost from Rego evaluation and per-sample policy cost. I’d be happy to help turn that into an issue after this lands.

PR #1612, opened 2026-07-20
fix(live-check): findings on resource samples carry resource attributes

fabiovincenzi:
## Problem

Findings on resource-level attributes are emitted (via `--emit-otlp-logs`) with
none of the source resource's attributes, while findings on span/metric/log
samples correctly carry them.

`Sample::resource()` returns `None` for a `Sample::Resource`: it only returns a
sample's *parent* resource, and a resource has no parent.

## Fix

Make a resource sample its own resource.

The only other callers are `sample_log.rs` and `sample_metric.rs`, and both only
ever run on log/metric samples, so this doesn't change anything for them.

## Testing

Unit tests covering: a resource sample is its own resource, span samples still
return their parent resource, samples without one still return `None`, and the
emit path flattens a resource sample's attributes into a finding.
Without the fix these fail with `left: None, right: Some("my-test-service")`.


crates/weaver_live_check/src/lib.rs
```diff
@@ -620,3 +624,59 @@
...
+        let resource = test_resource();
+        let sample = Sample::Resource(resource.clone());
+
+        assert_eq!(sample.resource(), Some(&resource));
+    }
+
+    #[test]
+    fn test_signal_sample_returns_parent_resource() {
+        let resource = test_resource();
+        let span = SampleSpan {
```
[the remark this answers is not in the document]
fabiovincenzi: Fixed!

lmolkova (approved):
Thanks for the fix!

A couple small things:

1. Please add n entry in CHANGELOG.md.
2. Should `Sample::instrumentation_scope()` also return scope for `Sample::InstrumentationScope(s)` for symmetry? This is probably not in scope of this PR

Otherwise looks good to me.

fabiovincenzi:
> Thanks for the fix!
> 
> A couple small things:
> 
> 1. Please add n entry in CHANGELOG.md.
> 2. Should `Sample::instrumentation_scope()` also return scope for `Sample::InstrumentationScope(s)` for symmetry? This is probably not in scope of this PR
> 
> Otherwise looks good to me.

Thanks! @lmolkova CHANGELOG entry added.

On `instrumentation_scope()`: looks intentional, `tests/instrumentation_scope.rs` asserts it returns None for a scope sample, since its only consumer is the Rego policy input and returning it there would just duplicate input.sample.

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ All notable changes to this project will be documented in this file.
...
   - Added [Matchers](crates/weaver_live_check/docs/matchers.md), a guide with a worked example for each sample type. Corrected the config section in the live-check and config READMEs from `[live_check]` to `[live-check]`. The underscore form was silently ignored.
 - Config sections are checked when they are read. A command section in `.weaver.toml` that does not deserialize, or an unknown key under `[live-check]`, now stops the run instead of being ignored. ([#1721](https://github.com/open-telemetry/weaver/pull/1721) by @jerbly)
+- Fix findings on resource samples being emitted without the resource's attributes. `Sample::resource()` returned `None` for a resource sample: it only returned a sample's parent resource, and a resource has no parent. A finding on a resource-level attribute therefore reached `--emit-otlp-logs` with none of its source resource's attributes, while findings on span, metric and log samples carried them. ([#1612](https://github.com/open-telemetry/weaver/pull/1612) by @fabiovincenzi)
```
lmolkova:
```suggestion
- Fix findings on resource samples being emitted without the resource's attributes. ([#1612](https://github.com/open-telemetry/weaver/pull/1612) by @fabiovincenzi)
```

neil-the-nowledgeable, line 18: lgtm

neil-the-nowledgeable (approved):
looks right to me

fabiovincenzi:
@lmolkova @neil-the-nowledgeable Thanks for the review! Does this need anything else to be merged?

PR #1619, opened 2026-07-22
Live-check: Add sample_name scoping to finding filters

### LINKED ISSUE #1613: live-check: scope finding exclusions by sample name (with wildcards), combinable with advice type

When validating telemetry with `registry live-check`, it's common to have a *small,
known* set of findings that are intentional and reviewed, while still wanting to fail
on everything else. For example, a registry may deliberately declare a few attributes
that trigger `illegal_namespace` (namespace collisions with existing upstream
attributes). Those specific cases are accepted on purpose — but a **new**
`illegal_namespace` on any *other* sample should still fail.

Today `[[live-check.finding_filters]]` allows excluding by advice id **or** by sample,
but within a single filter `exclude` (by id) and `exclude_samples` (by sample) combine
as **OR** (scoping is otherwise by signal type). So the only expressible options are:

- "drop **all** occurrences of an advice" (`exclude = ["illegal_namespace"]`) → too
  broad, hides genuinely new problems; or
- "drop **everything** on these samples" (`exclude_samples = [...]`) → also too broad,
  hides unrelated findings on those samples.

There is no way to say *"ignore `illegal_namespace` **only** for these specific
samples."*

Allow a finding filter to scope by **sample name AND advice type together** (AND, not
OR), with **wildcard** support on sample names. for example:

```toml
[[live-check.finding_filters]]
advice = ["illegal_namespace"]          # match this advice type/id
samples = [                             # ...AND only on these samples
  "server.address", "server.port",
  "client.address", "client.port",
  # wildcards welcome, e.g. "http.*"
]
action = "exclude"
```

Anything matching *both* the advice and a listed sample is dropped; the same advice on
a different sample still surfaces as a finding. Wildcards (`http.*`) would keep sample
lists compact.

- `exclude = ["<advice>"]` — suppresses the advice type globally, hiding genuinely new
  occurrences.
- `exclude_samples = [...]` — drops *every* finding on those samples, not just the one
  advice being accepted.
- Post-processing the report outside weaver to drop specific message/sample
  combinations — works, but keeps this policy out of `weaver.toml` where it's harder to
  review and can't be shared as config.

**Additional context**

This is an issue encountered on the [OBI](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation) project, where we used a [post-processing of the weaver findings](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/1cfa8ea12ca193ce30a2c714896cb9b0cb939af6/internal/test/weavercheck/weavercheck.go#L65) to ignore those messages

jerbly:
Live-check: (Fixes: [#1613](https://github.com/open-telemetry/weaver/issues/1613)) add `sample_names` to `[[live-check.finding_filters]]` to scope a filter to matching sample names, with glob wildcard support (also added to `exclude_samples`).

Allows you to have config like so:

```toml
# What we could do in the current release:
[[live-check.finding_filters]]
exclude = ["invalid_format"]

# Now exclude missing_attribute findings but only for all aws samples and task.id
[[live-check.finding_filters]]
exclude = ["missing_attribute"]
sample_names = ["aws.*", "task.id"]

# Now only include violation level findings for samples starting with "Task"
[[live-check.finding_filters]]
min_level = "violation"
sample_names = ["Task*"]
```

### DIFF
--- a/crates/weaver_config/src/live_check.rs
+++ b/crates/weaver_config/src/live_check.rs
@@ -198,7 +198,8 @@ impl Default for LiveCheckEmitConfig {
 }
 
 /// A filter that drops findings by ID exclusion or minimum level.
-/// Optional `signal_type` scopes the filter to a specific signal type.
+/// Optional `signal_type` and `sample_names` scope the filter to a specific
+/// signal type and/or set of sample names.
 #[derive(Debug, Clone, Deserialize, PartialEq, JsonSchema)]
 pub struct FindingFilter {
     /// Drop findings with these IDs.
@@ -208,12 +209,18 @@ pub struct FindingFilter {
     /// Optional signal type scope. When set, this filter only applies to
     /// findings with a matching signal_type.
     pub signal_type: Option<String>,
-    /// Drop all findings for samples with these names.
+    /// Drop all findings for samples with these names (supports glob
+    /// wildcards, e.g. `"trace.*"`).
     /// For attribute samples, this matches the attribute key — e.g.
     /// `["trace.parent_id", "trace.span_id"]` suppresses all findings
     /// (e.g. `missing_attribute`) for those attribute keys.
     #[serde(default)]
     pub exclude_samples: Vec<String>,
+    /// Optional sample name scope (supports glob wildcards, e.g. `"http.*"`).
+    /// When set, this filter only applies to samples whose name matches one
+    /// of these patterns, in addition to any `signal_type` scope.
+    #[serde(default)]
+    pub sample_names: Vec<String>,
 }
 
 #[cfg(test)]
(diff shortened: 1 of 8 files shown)


crates/weaver_config/src/live_check.rs
```diff
@@ -208,12 +209,18 @@ pub struct FindingFilter {
...
     /// For attribute samples, this matches the attribute key — e.g.
     /// `["trace.parent_id", "trace.span_id"]` suppresses all findings
     /// (e.g. `missing_attribute`) for those attribute keys.
     #[serde(default)]
     pub exclude_samples: Vec<String>,
+    /// Optional sample name scope (supports glob wildcards, e.g. `"http.*"`).
+    /// When set, this filter only applies to samples whose name matches one
+    /// of these patterns, in addition to any `signal_type` scope.
+    #[serde(default)]
+    pub sample_names: Vec<String>,
```
lmolkova, line 223: i've been trying to gaslight my AI that this already exists a few days ago :)

lmolkova (approved):

PR #1620, opened 2026-07-22
Align MSRV settings with CI

jerbly:
CI is checking against MSRV 1.88 - config in cargo.toml and clippy.toml was still on 1.81. This meant linting (and other things) was misaligned with our minimum.

(Spotted this because I wanted to use [is_none_or](https://doc.rust-lang.org/std/option/enum.Option.html#method.is_none_or) which is in 1.82)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,7 +33,7 @@ edition = "2021"
 repository = "https://github.com/open-telemetry/weaver"
 license = "Apache-2.0"
 publish = false
-rust-version = "1.81.0"
+rust-version = "1.88.0"
 
 [workspace.dependencies]
 serde = { version = "1.0.228", features = ["derive", "rc"] }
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -33,7 +33,7 @@ edition = "2021"
 repository = "https://github.com/open-telemetry/weaver"
 license = "Apache-2.0"
 publish = false
-rust-version = "1.81.0"
+rust-version = "1.88.0"
```
lmolkova, line 36: is this not controlled by renovate?

lmolkova (approved):

Cargo.toml
jerbly: No, and for good reason, we need to set our own strategy on how far back in versions we want our crates to be compatible. Not that we've ever released the crates yet!

PR #1622, opened 2026-07-22
make deprecated.note optional on renamed deprecations

### LINKED ISSUE #1556: Renamed_to deprecations now require note with v2
**Describe the bug**
Prior to v1 we could define a deprecation without a note, however with v2 it is now required.

**To Reproduce**
1. Deprecate an attribute in your yaml using v1 syntax or check out semconv
2. Confirm weaver passes
3. Switch yaml with deprecated attribute to v2 syntax
4. Observe failure with the error below:
```
Diagnostic report:

  × The following YAML snippet does not match any of the allowed schemas.
  │ (Variant 1):
  │ - Value {"reason":"renamed","renamed_to":"http.request.method"} does not
  │ match any schema in a 'oneOf' group; it must match exactly one.
  │   (Variant 1):
  │   - Missing required property: "note".
  │   (Variant 2):
  │   - Missing required property: "note".
  │   - Value "renamed" does not match the required constant: "obsoleted".
  │   (Variant 3):
  │   - Missing required property: "note".
  │   - Value "renamed" does not match the required constant: "uncategorized".
  │   (Variant 4):
  │   - Missing required property: "note".
  │   - Value "renamed" does not match the required constant: "unspecified".
  │ (Variant 2):
  │ - Value {"reason":"renamed","renamed_to":"http.request.method"} is not of
  │ the required type: 'null'.

```

**Expected behavior**
When deprecating an attribute as it has been renamed, we shouldn't need to specify a note.

**Additional context**

Example attribute valid in v1
```yaml
  - key: http.method
    type: string
    brief: 'Deprecated, use `http.request.method` instead.'
    stability: development
    deprecated:
      reason: renamed
      renamed_to: http.request.method
    examples: ["GET", "POST", "HEAD"]
```

The following page says the note is optional https://github.com/open-telemetry/weaver/blob/2ccfcaf7d4c4e7240dfc6afd3f49c4c41f7674c9/docs/schema-changes.md#data-model

lmolkova:
Fixes #1556

on v1 we had optional deprecated.note (at least for renamed_to) and it make sense

```yaml
deprecated:
  reason: renamed
  rename_to: foo.bar
```

It's self-descriptive. We can generate a perfect note like`Renamed to 'foo.bar'`. 
Requiring a note would mean duplication in 99% of cases.

So this PR makes note optional for renamed.

Unfortunately it leaks to resolved schema where we actually would always populate note (inferred from structured props) because we share models between definition and resolved schema and it'd be a giant change to prevent it and make note required on resolved side only.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ All notable changes to this project will be documented in this file.
 
 # Unreleased
 
+- Make `deprecated.note` optional for `{reason: renamed}` - inferred from `renamed_to`. ([#TODO](https://github.com/open-telemetry/weaver/pull/TODO) by @lmolkova)
 - Add a tree view to the `serve` UI's search page, grouping results by namespace with expand/collapse controls, and a "Hide deprecated" toggle (on by default) for both the list and tree views. ([#1595](https://github.com/open-telemetry/weaver/pull/1595) by @jerbly)
 - Support signal refinements over a published dependency. ([#1587](https://github.com/open-telemetry/weaver/pull/1587) by @lmolkova)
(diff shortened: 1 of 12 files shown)


jsuereth (approved):
Cleaning up / detangling resolved schema - we can investigate later.  This looks good to go.

PR #1632, opened 2026-07-24
Fix stop race condition for http output

### LINKED ISSUE #1626: live-check can exit before the /stop response body is fully delivered
**Describe the bug**

When `weaver registry live-check --output http` receives `POST /stop`, Weaver can exit before sending the complete response. This can reset the connection, especially with HTTP/2 and larger reports, leaving the client without the report.

**To Reproduce**

1. Start live-check with HTTP output and an OTLP input, for example:

   ```console
   weaver registry live-check \
     -r <registry> \
     --format json \
     --output http \
     --otlp-grpc-port 4317 \
     --admin-port 8080
   ```

2. Send enough OTLP telemetry to produce a report larger than typical socket buffers. A report of approximately 2 MiB reproduces the problem reliably.
3. Use an HTTP/2 client to send `POST http://localhost:8080/stop`.
4. After receiving the response headers, delay consuming the response body briefly. With a `reqwest` client, `send().await` receives the headers and `text().await` later drains the body.
5. Observe that reading the body can fail with `ConnectionReset` because Weaver exits while the response is still being transmitted.

The issue has also occurred in a downstream JMX test without an artificial delay:
https://github.com/open-telemetry/opentelemetry-java-instrumentation/actions/runs/29557928989/job/87829898889

**Expected behavior**

`POST /stop` should return the complete generated report. Weaver should remain alive until the admin server has finished gracefully delivering the response, then exit normally.

jerbly:
Fixes: #1626

A shutdown race where `live-check --output http` could exit before the `/stop` response was fully delivered, resetting the connection on large reports. The live-check command now waits for the admin HTTP server's graceful shutdown to finish before exiting.

Added a ShutdownCoordinator which uses oneshot channels to hand the report to a waiting `/stop` request and signal once the admin server's graceful shutdown has actually finished sending it. The live-check command now blocks on that signal before returning, so the process can't exit mid-write. All the related locking is encapsulated in this one struct instead of being spread across the admin handlers and the live-check command (replacing the old AdminReportSender).

```mermaid
sequenceDiagram
    participant C as HTTP client
    participant A as Admin thread (axum /stop)
    participant M as Main thread (live-check)

    C->>A: POST /stop
    A->>M: begin_report_wait()<br/>send OtlpRequest::Stop
    Note over M: stop ingesting,<br/>build report
    M->>A: deliver_report(content_type, body)
    A-->>C: 200 OK (writing body...)
    A->>A: trigger_admin_shutdown()
    Note over A: graceful shutdown starts,<br/>stops accepting new conns
    M->>M: wait_for_admin_shutdown()<br/>(blocks)
    Note over C,A: client finishes reading<br/>the full body
    Note over A: graceful shutdown completes
    A->>M: signal_admin_shutdown_complete()
    Note over M: unblocks, returns,<br/>process exits
```


src/registry/otlp/mod.rs
```diff
@@ -40,16 +41,114 @@ use tonic::transport::Server;
...
-/// the report to be sent through the oneshot channel before responding.
-/// When `false`, `/stop` returns 200 immediately (the default).
+/// How long `/stop` waits for the report, and how long the admin server's
+/// graceful shutdown gets to finish delivering it (see
+/// [`ShutdownCoordinator::wait_for_admin_shutdown`]).
+const ADMIN_REQUEST_TIMEOUT: Duration = Duration::from_secs(60);
+
+/// A poisoned lock means another thread already panicked mid-update, so
+/// panicking here too is safer than continuing with inconsistent state.
+const LOCK_POISONED_MSG: &str = "shutdown coordinator lock poisoned";
```
jsuereth: Can we update this?

"One of the threads responsible for reporting status on shutdown has crashed.  Killing the process because weren't not sure whether our memory is still safe to use"

src/registry/otlp/mod.rs
```diff
@@ -40,16 +41,114 @@ use tonic::transport::Server;
...
 #[derive(Clone)]
-pub struct AdminReportSender {
-    pub sender: Arc<Mutex<Option<oneshot::Sender<(String, String)>>>>,
-    /// Set to `true` to have `/stop` wait and return the report as its response body.
-    pub expect_report: Arc<AtomicBool>,
+pub struct ShutdownCoordinator {
+    /// `true` when `--output http` is set: `/stop` waits for and returns
+    /// the report instead of responding immediately.
+    expect_report: Arc<AtomicBool>,
+    report_slot: Arc<Mutex<Option<oneshot::Sender<(String, String)>>>>,
```
jsuereth, line 67: Why is this an Option? So we avoid sending after we've already sent?

src/registry/otlp/mod.rs
```diff
@@ -40,16 +41,114 @@ use tonic::transport::Server;
...
+        (coordinator, shutdown_rx)
+    }
+
+    /// Set when `--output http` is in effect.
+    pub fn set_expect_report(&self, expect: bool) {
+        self.expect_report.store(expect, Ordering::Relaxed);
+    }
+
+    pub fn expects_report(&self) -> bool {
+        self.expect_report.load(Ordering::Relaxed)
```
jsuereth: We probably want this to be "Acquire" to be accurate here.

src/registry/otlp/mod.rs
```diff
@@ -40,16 +41,114 @@ use tonic::transport::Server;
...
+            .take();
+        if let Some(tx) = tx {
+            let _ = tx.send(());
+        }
+    }
+
+    /// Blocks until the admin server confirms its graceful shutdown has
+    /// finished, so the process doesn't exit mid-write. No-op if already
+    /// consumed.
+    pub fn wait_for_admin_shutdown(&self) {
```
jsuereth, line 145: I do think this may be simpler with a CondVar - but since you've encapsulated this (thank you!) we can tweak the concurrency primitives behind the scenes as long as we understand the concurrency behavior that's desired.

src/registry/otlp/mod.rs
```diff
@@ -40,16 +41,114 @@ use tonic::transport::Server;
...
+/// Coordinates delivering the live-check report to a waiting `/stop`
+/// request and confirming the admin server has finished writing it before
+/// the process exits. All locking lives here; call sites only use these
+/// methods.
 #[derive(Clone)]
-pub struct AdminReportSender {
-    pub sender: Arc<Mutex<Option<oneshot::Sender<(String, String)>>>>,
-    /// Set to `true` to have `/stop` wait and return the report as its response body.
-    pub expect_report: Arc<AtomicBool>,
+pub struct ShutdownCoordinator {
```
jsuereth, line 61: Where some of these `Option`s are produced together, should we instead keep *both* sides in a single `Arc<Mutex<Option<(Sender,Receiver)>>` so we don't have potential race conditions trying to change them?

jsuereth (approved):
Mostly minor nits / clarifications.  Overall design is much clearer what we're doing and I think isolating the concurrency handling into one struct lets us address issues over time.

The usage of `expects_report` has a potential concurrency issue without updating the ordering to be Acquire on read, given how you use it to track whether to call `begin_report_wait`.  Should be a simple fix.

src/registry/otlp/mod.rs
jerbly: correct

src/registry/otlp/mod.rs
jerbly: I want to do the bigger refactor on this anyway so this will change again.

jerbly:
@jsuereth - addressed all in [2431b8c](https://github.com/open-telemetry/weaver/pull/1632/commits/2431b8c4ba33e6461d8d2212af490c329b436d44)

PR #1634, opened 2026-07-25
fix ref_group resolution on refinements 

lmolkova:
```yaml
file_format: definition/2
attributes:
  - key: test.attr
    type: string
    brief: Test attribute
    stability: stable

metrics:
  - name: base.metric
    instrument: histogram
    unit: s
    brief: Base
    stability: stable
    attributes:
      - ref: test.attr
        requirement_level:
          conditionally_required: "on the base signal"

attribute_groups:
  - id: some.group
    visibility: internal 
    attributes:
      - ref: test.attr
        examples: ["MyExample"]

metric_refinements:
  - id: refined.metric
    ref: base.metric
    brief: Refined
    attributes:
      - ref_group: some.group
```

`test.attr` on `refined.metric`

|  | requirement_level | examples
-- | -- | --
Expected | conditionally_required | MyExample
Actual | recommended ❌ | MyExample

when we do `ref` on a refinement, we ignore missing properties, when we do `ref_group`, we use the resolved group with default values of missing properties.

### DIFF
--- a/crates/weaver_resolver/src/registry.rs
+++ b/crates/weaver_resolver/src/registry.rs
@@ -871,32 +871,28 @@ fn resolve_inheritance_attrs_unified(
     for (parent_group_id, included_group) in include_groups {
         for parent_attr in included_group.iter() {
             let attr_id = parent_attr.spec.id();
-            match inherited_attrs.get_mut(&attr_id) {
-                Some(existing) => {
-                    existing.spec = resolve_inheritance_attr(
-                        &parent_attr.spec,
-                        &existing.spec,
-                        &mut existing.lineage,
-                    );
-                    existing.lineage.source_group = parent_group_id.to_owned();
-                }
-                None => {
-                    let lineage =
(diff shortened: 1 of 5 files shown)


crates/weaver_resolver/src/registry.rs
```diff
@@ -1854,4 +1863,85 @@ groups:
...
         serde_json::to_string_pretty(value).unwrap()
     }
+
+    fn bare_ref(
+        id: &str,
+        requirement_level: Option<weaver_semconv::attribute::RequirementLevel>,
+        examples: Option<weaver_semconv::attribute::Examples>,
+    ) -> UnresolvedAttribute {
+        UnresolvedAttribute {
+            spec: weaver_semconv::attribute::AttributeSpec::Ref {
```
jerbly: super nit: could match style by using `use` imports over qualified paths here.

jerbly (approved):
LGTM.

PR #1639, opened 2026-07-27
Fix same dependency version reported as an unresolvable conflict

jerbly:
Two references to the same version need no resolution, and saying so should not depend on parsing the version.

Repro: a registry whose manifest omits `schema_url` on a dependency entry fails `weaver registry check` with `Ambiguous reference ... https://otel/unknown and https://otel/unknown`.

### DIFF
--- a/crates/weaver_resolver/src/conflict_strategy.rs
+++ b/crates/weaver_resolver/src/conflict_strategy.rs
@@ -25,6 +25,14 @@ pub(crate) struct UseLatestMajorVersion;
 
 impl DependencyVersionConflictStrategy for UseLatestMajorVersion {
     fn resolve_conflict(&self, url1: &SchemaUrl, url2: &SchemaUrl) -> Result<SchemaUrl, Error> {
+        // The same version seen twice is not a conflict, and saying so must not
+        // depend on parsing the version: a dependency declared without a
+        // `schema_url` gets a synthesized, version-less URL, which would
+        // otherwise fail `semver()` below and be reported as unresolvable
+        // against itself.
+        if url1 == url2 {
+            return Ok(url1.clone());
+        }
         if url1.name() != url2.name() {
             return Err(Error::AmbiguousReference {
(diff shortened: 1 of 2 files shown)


crates/weaver_resolver/src/conflict_strategy.rs
```diff
@@ -25,6 +25,14 @@ pub(crate) struct UseLatestMajorVersion;
 
 impl DependencyVersionConflictStrategy for UseLatestMajorVersion {
     fn resolve_conflict(&self, url1: &SchemaUrl, url2: &SchemaUrl) -> Result<SchemaUrl, Error> {
+        // The same version seen twice is not a conflict, and saying so must not
+        // depend on parsing the version: a dependency declared without a
+        // `schema_url` gets a synthesized, version-less URL, which would
```
lmolkova, line 30: > dependency declared without a `schema_url`

how do we support it? it should not be allowed

jsuereth: I think I understand this - although we should make the comments better here.

in the event I have a SchemaURL that has no version or is malformed (some of our legacy schemas) - this would allow a diamond dependency using that schema because we never try to parse the version.  If it *had* a legit version that was the same - the algorithm would have already worked.


The real fix here is NOT what's described in the PR - it's avoiding parsing the schema-url version (seen on line 43).


Effectively - we broke dependency chains with latest release and we're asking folks to start using valid SchemaURLs.

The test here, e.g. using `unknown` is an *invalid schema url*.


The open question here is if we allow that or not?

@jerbly I'd be "ok" allowing that - but I think we should be issuing a *warning* at a minimum here, telling the user that their schema-url is not valid and this will be considered a hard failure in the future.  However, I have concerns that not fixing the schema-url will just lead to other bugs elsewhere.

jerbly:
As discussed in Slack - the right course of action here is to make `schema_url` mandatory as a breaking change. In which case this PR is not needed. Closing.

PR #1640, opened 2026-07-27
Fix attributes from a dependency being attributed to the wrong registry

jerbly:
Record the defining registry's schema URL in dependency lineage.

V1 lineage encoded an attribute's dependency source by registry name only, and readers tried to recover the URL by hunting for a local group with matching provenance. A registry that merely `ref:`s an attribute has no such group, so the source fell back to that registry's own URL and the attribute reached a downstream consumer under two identities, failing with a spurious "Ambiguous reference".

The name also could not distinguish versions, so an attribute upgraded to a newer compatible version still pointed at the older one. The full schema URL is now recorded and used directly.


jerbly:
This was rushed and needs to be checked not just for correctness but whether we want support in certain scenarios.

Issue #1623, opened 2026-07-22
Weaver filters return refinements


thompson-tomo:
## Details

We should offer an optional parameter for the get signals filters to optionally also return the refinements.

## Alternatives considered

Introducing additional filters to get the refinements

## Additional context

Clarification been confirmed that they are there, just under refinements. Should we offer a filter to get all spans including refinements?

_Originally posted by @thompson-tomo in https://github.com/open-telemetry/weaver/issues/1582#issuecomment-5004825793_

Issue #1617, opened 2026-07-21
Indicate which attributes will have a constant value


thompson-tomo:
**Is your feature request related to a problem? Please describe.**
Currently we are frequently needing to be repetitively adding nearly identical notes to attributes to indicate that the value in this context is fixed. See https://github.com/open-telemetry/semantic-conventions/pull/3904 as an example.

**Describe the solution you'd like**
When we add an attribute to a signal via a ref, it should be possible to set the value of that attribute. This can be done by a new option on the ref - `constant_value`. When this is set the `example` field on the `attribute` on the `signal` in the resolved schema should only contain the constant value and no other options. It should also contain a `value_mode` field which is set to `constant`, for enums this could be set to enum.

**Describe alternatives you've considered**
Use annotations.
Calling it fixed.

**Additional context**

Within documentation we can then within the attribute row indicate which attributes will have a constant value.

<sub>**Tip**: [React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

Issue #1590, opened 2026-07-13
Don't allow to refine refinements


lmolkova:
Turns out one can write something like 

```yaml
metrics:
  - name: base.metric
    requirement_level: recommended
    instrument: histogram
    ...
metric_refinements:
  - id: metric.refined.metric 
    ref: base.metric
  - id: metric.refined2.metric
    ref: refined.metric               # !!!!!!!
```

and it will be resolved to both of them refining original metric

```yaml
refinements:
  metrics:
  - id: refined.metric
    name: base.metric
    instrument: histogram
    ...
  - id: refined2.metric
    name: base.metric
    instrument: histogram
    ...
```

but it should not be allowed.

Issue #1146, opened 2026-01-20
v2: stability should be required on enum members from JSON schema perspective


lmolkova:
See https://github.com/open-telemetry/weaver/pull/1121#discussion_r2692017832

PR #1643, opened 2026-07-27
Design doc: attribute constants, discriminator on refinements, enum refinements

lmolkova:
Related issues #1617, #803, #479, #520, #1590, #892, #878, #1146, #1623, #329

Examples:

### Pinning attribute value to constant when referencing attribute and discriminators

```yaml
spans:
  - type: gen_ai.inference.client
    kind: client
    discriminator: [gen_ai.provider.name]
    # ...

span_refinements:
  - id: openai.inference.client
    ref: gen_ai.inference.client
    attributes:
      - ref: gen_ai.provider.name
        constant: openai   # can be used independently of `discriminator` on base span definition
```

### Specifying a set of known values on a enum

```yaml
attributes:
  - key: error.type
    stability: stable
    brief: Describes a class of error the operation ended with.
    type:
      members:
        - id: other
          value: _OTHER
          brief: A fallback error value used when the instrumentation has no custom value.
          stability: stable

spans:
  - type: socket.connect
    attributes:
      - ref: error.type
        brief: Socket error code.
        type:
          members:
            - ref: other                         # explicitly list all you want to apply
            - id: network_down
              brief: The network subsystem is unavailable.
              stability: development
            - id: connection_refused
              brief: The remote host actively refused the connection.
              stability: development
            # ... 14 more
```

### Specifying a set of known values on a string  attribute

```yaml
attributes:
  - key: messaging.operation.name
    type: string
    stability: development
    brief: The system-specific name of the messaging operation.

span_refinements:
  - id: azure.servicebus.producer.create
    ref: messaging.producer.create
    attributes:
      - ref: messaging.operation.name
        brief: Azure Service Bus operation name.
        type:
          members:
            - id: send
              brief: Sends a message to a queue or topic.
              stability: development
            - id: schedule
              brief: Schedules a message for future delivery.
              stability: development
```

### DIFF
--- a/docs/specs/attribute-value-constraints/attribute_value_constraints.md
+++ b/docs/specs/attribute-value-constraints/attribute_value_constraints.md
@@ -28,7 +28,7 @@ The only way to comply is more prose. The .NET socket connect span lists its 16 
 - ref: error.type
   brief: "Socket error code."
   note: |
-    The following errors codes are reported:
+    The following error codes are reported:
 
     - `network_down`
     - `address_already_in_use`
--- a/docs/specs/attribute-value-constraints/attribute_value_constraints.md
+++ b/docs/specs/attribute-value-constraints/attribute_value_constraints.md
@@ -268,13 +268,13 @@ attributes:
           stability: stable
 
 spans:
-  - type: dotnet.socket.connect.internal
-    kind: internal
+  - type: socket.connect
     attributes:
       - ref: error.type
         brief: Socket error code.
         type:
           members:
+            - ref: other
             - id: network_down
               brief: The network subsystem is unavailable.
               stability: development
--- a/docs/specs/attribute-value-constraints/attribute_value_constraints.md
+++ b/docs/specs/attribute-value-constraints/attribute_value_constraints.md
@@ -46,8 +46,8 @@ own and can ship on its own, so each gets its own phase.
 | # | Problem | Field | Phase | Reference sites today |
 |---|---|---|---|---|
 | 1 | This attribute always has one specific value in this context | `constant` | 1 | ~110 |
-| 2 | Which refinement does an observed piece of telemetry belong to? | `discriminator` | 2 | ~63 refinements |
-| 3 | This reference draws from a limited set, though the global definition is open-ended | `values`, `type.members` | 3, 4 | ~32 |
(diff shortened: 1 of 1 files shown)


docs/specs/attribute-value-constraints/attribute_value_constraints.md
```diff
@@ -0,0 +1,326 @@
...
+              stability: development
+            - id: connection_refused
+              brief: The remote host actively refused the connection.
+              stability: development
+            # ... 14 more
+```
+
+There is no "exclude these members" form: `type.members` on a reference replaces the members of the definition.
+A new member added to `gen_ai.operation.name` must **not** silently become part of a refinement's set — someone
+has to decide which span types it belongs to.
```
thompson-tomo, line 289: It would be good to understand how `hw.state` would be modelled using this design. As That is probs one of the most complex.

docs/specs/attribute-value-constraints/attribute_value_constraints.md
```diff
@@ -0,0 +1,326 @@
...
+  Weaver cannot check that the value matches the prose it replaces — that is a review task during the migration.
+- Changing an existing `constant` value **is** a breaking change.
+- Changing which attributes discriminate a signal **is** a breaking change.
+- Narrowing or extending a known set is **not** a breaking change. The set is open, so no telemetry stops
+  conforming; only documentation and validator advice change.
+
+## Out of Scope
+
+- **Turning enums into standalone named types** shared by several attributes. Unrelated to this proposal; the
+  fields above work either way.
```
thompson-tomo, line 303: Would standalone types not change the solution for 3?

For instance we could just overwrite the type on the ref to a new type which has the same value type. That new type can either be a refinement (additional members) or a new type.

docs/specs/attribute-value-constraints/attribute_value_constraints.md
lmolkova: like this

```
spans:
  - type: gen_ai.inference.client
    kind: client
    discriminator: [gen_ai.provider.name]
    # ...

span_refinements:
  - id: openai.inference.client
    ref: gen_ai.inference.client
    attributes:
      - ref: gen_ai.provider.name
        constant: openai   # can be used independently of `discriminator` on base span definition
```

thompson-tomo: hw.state usages is different as it isn't a constant but rather a subset of the options. Some of different usages contain a different subset.

PR #1645, opened 2026-07-27
Fix live-check admin server (and /stop) shutting down 60s after startup

### LINKED ISSUE #1644: `registry live-check`: admin `/stop` endpoint dies ~60s after startup; with continuous OTLP the process hangs and never writes a report
**Describe the bug**
The live-check admin HTTP server (serving `/health` and `/stop`) is shut down roughly **60 seconds after weaver starts**, regardless of activity. After that:

- `POST /stop` gets no response (connection failure) and can no longer stop the live-check.
- If OTLP data is still arriving — so the `--inactivity-timeout` never fires — the ingestion loop never terminates, **no report is ever produced, and the process hangs indefinitely.**

**To Reproduce**

Any registry works; below uses a minimal one. Requires Docker.

```bash
# 1. minimal registry
mkdir reg && cat > reg/manifest.yaml <<'YAML'
name: minimal
schema_url: https://example.com/schemas/0.1.0
YAML
cat > reg/groups.yaml <<'YAML'
groups:
  - id: registry.demo
    type: attribute_group
    brief: demo
    attributes:
      - id: demo.attr
        type: string
        brief: demo
        stability: development
        examples: ["x"]
YAML

# 2. start live-check (file output; long inactivity timeout)
docker run -d --name wv --network host -v "$PWD/reg":/reg:ro -w /reg \
  otel/weaver:v0.25.0 registry live-check --registry /reg \
  --inactivity-timeout 300 --admin-port 4320 \
  --otlp-grpc-address 0.0.0.0 --otlp-grpc-port 4317 --output /tmp/out

# 3. stream OTLP continuously (so inactivity never fires)
docker run -d --name tg --network host \
  ghcr.io/open-telemetry/opentelemetry-collector-contrib/telemetrygen:latest \
  traces --otlp-endpoint localhost:4317 --otlp-insecure --duration 200s --rate 50

# 4. wait past 60s, then POST /stop while data is still flowing
sleep 90
curl -v -X POST http://localhost:4320/stop        # -> connection failure, no response
docker inspect -f '{{.State.Running}}' wv          # -> true (still running)
docker logs wv | grep graceful                     # -> "graceful shutdown did not complete within 60s"
# weaver never exits and /tmp/out/*.json is never written
```

Contrast: `POST /stop` **within** the first 60s (or with telemetry that stops, so the inactivity timeout eventually fires) works and produces a report.

**Expected behavior**
- **Expected:** `/stop` responds and stops the live-check at any point during the session; the 60s bound applies only to the graceful-shutdown drain (or per request), not the admin server's whole lifetime.
- **Actual:** the admin server (and `/stop`) is gone ~60s after startup; a later `/stop` cannot stop the run, and under continuous OTLP the process hangs forever with no report.

**Additional context**
`otel/weaver:v0.25.0` (digest `sha256:bef6000b4a4be46f81242f9ee785e0ebf0604606c15f92cb54a59893a741ec0c`).

NimrodAvni78:
Fixes #1644 

`registry live-check`'s admin HTTP server (serving `/health` and `/stop`) is torn
down roughly **60 seconds after startup**, regardless of activity. After that,
`POST /stop` gets no response and can no longer stop the live-check. If OTLP data
is still arriving — so `--inactivity-timeout` never fires — the ingestion loop
never terminates, no report is ever written, and the process hangs indefinitely.

This is a regression introduced in #1632: bounding the *whole* admin server in a
60s timeout means any live-check session longer than a minute (instrument an app,
exercise it, then `/stop`) can't be stopped via `/stop`.

## Root cause

In `spawn_http_admin_handler`, the entire server future was wrapped in the request
timeout:

```rust
let result = tokio::time::timeout(ADMIN_REQUEST_TIMEOUT, async {
    axum::serve(listener, app)
        .with_graceful_shutdown(async { let _ = admin_shutdown_rx.await; })
        .await
})
.await;
```

`tokio::time::timeout(60s, serve_future)` bounds the *accept* phase, not just the
graceful-shutdown drain — so the server (and `/stop`) is dropped 60s after it
starts. The intent (per the `ADMIN_REQUEST_TIMEOUT` doc comment — "how long the
admin server's graceful shutdown gets to finish delivering it") was to bound only
the drain.

### DIFF
--- a/src/registry/otlp/mod.rs
+++ b/src/registry/otlp/mod.rs
@@ -525,20 +525,32 @@ async fn spawn_http_admin_handler(
                 .with_state(state);
 
             let _ = tasks.spawn(async move {
-                // Bounded so a stalled client can't hang the CLI forever.
-                let result = tokio::time::timeout(ADMIN_REQUEST_TIMEOUT, async {
-                    axum::serve(listener, app)
-                        .with_graceful_shutdown(async {
-                            let _ = admin_shutdown_rx.await;
-                        })
-                        .await
-                })
-                .await;
+                // ADMIN_REQUEST_TIMEOUT bounds only the graceful-shutdown drain
+                // (after the stop signal), not the server's whole lifetime.
+                // Wrapping the entire `serve` in the timeout would tear the admin
+                // server (and `/stop`) down that long after startup, so a `/stop`
+                // sent later in a long-running live-check could never stop it.
+                let (drain_started_tx, drain_started_rx) = oneshot::channel();
+                let serve = axum::serve(listener, app).with_graceful_shutdown(async move {
+                    let _ = admin_shutdown_rx.await;
+                    let _ = drain_started_tx.send(());
+                });
 
-                match result {
-                    Ok(Ok(())) => {}
-                    Ok(Err(e)) => warn!("Admin HTTP server error: {e}"),
-                    Err(_) => warn!(
+                let drain_deadline = async {
+                    if drain_started_rx.await.is_ok() {
+                        sleep(ADMIN_REQUEST_TIMEOUT).await;
+                    } else {
+                        std::future::pending::<()>().await;
+                    }
+                };
+
+                tokio::select! {
+                    result = serve => {
+                        if let Err(e) = result {
+                            warn!("Admin HTTP server error: {e}");
+                        }
+                    }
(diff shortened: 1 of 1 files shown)


jerbly (approved):
Oof! Thanks.

PR #1646, opened 2026-07-27
Failing tests: dependency provenance and mandatory schema_url

jerbly:
Three intentionally-red tests in `weaver_resolver` pinning down expected behavior. Each fails until the corresponding fix lands — the CI failure on this PR is intentional.

### 1. Imported attributes must keep their origin provenance

`main` imports a metric from `base` and one from `middle`; both use `host.id`, which is defined only in `base` (a dependency of `middle`). Middle's re-export is stamped with middle's schema_url instead of base's, so resolution fails with `AmbiguousReference`. Fixture: `crates/weaver_resolver/data/correct-provenance/`.

```
cargo test -p weaver_resolver test_imported_attribute_keeps_origin_provenance
```

### 2. Upgraded groups must carry the winning version's definition and provenance

The `compatible-version-conflict` fixture resolves C v1.1 vs v1.2 in favour of v1.2 for the attribute catalog, but `metric.c.metric.1` keeps v1.1's body and provenance — the resolved schema is internally inconsistent.

```
cargo test -p weaver_resolver test_upgraded_group_keeps_winning_version_provenance
```

### 3. `schema_url` is mandatory for dependencies

A dependency declared with only `name` + `registry_path` currently gets a synthesized schema URL with version `unknown` and resolves silently. It must instead fail with a useful error naming the offending dependency. Fixture: `crates/weaver_resolver/data/mandatory-schema-url/`.

```
cargo test -p weaver_resolver test_dependency_without_schema_url_is_rejected
```

### DIFF
--- a/crates/weaver_resolver/data/correct-provenance/base/manifest.yaml
+++ b/crates/weaver_resolver/data/correct-provenance/base/manifest.yaml
@@ -1,0 +1,3 @@
+name: base
+description: Registry defining the `host.id` attribute and a metric that uses it.
+schema_url: https://example.com/base/1.0.0
--- a/crates/weaver_resolver/data/correct-provenance/base/semconv.yaml
+++ b/crates/weaver_resolver/data/correct-provenance/base/semconv.yaml
@@ -1,0 +1,15 @@
+file_format: definition/2
+attributes:
+  - key: host.id
+    type: string
+    stability: stable
+    brief: Unique host ID.
+    examples: ["abc123"]
+metrics:
+  - name: base.uptime
+    brief: Uptime metric defined in base.
+    instrument: gauge
+    unit: "s"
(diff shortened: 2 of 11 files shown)


jsuereth:
>  Middle's re-export is stamped with middle's schema_url instead of base's, so resolution fails with

This is a bug - I assume this happened from a previous release before we had better multi-dependency resolution?

jsuereth:
100% agree 2 + 3 are our bugs that must be fixed.

For 1 - it's an interesting, temporary, problem.  Fixing the `include_unrefereenced` flag to remember provenance *should* fix this issue once published repositories upstream are fixed.  If we have a fix - I think it should be a temporary issue.

jerbly:
Fixes in 0.25.1 release

PR #1649, opened 2026-07-28
Fix imported attributes losing their origin registry provenance

jerbly:
Fixes case 1 of #1646 (`test_imported_attribute_keeps_origin_provenance`, included here with its fixture).

## Problem

With a dependency chain `main -> middle -> base`, where `middle` defines a metric referencing an attribute defined only in `base`, resolving `main` (which imports one metric from each dependency) fails:

```
Ambiguous reference 'host.id' found in multiple dependencies with different
SchemaURLs: https://example.com/base/1.0.0 and https://example.com/middle/0.1.0
```

`middle`'s resolved schema correctly records the attribute's origin (`source_group: v2_dependency.example.com/base`), but when `main` consumes it, the resolver recovers the full origin schema URL by scanning `middle`'s resolved groups for one carrying `base`'s provenance. If `base` contributed no whole group to `middle` — only an attribute — the scan misses and the attribute is stamped with `middle`'s own schema URL. The same attribute then reaches `main` under two different registry names, which conflict resolution rejects as ambiguous.

## Fix

Recover the origin schema URL from the resolved schema's `dependencies` list first (matching by registry name, exactly as the v1→v2 conversion in `weaver_resolved_schema` already does), and only fall back to the group-provenance scan. Applied to both lookup sites: `find_attribute_source` in `dependency.rs` and `AttributeLookup for V1Schema` in `attribute.rs`.

## Testing

- New fixture `crates/weaver_resolver/data/correct-provenance/` modeling the shape above.
- `cargo test -p weaver_resolver test_imported_attribute_keeps_origin_provenance` — red before this fix, green after.
- Full workspace test suite passes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ All notable changes to this project will be documented in this file.
 
 # Unreleased
 
+- Fix imported attributes losing their origin registry provenance ([#1649](https://github.com/open-telemetry/weaver/issues/1649))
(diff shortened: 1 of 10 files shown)


lmolkova (approved):

jsuereth (approved):
Fix looks good, thanks for catching this bug.  Looks like an opportunity to refacor/clean up the logic after a point fix release!

PR #1650, opened 2026-07-28
Fix imported groups keeping the losing version's definition and provenance

jerbly:
Fixes case 2 of #1646 (`test_upgraded_group_keeps_winning_version_provenance`), covering both the definition (V1) and published (resolved V2) dependency import paths.

## Problem

In the `compatible-version-conflict` fixture, `main` imports `c.metric.1` via A (which pins C v1.1) and `c.span.2` via B (which pins C v1.2). Version conflict resolution correctly chooses C v1.2 — but only for the attribute catalog. Resolving `main` produces an internally inconsistent schema:

```yaml
- id: metric.c.metric.1
  brief: Metric 1 in C v1.1            # losing version's body
  attributes:
  - name: c.attr1
    brief: Attribute 1 from C v1.2 (updated)   # upgraded attribute
  lineage:
    provenance:
      schema_url: https://example.com/c/1.1.0   # losing version's provenance
```

`resolve_dependency_imports` in `registry.rs` already relabels the stored provenance URL to the winning version, but the group's own body and `lineage` — which is what the resolved output reports — are cloned unchanged from the intermediate dependency's stale copy.

The same inconsistency exists when the intermediate dependency is published as a `resolved/2.0` schema: `V2Schema::import_groups` builds the group from the published registry's embedded (stale) copy while the attributes get upgraded. The new `compatible-version-conflict-published` fixture demonstrates this: registry A is a published resolved V2 schema whose embedded `c.metric.1` was resolved against C v1.1.

## Fix

A shared `upgrade_imported_group` helper, used by both `V1Schema::import_groups` and `V2Schema::import_groups`: when an imported group's origin registry (from its lineage provenance, which may point at a transitive dependency) has a chosen newer compatible version in the resolution cache, the group's definition is taken from the chosen schema — body, lineage, and attribute refs registered consistently with the upgraded attribute catalog. This mirrors what `upgrade_attribute_with_source` already does for attributes. If the chosen registry no longer defines the group, the existing copy is kept (previous behavior).

Note: the definition and published import paths disagree on group id prefixes (`metric.c.metric.1` vs `c.metric.1`), so groups are matched in the chosen registry by type and name rather than raw id. A side benefit is that upgraded groups imported through the published path now come out with the canonical prefixed id and a real provenance path.

## Testing

- `cargo test -p weaver_resolver test_upgraded_group_keeps_winning_version_provenance` — V1 path, red before, green after.
- `cargo test -p weaver_resolver test_upgraded_group_keeps_winning_version_provenance_published` — published V2 path with the new fixture, red before, green after.
- Resolved output for both fixtures now reports the v1.2 body, v1.2 provenance, and v1.2 attributes consistently.
- Full workspace test suite passes (899 tests).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,8 @@ All notable changes to this project will be documented in this file.
 
 # Unreleased
 
-- Fix imported attributes losing their origin registry provenance ([#1649](https://github.com/open-telemetry/weaver/issues/1649))
+- Fix imported groups keeping the losing version's definition and provenance ([#1650](https://github.com/open-telemetry/weaver/issues/1650) by @jerbly)
+- Fix imported attributes losing their origin registry provenance ([#1649](https://github.com/open-telemetry/weaver/issues/1649) by @jerbly)
 - Fix signals imported from a dependency losing their per-signal attribute data. When several signals reference the same attribute with different `requirement_level` or `role`, each imported signal was re-pointed at whichever variant of the attribute was registered first. E.g. silently rewriting requirement levels or dropping `role: identifying` from imported entities. Each signal now references the attribute variant it actually declares. Per-name conflict resolution still applies to root-attribute provenance. ([#1635](https://github.com/open-telemetry/weaver/pull/1635) by @jerbly)
 
 # [0.25.0] - 2026-07-24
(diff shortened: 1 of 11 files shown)


jsuereth (approved):
Fix is good but lots of duplicated code or repeated logic.

Given urgency of the bugs we can clean this up post submission.

This is also one of those "duh" moments - we know attributes can change across version and treating the whole group as a version is a good thing

PR #1651, opened 2026-07-28
Make schema_url mandatory for manifest dependencies

jerbly:
Fixes case 3 of #1646 (`test_dependency_without_schema_url_is_rejected`, included here with its fixture).

## Problem

A manifest dependency declared with only `name` (and optionally `registry_path`) was silently accepted: a schema URL was synthesized with version `unknown`, and resolution proceeded without complaint. The `schema_url` is the identity that provenance tracking and version conflict resolution key on, so a fabricated version quietly undermines both.

## Fix

Reject such dependencies at manifest deserialization (`Dependency`'s `Deserialize` impl in `weaver_semconv/src/manifest.rs`) with an error that names the offending dependency, explains what the `schema_url` is for, and shows an example value. The console output:

```
× The registry manifest at "data/mandatory-schema-url/main/manifest.yaml" is
│ invalid. dependencies: missing required field 'schema_url' for dependency
│ 'dep'. The schema_url uniquely identifies the dependency registry and its
│ version, e.g. https://example.com/dep/1.0.0 at line 5 column 3
```

The `name` field is still accepted alongside `schema_url` (it remains informational); only the name-only form is rejected. This removes the last consumer of the deprecated name-only fallback:

- `test_dependency_deserialize_with_deprecated_name` is replaced by `test_dependency_deserialize_name_only_is_rejected`.
- The `circular-registry-test` fixture used the name-only form; it now declares a proper `schema_url` so it keeps exercising circular dependency detection (verified: it still fails with the circular dependency error).

## Testing

- New fixture `crates/weaver_resolver/data/mandatory-schema-url/` with a `name`-only dependency.
- `cargo test -p weaver_resolver test_dependency_without_schema_url_is_rejected` — red before this fix (resolution succeeded silently), green after. The test asserts the error message contains both `schema_url` and the dependency name, and accepts the error surfacing at either manifest load or resolution.
- Full workspace test suite passes (898 tests).

### DIFF
--- a/crates/weaver_semconv/src/manifest.rs
+++ b/crates/weaver_semconv/src/manifest.rs
@@ -128,11 +128,18 @@ impl<'de> Deserialize<'de> for Dependency {
 
         let schema_url = match (helper.schema_url, helper.name) {
             (Some(url), _) => url,
-            (None, Some(name)) => SchemaUrl::try_from_name_version(&name, "unknown")
-                .map_err(serde::de::Error::custom)?,
+            (None, Some(name)) => {
+                return Err(serde::de::Error::custom(format!(
+                    "missing required field 'schema_url' for dependency '{name}'. \
+                     The schema_url uniquely identifies the dependency registry and its \
+                     version, e.g. https://example.com/{name}/1.0.0"
+                )))
+            }
             (None, None) => {
                 return Err(serde::de::Error::custom(
-                    "Either 'schema_url' or 'name' must be provided for a dependency",
+                    "missing required field 'schema_url' for a dependency. \
+                     The schema_url uniquely identifies the dependency registry and its \
+                     version, e.g. https://example.com/my-registry/1.0.0",
                 ))
             }
         };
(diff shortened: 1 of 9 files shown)


crates/weaver_semconv/src/manifest.rs
```diff
@@ -128,11 +128,18 @@ impl<'de> Deserialize<'de> for Dependency {
...
+            }
             (None, None) => {
                 return Err(serde::de::Error::custom(
-                    "Either 'schema_url' or 'name' must be provided for a dependency",
+                    "missing required field 'schema_url' for a dependency. \
+                     The schema_url uniquely identifies the dependency registry and its \
+                     version, e.g. https://example.com/my-registry/1.0.0",
                 ))
             }
         };
```
lmolkova, line 145: it seems we can just drop name at this point, if someone provides name and no schema url we fail anyway.
Name is deprecated and is ignored regardless.

We can provably update DependencyHelper on line 121 to have schema_url required and either remove name (in breaking manner) or at least remove the first error with name

jsuereth (approved):
Another thing we should have fixed!  Good catch

PR #1655, opened 2026-07-29
Fix transitive dependency provenance

jerbly:
Fix elements inherited from a transitive dependency being reported as locally defined. A resolved schema's `dependencies` set is the table that `DependencyRef` provenance indexes into, but it listed only direct dependencies, so anything reaching the registry through a dependency-of-a-dependency had no entry to point at. It now records the full closure.


lmolkova (approved):
LGTM!

PR #1656, opened 2026-07-29
Fix: serve imported attributes

jerbly:
Update - found the same issue in `stats` and critically `live-check`:

Fix attributes imported from a dependency being treated as if they were not part of the registry. Such an attribute is only ever referenced by the signals that use it, never listed as a definition, so `weaver registry stats` counted it nowhere, `weaver serve` left it out of search and detail pages, and `weaver registry live-check --v2` reported it as a `missing_attribute` violation and excluded it from registry coverage. All three now consider every attribute the registry can reach.

----

Attributes imported from a dependency are inlined into the signals that use them and never listed in `registry.attributes`. `weaver serve` indexed only that list, so those attributes were unsearchable, had no detail page, and the attribute links on every imported signal's page dead-ended on "Attribute not found". Registry stats counted the same list, so the attributes tile showed 0 while the API was serving them.

Search, detail pages and stats now cover every attribute the registry can reach. `weaver registry stats` is unchanged.

Repro, from `crates/weaver_resolver`:
```
    weaver serve --v2 -r data/transitive-provenance/main --skip-policies
```
`/api/v1/registry/attribute/base.host.id` returns 404 before and 200 after.
`/api/v1/registry/search?type=attribute` reports 0 attributes before and 1 after.

Follow-up, not in this PR: a `ref:` can override `brief`/`note`, so an attribute can have several variants. Where the attribute is imported there is no definition to prefer, and the detail page in the UI shows whichever variant is encountered first. A follow-up should surface all variants and where each is used. Note the variants can share a schema URL, they are distinguished by the referencing signal, not by origin registry.

### DIFF
--- a/crates/weaver_forge/src/v2/registry.rs
+++ b/crates/weaver_forge/src/v2/registry.rs
@@ -55,6 +55,44 @@ pub struct Registry {
     pub entities: Vec<Entity>,
 }
 
+impl Registry {
+    /// Every attribute reachable from this registry: the definitions first, then
+    /// the copy inlined into each signal that uses one.
+    ///
+    /// Attributes imported from a dependency only ever exist inlined. A key can
+    /// appear several times; callers keep the first for one entry per key.
+    pub fn all_attributes(&self) -> impl Iterator<Item = &Attribute> {
+        self.attributes
+            .iter()
+            .chain(
+                self.attribute_groups
+                    .iter()
+                    .flat_map(|g| g.attributes.iter().map(|a| &a.base)),
+            )
+            .chain(
+                self.metrics
+                    .iter()
+                    .flat_map(|m| m.attributes.iter().map(|a| &a.base)),
+            )
+            .chain(
+                self.spans
+                    .iter()
+                    .flat_map(|s| s.attributes.iter().map(|a| &a.base)),
+            )
+            .chain(
+                self.events
+                    .iter()
+                    .flat_map(|e| e.attributes.iter().map(|a| &a.base)),
+            )
+            .chain(self.entities.iter().flat_map(|e| {
+                e.identity
+                    .iter()
+                    .chain(e.description.iter())
+                    .map(|a| &a.base)
+            }))
+    }
+}
+
 /// The set of all refinements for a semantic convention registry.
 ///
 /// A refinement is a specialization of a signal for a particular purpose,
--- a/crates/weaver_forge/src/v2/registry.rs
+++ b/crates/weaver_forge/src/v2/registry.rs
@@ -55,44 +55,6 @@ pub struct Registry {
     pub entities: Vec<Entity>,
 }
 
-impl Registry {
-    /// Every attribute reachable from this registry: the definitions first, then
-    /// the copy inlined into each signal that uses one.
-    ///
-    /// Attributes imported from a dependency only ever exist inlined. A key can
-    /// appear several times; keep the first for one entry per key.
-    pub fn all_attributes(&self) -> impl Iterator<Item = &Attribute> {
-        self.attributes
-            .iter()
-            .chain(
-                self.attribute_groups
-                    .iter()
-                    .flat_map(|g| g.attributes.iter().map(|a| &a.base)),
-            )
-            .chain(
-                self.metrics
-                    .iter()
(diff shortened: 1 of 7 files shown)


crates/weaver_forge/src/v2/registry.rs
```diff
@@ -55,6 +55,44 @@ pub struct Registry {
     pub entities: Vec<Entity>,
 }
 
+impl Registry {
+    /// Every attribute reachable from this registry: the definitions first, then
+    /// the copy inlined into each signal that uses one.
+    ///
+    /// Attributes imported from a dependency only ever exist inlined. A key can
+    /// appear several times; keep the first for one entry per key.
+    pub fn all_attributes(&self) -> impl Iterator<Item = &Attribute> {
```
lmolkova: it seems it's only used in serve, can we make it a private package there? it seems everything is public and accessible there already

crates/weaver_resolved_schema/src/v2/registry.rs
```diff
@@ -50,6 +50,58 @@ pub struct Registry {
 }
 
 impl Registry {
+    /// Every attribute reference in this registry: its own definitions first,
+    /// then one per signal attribute. A single attribute yields several
+    /// references, since each signal keeps its own variant.
+    pub fn attribute_refs(&self) -> impl Iterator<Item = &AttributeRef> {
```
lmolkova: it's only used in `reachable_attributes` and tests, does it need to be public?

crates/weaver_resolved_schema/src/v2/registry.rs
```diff
@@ -50,6 +50,58 @@ pub struct Registry {
...
+            }))
+    }
+
+    /// Every distinct attribute the registry can reach, de-duplicated by key.
+    ///
+    /// A superset of `attributes`: an attribute imported from a dependency is
+    /// referenced by signals but never listed as a definition. Definitions come
+    /// first, so they win de-duplication.
+    #[must_use]
+    pub fn reachable_attributes<T: AttributeCatalog>(&self, catalog: &T) -> Vec<AttributeRef> {
```
lmolkova: it's only used in serve crate and it uses public properties already exposed in this registry, do we need to have this method in this crate?

The reason for these comments is that it seems this logic is only needed for serve - resolution and forge have some other means to serialize everything, but it adds more confusion to already complicated API we  have here. having private method in serve makes it much easier for anyone (like me) reading these code later and wondering why do we need this for resolution

crates/weaver_forge/src/v2/registry.rs
jerbly: It's used by search, so I moved this from forge to there in https://github.com/open-telemetry/weaver/pull/1656/commits/6d08ca8c0ca34aa463992359b1d3ca18e944e7ff

crates/weaver_resolved_schema/src/v2/registry.rs
jerbly: Reduced scope to private here: https://github.com/open-telemetry/weaver/pull/1656/commits/75cf66d60a7cf45fba2168cb3a0690bd91d8b6a9

crates/weaver_resolved_schema/src/v2/registry.rs
jerbly: Reduced scope to private: https://github.com/open-telemetry/weaver/pull/1656/commits/75cf66d60a7cf45fba2168cb3a0690bd91d8b6a9

It's used by stats as well and in fact stats had the same bug that it wasn't counting imported attributes. In terms of API these are functions over the registries so it kinda feels like it belongs in the registry crates. I did move the forge one as requested though because that one is only used in one place.

crates/weaver_forge/src/v2/registry.rs
jerbly: I found the same issue in live-check which is actually the worst case! So this is now used in two places so I've put it back into the forge crate. https://github.com/open-telemetry/weaver/pull/1656/commits/49030e0cd4df03fd51b141edb6d99e5785b649c8

jerbly:
Closing this PR since the problem is deeper than this surface level. See: https://github.com/open-telemetry/weaver/issues/1658

PR #1665, opened 2026-08-04
Live-check: multi-registry v2 attribute lookup

### LINKED ISSUE #1658: Attribute Resolution Issues
# Attribute Resolution Issues

Three things noticed so far:

- Live-check - showing `missing_attribute` for an attribute that is imported from another registry, even though the attribute is present in the resolved registry.
- Stats - counts of attributes don't match expectations. This could be explained as "the count of locally defined attributes" vs "the count of all attributes in scope for this registry", but it is not clear which is the correct interpretation. If it's the latter, then the count should include imported attributes, but it does not.
- Serve - Links to attribute details can return 404.

## Live-check

TL;DR: Live-check needs attribute definitions independent of the signals that use them. There's currently no way to **import** an attribute from another registry but you can make an attribute_group that **refs** all the attributes you need. Even then, no referenced attribute will be present in `registry.attributes` in the resolved registry, so live-check will not find it. So, we need either an explicit way to import attributes to be used by live-check, or an implicit way to include all attributes that are in scope for this registry in the resolved registry.

---

Take this Fixture:

`main` defines nothing. It imports one metric from `base`, and that metric is the only route by which the `host.id` attribute reaches `main`.

```yaml
# tests/imported_attribute/base/manifest.yaml
name: base
description: Registry defining `host.id` and the metric that uses it.
schema_url: https://example.com/base/1.0.0

# tests/imported_attribute/base/semconv.yaml
file_format: definition/2
attributes:
  - key: host.id
    type: string
    stability: stable
    brief: Unique host ID.
    examples: ["abc123"]
metrics:
  - name: base.uptime
    brief: Uptime metric defined in base.
    instrument: gauge
    unit: "s"
    stability: stable
    attributes:
      - ref: host.id

# tests/imported_attribute/main/manifest.yaml
name: main
description: Registry importing `base.uptime`; its `host.id` attribute is defined only in base.
schema_url: https://example.com/main/0.1.0
dependencies:
  - name: base
    schema_url: https://example.com/base/1.0.0
    registry_path: tests/imported_attribute/base

# tests/imported_attribute/main/imports.yaml
file_format: definition/2
imports:
  metrics:
    - base.uptime
```

resolves to forge form with: `weaver registry resolve -r tests/imported_attribute/main --v2`

```yaml
schema_url: https://example.com/main/0.1.0
registry:
  attributes: []
  attribute_groups: []
  metrics:
    - name: base.uptime
      instrument: gauge
      unit: s
      attributes:
        - key: host.id
          type: string
          examples:
            - abc123
          brief: Unique host ID.
          stability: stable
          provenance:
            source: https://example.com/base/1.0.0
      brief: Uptime metric defined in base.
      stability: stable
      provenance:
        source: https://example.com/base/1.0.0
        path: tests/imported_attribute/base/semconv.yaml
  spans: []
  events: []
  entities: []
refinements:
  metrics:
    - id: base.uptime
      name: base.uptime
      instrument: gauge
      unit: s
      attributes:
        - key: host.id
          type: string
          examples:
            - abc123
          brief: Unique host ID.
          stability: stable
          provenance:
            source: https://example.com/base/1.0.0
      brief: Uptime metric defined in base.
      stability: stable
      provenance:
        source: https://example.com/base/1.0.0
        path: tests/imported_attribute/base/semconv.yaml
  spans: []
  events: []
  entities: []
```

resolves to package form with: `weaver registry package -r tests/imported_attribute/main --v2 -o <dir> --resolved-registry-uri https://example.com/main/0.1.0`

```yaml
file_format: resolved/2.0
schema_url: https://example.com/main/0.1.0
attribute_catalog:
  - key: host.id
    type: string
    examples:
      - abc123
    brief: Unique host ID.
    stability: stable
    provenance:
      source: 0
registry:
  attributes: []
  attribute_groups: []
  spans: []
  metrics:
    - name: base.uptime
      instrument: gauge
      unit: s
      attributes:
        - base: 0
      brief: Uptime metric defined in base.
      stability: stable
      provenance:
        source: 0
  events: []
  entities: []
refinements:
  spans: []
  metrics:
    - id: base.uptime
      name: base.uptime
      instrument: gauge
      unit: s
      attributes:
        - base: 0
      brief: Uptime metric defined in base.
      stability: stable
      provenance:
        source: 0
  events: []
  entities: []
dependencies:
  - https://example.com/base/1.0.0
```

**OBSERAVTION**: In both forms the unmodified imported metric shows as a refinement??

So in both forms `host.id` is present and fully described — inlined into the metric in forge form, and in the `attribute_catalog` (reached via `base: 0`) in package form. What is empty in both is `registry.attributes`.

Perhaps `registry.attributes` means "attributes this registry **defines**", not "attributes this registry **has**"? Every consumer that reads it as the latter gets the wrong answer for an imported attribute.

Live-check today looks in `registry.attributes` to find the definition of an attribute. If it is not there, it is considered undefined.

Advice is given on attributes independently of the signal they appear in. This is important because:

- not all signals are typed, e.g. Spans, Logs, Resources, InstrumentationScope.
- The signal name may be incorrect or that signal my not be defined in a registry in an incoming sample but we'd still like advice on the attributes used.
- There may be additional attributes in the sample that are not part of the signal, and we still want advice on them. (**TODO**: Should live-check have a finding: `unexpected_attribute` for attributes that are not part of a matched signal?).

**FIX 1**: Live-check should use the inline attribute definition when the signal is typed and that type has been matched to a definition.

If we need to explicitly import every attribute we want to use, then the `imports.yaml` should be updated to include `host.id`:

```yaml
imports:
  metrics:
    - base.uptime
  attributes:
    - host.id
```

Today this is invalid, there's no `attributes` variant to `imports`. There's no way to import an attribute declared the way `host.id` is declared in `base`.

`attribute_groups` is allowed in `imports`, but that suffers the same issue as `metric` or any signal. Those attributes do not land in `registry.attributes`.

Perhaps what we need for live-check, in the case where an attribute is outside of a matched signal, is a list of "root attributes" - the original definitions of the attributes that are in scope for this registry?

**FIX 2 - OPTION 1**: Add `attribute_origins` to the resolved registry form, and populate it with the definitions of all attributes that are in scope for this registry. This would allow live-check to find the definitions of imported attributes even if they are not part of a matched signal. This is a newly required field for `resolved/2.0` registries since the origin definitions are erased in `attribute_catalog`. Any use of attributes already in the `attribute_catalog` could only be somewhat random.

**FIX 2 - OPTION 2**: No new field. Fix the `attribute_catalog` instead.

An attribute can only be defined in one place - the `attributes` section - and there can only be one definition per key. Every `ref:` points at that one definition and refines it _for that signal only_. So the catalog should hold one entry per key: the definition. A signal's changes belong on the ref that made them.

The schema already works this way for two fields. `requirement_level` and `sampling_relevant` sit on the ref, because the syntax says they can only be set there. But the six fields a ref _can_ refine - `brief`, `note`, `examples`, `annotations`, `stability`, `deprecated` - get merged into a brand new catalog entry instead. That merge is what creates the variants, and what loses the definition.

Move those six onto the ref as well:

```yaml
attribute_catalog:
  - key: a.attr
    brief: CANONICAL brief from the definition. # one entry - the definition
metrics:
  - name: m.zzz
    attributes:
      - base: 0
        brief: OVERRIDDEN by m.zzz # the refinement stays here
```

Then `attribute_catalog` _is_ the list of origin definitions. Nothing to mark, nothing to add, and the catalog gets smaller rather than bigger. Live-check looks up a key and gets the definition; if it matched a signal it also has that signal's overrides.

jerbly:
Fixes #1658 

Fix 1 - always use the matched signal's attribute definitions first
Fix 2 - use the dependency registries to find attribute definitions when https://github.com/open-telemetry/weaver/pull/1671 lands

In the PR there is a new doc `attribute_lookups.md` which explains in detail, with an example, the lookup strategy.

In a follow-up PR I will:
* Add an option to disable dependency searches
* Add a new `unexpected_attribute` finding
* Add `schema_url` field on every finding


crates/weaver_live_check/src/lib.rs
```diff
@@ -114,6 +117,47 @@ pub enum VersionedRegistry {
...
+    /// Only a dependency of the registry under check defines the attribute.
+    Dependency,
+}
+
+/// A registry and its dependency closure, nearest first.
+///
+/// Depth-first: the registry, then each dependency, then that dependency's own
+/// dependencies.
+#[must_use]
+pub fn registries_nearest_first(registry: &ForgeResolvedRegistry) -> Vec<&ForgeResolvedRegistry> {
```
lmolkova, line 136: Live-check defines its own cross-dependency resolution order here, but it's  different from the resolver's.

- **Live-check**: DFS preorder, first write wins. But `ForgeResolvedRegistry.dependencies` comes from `ResolvedTelemetrySchema.dependencies`, a `BTreeSet<SchemaUrl>` of the *transitive* closure, so the root's children are already every dependency, sorted by URL string. Depth carries no distance, and first-wins resolves to whichever schema URL sorts first alphabetically.
- **Resolver**, same key via `ref:` (`weaver_resolver/src/attribute.rs:538`): collects every match, arbitrates with `resolve_conflict` (definition beats reference, then `UseLatestMajorVersion`), raises `AmbiguousReference` rather than guessing.

They also disagree on what counts as a closure source: this walk offers public attribute-group references, `V2Schema::lookup_attribute` (`attribute.rs:680`) reads `registry.attributes` only.

Ideally there should be just one implementation. It's probably difficult because of v1 hop, so if it's too much we can do a short term solution:

1. Match the resolver's order and sources exactly. 
2. Have a test pinning the agreement so they don't diverge. 
3. File an issue to collapse them once the v1 hop is gone

crates/weaver_live_check/docs/attribute_lookup.md
```diff
@@ -0,0 +1,375 @@
...
+
+The pairs are strictly ordered, which is easy to get wrong in both directions:
+
+- A template in the registry under check beats an exact key that only a
+  dependency defines. Exact beats template only within a pair, and step 4 runs
+  before step 5.
+- A shorter template in the registry under check beats a longer one in a
+  dependency. Length only decides between templates found in the same step.
+
+Within the dependency closure, nearer wins: the closure is walked
```
lmolkova: "nearer wins ... walked nearest-first" isn't what happens. The closure is flat and URL-sorted, so the tree has no proximity to walk. Same claim at line 163.

defaults/jq/advice.jq
```diff
@@ -1,14 +1,24 @@
...
-{
+#
+# The live checker keeps the registry under check apart from its dependency
+# closure, because the attribute lookup searches the registry first. The policies
+# only ask whether a definition exists anywhere, so the two are merged here.
+# The registry under check is the right operand of `+`, so its definition wins
+# on a key that both sides hold. That matches the lookup, which searches the
+# registry under check before any dependency.
+. as $root
+| (($root.dependency_attributes // {}) + $root.semconv_attributes) as $attributes
```
lmolkova: nit: worth a CHANGELOG line. Merging the dependency closure into `attributes_set` makes `illegal_namespace` (a violation, so non-zero exit) fire on keys only a dependency defines. `probe.a.b.child=x` against `lookup_precedence/local` reports it now and didn't before.

crates/weaver_live_check/src/lib.rs
jsuereth: I actually have a different view here -

Live-check should check against the schema-url used in the telemetry, exactly.   As an option - if I *override* schema-url or say, use my local source instead of the one found at location X, then that's on me to make sure the resolution override works.

lmolkova: sorry for AI comment, the point I was making that live check should not have individual logic that decides how to walk dependency tree.

jerbly: So, I tried to spell this all out in the document. Essentially live-check checks in this order:
1. if the sample matches a signal in the local forge registry, use the attribute there
2. if the sample does not match a signal, or if it does but is an attribute not found in the signal definition, use the locally defined loose attributes (these are locally defined attributes + public attribute groups + span attributes)
3. if not found in 1 or 2 above then go look in the dependencies just in case there's a matching attribute somewhere.

This is for --v2 only. v1 just kinda randomly picks attributes from the forge registry but does gain (1) above.

(2) is, I believe, the least surprising way for live-check to work with --v2
(3) is a NEW feature, a follow-up PR will make it clear in the findings that an attribute was found in your dependencies and not in your local forge registry.

This has uncovered a fault in the way we make the dependency tree in forge now, which I reviewed and didn't see! 

If I have a chain dependency like c->b->a you get this in forge:
```
c
--a
--b
  --a
```
a is duplicated and brought to the top. I have a PR to change live-check to a BFS but we need to fix the tree structure too.

lmolkova: > 1. if the sample matches a signal in the local forge registry, use the attribute there

👍 

> 2. if the sample does not match a signal, or if it does but is an attribute not found in the signal definition, use the locally defined loose attributes (these are locally defined attributes + public attribute groups + span attributes)

I think it should be just "these are locally defined attributes". Ignore attribute groups and span attributes - there are multiple copies of them and it's error prone to use a random copy.

> 3. if not found in 1 or 2 above then go look in the dependencies just in case there's a matching attribute somewhere.

👍  with caveat:

This logic for 2 and 3 is essentially is implemented by resolution: can we reuse or make lookup attribute method that does 2 and 3 and reuse it in live check? so that when we look up attribute to resolve a signal (not refinement) and when we look it up in live check we always end up with the same one?

Or maybe if we exclude attribute refs (public groups and span attributes) we can rely on definition uniqueness? Registry won't be resolved if attribute is defined in more than on registry in the tree and then order and logic duplication is an internal minor code organization problem that won't affect users and then is totally ok to proceed with

jerbly: > I think it should be just "these are locally defined attributes". Ignore attribute groups and span attributes - there are multiple copies of them and it's error prone to use a random copy.

So, all attributes and all spans would only be checked against the local definition (not refinement) and then we rely on the search in the dependencies to find any other definition. It seems a shame to count something locally defined as a span attribute or public group as a "found this in a dependency".

lmolkova: The problem is that we don't know if a random span we find is a right scope. We'll fix it with span type and might figure something out for public groups. And I think it's least confusing to get original definition on unrecognized signal

jerbly: I think, ultimately, we're going to need some config options for different use cases.

In my case I would miss being able to refine a loose/public attribute through an attribute_group in order to add an annotation for codegen and for live-check. For example, if I have this setup: `app<--core<--otel` I'd like to add annotations to some attributes defined in `core` that are used in spans or logs or scope or resource-on-untyped-signals etc. I can do this for my codegen but not live-check? Now I have asymmetry from the code I've produced and how I'm able to test it.

So perhaps we need usage modes?
* `strict` - only live-check matched-signals and their associated entities
* `local` - live-check against everything defined, referenced and imported to the local registry
* `tree` - as `local` but also search down the dependency tree to surface matches that you may wish to bring to the surface.

lmolkova: should we invent some way to map to an arbitrary attribute group in live-check? this would be helpful for resources and spans until they get proper type. 

With any approach except strict, there is nothing that guarantees there is one refinement per attribute, i.e. the live-check behavior becomes undefined - it depends on the order of attributes / dependencies.

Another approach - for loose attributes do strict AND in the custom advice list all refinements attr can match (local and from dependencies) with provenance.

jerbly: > With any approach except strict, there is nothing that guarantees there is one refinement per attribute, i.e. the live-check behavior becomes undefined - it depends on the order of attributes / dependencies.

This is the current released v1 behaviour.

How about, for `local`:
* Ignore spans until we have a span_type.
* Take attributes from **public** visibility attribute_groups only, to use in unmatched areas. These public groups could also have been imported. Warn of any conflict and explain how the winner is picked. (If you want to make a flat module of constants for attribute names via codegen you have this issue there too).
* Search order becomes:
  * Matched Signals
  * Public attributes with conflicts resolved
  * Locally defined attributes

`tree` does all the same but tacks on dependencies.

lmolkova: I think this still leads to temporary workaround for a fundamental problem that there are multiple refs and that we can only validate signals that are documented. I think we should push users to document their signals rather than pretending we can validate without it.

Returning a list of matched refs to custom rego still allows users to validate undocumented signals, but asks them to do more work for it. There users can report warnings on conflict, match appropriate ref based on provenance, etc.

jerbly: I think that leaves us with `strict` as the default. `local` goes away, and `tree` searches in dependencies for original definitions only. In the interests of getting some v2 support I'll implement that in this PR and come back to other ways to improve maybe outside the schema.

`strict` means all logs, spans, instrumentation scopes, resources and undefined metrics and events will show `missing_attribute` for every attribute.

lmolkova: > strict means all logs, spans, instrumentation scopes, resources and undefined metrics and events will show missing_attribute for every attribute.

not sure I understand - why so?

lmolkova: maybe we define strict differently?

My proposal is to match attributes on unrecognized signals to original definitions (local or visible in dependencies). ~~We should probably issue an improvement advice when signal is not matched (for metrics, events, and entities)~~ and also improvement when attribute matched is not imported (not sure about this one)

jerbly: I feel like this is coming back around in a circle to be _close_ to what I've implemented 🙃 - as soon as we start caring about what has been imported/reffed into the local schema we might as well figure out how to use that nicely. And, if I've gone to the trouble to import (via attribute_group because there's no import for attribute) why can't I have my annotations?

lmolkova: because you can import that same attribute multiple times and there is nothing that would stop you. But there is always just one original definition.

jerbly: I've put this as a topic for discussion in the meeting tomorrow so we can come to some decision. Unfortunately the lax nature of v1 means we'll be losing some functionality in the stricter v2 and I'm just trying to find some compromise way to not lose that AND be more correct.

jerbly:
Replaced by #1721

PR #1666, opened 2026-08-04
fix(resolver): load registry files when the root path is dot-prefixed

### LINKED ISSUE #1558: Explicit `imports` resolve to zero against a dependency, but `--include-unreferenced` works
## Summary

The docs give two ways to pull signals and attributes from a dependency registry (e.g. semconv) into a custom registry: explicit `imports:` in a `definition/2` file, or the `--include-unreferenced` flag. The changelog says they're the same thing:

> Change `--include-unreferenced` so that this is the same as creating a set of `import:` statements in the registry manifest. (#1442)

and the resolver pushes you toward the explicit form:

> The flag `include_unreferenced` is deprecated. Please prefer manually adding the required imports to your schema files in the future.

They aren't the same. With a normal "depend on semconv from git" setup, explicit `imports` resolve to zero groups and live-check flags every standard attribute as `missing_attribute`. The deprecated flag, on the identical registry, resolves the full set and passes. So the documented, non-deprecated path silently imports nothing.

## Reproduction (weaver 0.22.1)

`registry_manifest.yaml`
```yaml
schema_url: https://example.com/schemas/1.0.0
dependencies:
  - schema_url: https://opentelemetry.io/schemas/1.41.1
    registry_path: https://github.com/open-telemetry/semantic-conventions.git@v1.41.1[model]
```

`main.yaml`
```yaml
file_format: definition/2
imports:
  attribute_groups: ["*"]
  spans: ["*"]
  metrics: ["*"]
```

`sample.txt`
```
http.request.method
url.path
client.address
```

Explicit imports (import everything with `"*"`):
```
$ weaver registry live-check -r . --input-source sample.txt --input-format text
    - [violation] Attribute 'http.request.method' does not exist in the registry.
    - [violation] Attribute 'url.path' does not exist in the registry.
    - [violation] Attribute 'client.address' does not exist in the registry.
```

Same registry, add the flag:
```
$ weaver registry live-check -r . --include-unreferenced --input-source sample.txt --input-format text
✔ Performed live check for registry `.`
```

`weaver registry resolve` tells the same story: `0` groups with imports, `935` with `--include-unreferenced`.

I tried every token I could think of (`registry.*`, `attributes.*`, `span.*`, bare namespaces, `"*"`), in both the manifest and a `definition/2` file, with a `schema_url`-matched dependency. All import nothing.

## Notes

A couple of things I found while digging, in case it's more than one problem:

- Imports do work when the dependency is a published (pre-resolved) registry rather than a raw model. But packaging semconv with `weaver registry package --v2` produces `registry.attribute_groups: []` (the `registry.*` groups get flattened into the attribute catalog), and the packaged spans/metrics have no `id` for the spans matcher to match, so even that route imports nothing usable from semconv.
- The doc examples don't line up with real semconv ids. `semconv-syntax.v2.md` shows `spans: - http.client` and `attribute_groups: - client.*`, but the source ids are `registry.*` and the resolved ids are `attributes.*` / `span.*`. Neither matches the examples.

## Environment

weaver 0.22.1 on macOS, dependency `semantic-conventions@v1.41.1[model]` (also reproduces on `main`).

## Ask

Either make explicit `imports` resolve against dependencies the way `--include-unreferenced` does (restoring the documented equivalence), or update the docs and the deprecation note to match what works today. Happy to share the full repro.

NimrodAvni78:
Fixes https://github.com/open-telemetry/weaver/issues/1558

The directory walker pruned the registry root when its final path component started with a dot (e.g. `-r .`, or a `.`-named registry dir), because walkdir yields the root as an entry whose file_name is "." and is_hidden matched it. Exempt the root (depth 0) from the hidden filter so a dot-rooted registry loads its own spec files while genuinely hidden sub-entries stay skipped.


lmolkova:
@NimrodAvni78 thanks for the contribution! And sorry we didn't look at it earlier.

Could you please share some context on the scenario?

allowing hidden directories would have some unexpected side-effects like `.git` being included.

Is there a need to keep registry files in hidden dirs? And having it allowed only in the root seems inconsistent.

PR #1669, opened 2026-08-05
 resolver: fix provenance and resolution of attributes inherited from dependencies

lmolkova:
Related to https://github.com/open-telemetry/weaver/issues/1658

1. Provenance was lost on attributes inherited through a refinement

```diff
# b/main.yaml — b depends on a, refines a's metric
metric_refinements:
  - id: b.refined
    ref: a.metric
    brief: refined

  # b's resolved.yaml
  attribute_catalog:
  - key: foo
    brief: the foo
+   provenance:
+     source: 0      # → a
```

2. A bare ref resolved against the refinement's own copy

```diff
# c/main.yaml — refines b.metric, and has its own metric
metric_refinements:
  - id: c.refined
    ref: b.metric
    attributes:
      - ref: foo
        brief: foo as used by c     # signal-local wording
metrics:
  - name: c.metric
    attributes:
      - ref: foo                    # unrelated metric

  - name: c.metric
    attributes:
-   - base: 0        # brief: foo as used by c
+   - base: 1        # brief: the foo
```

3. A dependency re-exported an attribute it never defined, as its own

```diff
# b/main.yaml — b defines nothing, only refines
metric_refinements:
  - id: b.refined
    ref: a.metric

# c/main.yaml — c depends on b only
metrics:
  - name: c.metric
    attributes:
      - ref: foo

  # c's resolved.yaml
- - key: foo
-   provenance:
-     source: 1      # ← b, but a defines it
+ × attribute reference is not resolved: foo
```

Case 3 only happened when the dependency was consumed from source; a published one already refused. Now both behave the same.

### DIFF
--- a/crates/weaver_resolved_schema/src/attribute.rs
+++ b/crates/weaver_resolved_schema/src/attribute.rs
@@ -17,6 +17,7 @@ use weaver_semconv::attribute::{
     AttributeRole, AttributeSpec, AttributeType, Examples, RequirementLevel,
 };
 use weaver_semconv::deprecated::Deprecated;
+use weaver_semconv::schema_url::SchemaUrl;
 use weaver_semconv::stability::Stability;
 use weaver_semconv::YamlValue;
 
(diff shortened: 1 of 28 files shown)


jsuereth (approved):
"Catalog" for v1 is something we  need to kill pretty soon - it's fraught with peril, but this change looks good! Nice cleanup, doesn't touch the hairy methods, sorry it took so long to review it.

PR #1670, opened 2026-08-05
Disallow stability and deprecated on v2 attribute refs

lmolkova:
These always come from the attribute definition. Using them on a ref is now a failure.

Related to #1658 

- we don't have tests for it
- we don't have a use-case for it
- since we don't always have original attribute definition and only know ref properties, we might be limited in weaver live checks

We can add them in some other form that does not hide original deprecation and stability.


lmolkova:
Superseded by #1720 (recreated from a fork branch).

PR #1676, opened 2026-08-06
Drop refinements from forge v1

lmolkova:
refinements defined in v2 translate into a duplicate metric in v1 forge.
We didn't catch it in semconv because we switched to v2 policies, but it's breaking codegen - see https://github.com/open-telemetry/semantic-conventions-java/pull/549#discussion_r3716151890

```yaml
file_format: definition/2
attributes:
  - key: demo.type
    type: string
    stability: development
    brief: "Component type."
metrics:
  - name: demo.errors
    stability: development
    brief: "Number of errors."
    instrument: counter
    unit: "{error}"
    attributes: [{ref: demo.type, requirement_level: required}]
metric_refinements:
  - id: metric.demo.disk.errors
    ref: demo.errors
    brief: "Number of errors on this disk."
    attributes: [{ref: demo.type, note: "MUST be set to `disk`."}]
```

Resolved output of semconv_grouped_metrics — two entries, same metric_name:

```
{"id":"metric.demo.disk.errors","metric_name":"demo.errors","type":"metric","brief":"Number of errors on this disk.","unit":"{error}","instrument":"counter"}
{"id":"metric.demo.errors",     "metric_name":"demo.errors","type":"metric","brief":"Number of errors.",           "unit":"{error}","instrument":"counter"}
```

while we could resolve refinements with `group.type == metric_refinement`, I think it's reasonable to drop them from v1 forge completely and ask users who want refinements to use --v2 to get them in resolved/forge

### DIFF
--- a/crates/weaver_forge/src/registry.rs
+++ b/crates/weaver_forge/src/registry.rs
@@ -194,6 +194,11 @@ impl ResolvedGroup {
 
 impl ResolvedRegistry {
     /// Create a new template registry from a resolved registry.
+    ///
+    /// V2 refinements are dropped: v1 has no notion of a refinement, and one resolves
+    /// into an extra group carrying the same signal name as the signal it refines, so
+    /// consumers selecting groups by type see it as a duplicate signal. Consumers that
+    /// need refinements use the v2 registry, where they are modelled explicitly.
     pub fn try_from_resolved_registry(
         registry: &Registry,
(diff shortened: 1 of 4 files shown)


joaopgrassi (approved):

jsuereth (approved):
This is a good workaround for now, since these cannot exist in pure v1 usage - THanks for tracking this down!

PR #1684, opened 2026-08-10
Entity dependencies

jerbly:
## Summary of what the resolver does now

1. **An entity association pulls the entity in.** Name an entity in `entity_associations` and it arrives, the way an attribute named in a `ref` already did. Code gen, doc gen and live-check can see it. The copy is the definition from the registry that declared it, unchanged, and it keeps that registry as its source.
2. **An association that nothing satisfies fails the resolve.** A name that no registry in scope defines is an error, not silence.
3. **A private entity stays private.** An entity marked `dependency_resolution.exclude: true` cannot be reached from a dependent, by an `imports` clause or by an association.
4. **One definition reached by two paths is one entity.** A diamond in the dependency graph no longer produces duplicate copies and duplicate warnings.
5. **An import that matches nothing is reported.** For every signal type, not only entities. Resolution still succeeds, with a warning.
6. **A resolved schema points at the entity instead of naming it.** `entity_associations` holds an index into the entity refinements, so a consumer gets a pointer rather than a name to look up again. Templates still read the name.

### DIFF
--- a/crates/weaver_resolved_schema/src/v2/mod.rs
+++ b/crates/weaver_resolved_schema/src/v2/mod.rs
@@ -153,12 +153,56 @@ impl TryFrom<crate::ResolvedTelemetrySchema> for ResolvedTelemetrySchema {
     }
 }
 
-fn fix_group_id(prefix: &'static str, group_id: &str) -> SignalId {
-    if group_id.starts_with(prefix) {
-        group_id.trim_start_matches(prefix).to_owned().into()
-    } else {
-        group_id.to_owned().into()
+/// Turns the names in an association expression into entity refinement
+/// indices. The shape of the expression does not change, only the leaves.
+fn convert_entity_associations(
+    associations: &[weaver_semconv::entity_association::EntityAssociation],
+    entity_refs: &HashMap<SignalId, entity::EntityRef>,
+    group_id: &str,
+) -> Result<Vec<entity::EntityAssociation>, crate::error::Error> {
+    associations
+        .iter()
+        .map(|assoc| convert_entity_association(assoc, entity_refs, group_id))
+        .collect()
+}
+
+fn convert_entity_association(
+    association: &weaver_semconv::entity_association::EntityAssociation,
+    entity_refs: &HashMap<SignalId, entity::EntityRef>,
+    group_id: &str,
+) -> Result<entity::EntityAssociation, crate::error::Error> {
+    use weaver_semconv::entity_association::EntityAssociation as SpecAssociation;
+    match association {
+        SpecAssociation::Ref(name) => entity_refs
+            .get(&SignalId::from(name.clone()))
+            .map(|r| entity::EntityAssociation::Ref(*r))
+            .ok_or_else(|| crate::error::Error::EntityAssociationNotFound {
+                group_id: group_id.to_owned(),
+                entity_type: name.clone(),
+            }),
+        SpecAssociation::OneOf { one_of } => Ok(entity::EntityAssociation::OneOf {
+            one_of: convert_entity_associations(one_of, entity_refs, group_id)?,
+        }),
+        SpecAssociation::AllOf { all_of } => Ok(entity::EntityAssociation::AllOf {
+            all_of: convert_entity_associations(all_of, entity_refs, group_id)?,
+        }),
     }
+}
+
+/// Strips the group-type prefix that a v2 definition adds to a group id.
+///
+/// A v2 definition names a signal by type alone, and the v1 group model holds
+/// every signal type in one flat id space, so reading a v2 file mints an id
+/// like `entity.host`. This undoes that.
+///
+/// Strips one prefix only. `trim_start_matches` would strip a repeat, so
+/// `entity.entity.host` would lose both.
+fn fix_group_id(prefix: &'static str, group_id: &str) -> SignalId {
+    group_id
+        .strip_prefix(prefix)
+        .unwrap_or(group_id)
+        .to_owned()
+        .into()
 }
 
 fn fix_span_group_id(group_id: &str) -> SignalId {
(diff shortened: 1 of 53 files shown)


schemas/semconv.resolved.v2.json
```diff
@@ -470,6 +470,12 @@
         "requirement_level"
       ]
     },
+    "EntityRef": {
+      "description": "A reference, by index, to an entity refinement of the schema.",
```
lmolkova, line 474: nit, please update md doc too

crates/weaver_semconv_gen/src/v2.rs
```diff
@@ -333,19 +385,37 @@ fn lookup_id(registry: &ResolvedTelemetrySchema, id: &str) -> Result<Option<Reso
...
             .refinements
             .spans
             .iter()
             .find(|s| s.id == id)
-            .map(|s| resolved_span(&s.span, &registry.attribute_catalog))),
+            .map(|s| {
+                resolved_span(
+                    &s.span,
+                    &registry.attribute_catalog,
+                    &registry.refinements.entities,
```
lmolkova: would it make sense to pass the whole registry to this and similar methods? sometime down the road I could imagine more cross-linking in the future like: 

- spans listing which events cover exception for this operation (useful today) or which events could record this span start (future needs)
- spans documenting which metric records duration of this span

crates/weaver_resolver/src/registry.rs
```diff
@@ -422,6 +452,109 @@ fn resolve_dependency_imports<C: crate::SchemaCacheLookup>(
...
+        .flat_map(|g| g.group.entity_associations.iter())
+        .flat_map(|assoc| assoc.referenced_entities())
+        .map(|name| name.to_owned())
+        .unique()
+        .collect()
+}
+
+/// The entity types that the registry defines.
+///
+/// An entity carries its type in `name`. A v1 group may leave `name` unset, so
```
lmolkova: is it the case today? we should probably fix it in semconv create by failing fast and assume it can't happen here

crates/weaver_resolver/src/dependency.rs
```diff
@@ -109,6 +110,205 @@ impl ResolvedDependency {
     }
 }
 
+/// Whether the resolver added this import itself, rather than a user naming
+/// the groups. A bulk import must skip an excluded group in silence, because
+/// nobody asked for that group by name.
+fn is_implicit_import(import: &ImportsWithProvenance) -> bool {
+    import.provenance.path == "--include-unreferenced"
```
lmolkova, line 117: unrelated to this PR, this gave me a big pause before I realized this is our existing hacky way:)

crates/weaver_resolver/src/registry.rs
```diff
@@ -384,11 +393,28 @@ fn resolve_dependency_imports<C: crate::SchemaCacheLookup>(
     ureg: &mut UnresolvedRegistry,
     attribute_catalog: &mut AttributeCatalog,
     cache_lookup: &C,
-) -> Result<(), Error> {
+) -> Result<Vec<Error>, Error> {
     // Import from our dependencies, and add to the final registry.
-    let imports = &ureg.imports;
+    let imports = std::mem::take(&mut ureg.imports);
+    let result = import_into_registry(ureg, &imports, attribute_catalog, cache_lookup)
```
lmolkova: nit:

```suggestion
    let nfes = import_into_registry(ureg, &imports, attribute_catalog, cache_lookup)
```

crates/weaver_resolver/src/registry.rs
```diff
@@ -384,11 +393,28 @@ fn resolve_dependency_imports<C: crate::SchemaCacheLookup>(
     ureg: &mut UnresolvedRegistry,
     attribute_catalog: &mut AttributeCatalog,
     cache_lookup: &C,
-) -> Result<(), Error> {
+) -> Result<Vec<Error>, Error> {
```
lmolkova: i think in other places we pass non-fatal errors a mutable vector in parameters.

lmolkova (commented):
a few minor comments, looks great in general

crates/weaver_resolver/src/registry.rs
jerbly: fixed in [f78a805](https://github.com/open-telemetry/weaver/pull/1684/commits/f78a805e2422c33deb98343f40076a1dd59508b7)

crates/weaver_resolver/src/registry.rs
jerbly: fixed in [f78a805](https://github.com/open-telemetry/weaver/pull/1684/commits/f78a805e2422c33deb98343f40076a1dd59508b7)

crates/weaver_semconv_gen/src/v2.rs
jerbly: fixed in [f78a805](https://github.com/open-telemetry/weaver/pull/1684/commits/f78a805e2422c33deb98343f40076a1dd59508b7)

schemas/semconv.resolved.v2.json
jerbly: fixed in [f78a805](https://github.com/open-telemetry/weaver/pull/1684/commits/f78a805e2422c33deb98343f40076a1dd59508b7)

crates/weaver_resolver/src/registry.rs
jerbly: The comment isn't quite accurate. This is actually about the `resource` type. [`resource-browser.yaml`](https://github.com/open-telemetry/weaver/blob/main/crates/weaver_resolver/data/registry-test-6-resources/registry/resource-browser.yaml#L2-L5)
```yaml
  - id: browser
    # NO NAME
    stability: stable
    type: resource
    ....
```
`type: resource` isn't in the group-type enum of [`semconv.schema.json`](https://github.com/open-telemetry/weaver/blob/main/schemas/semconv.schema.json), but weaver still parses it as an alias for `entity`.

I think we should consider tidying up `resource` type in another PR.

jerbly: I corrected the comment and add a test for this case.

schemas/semconv-schemas.md
```diff
@@ -367,6 +372,30 @@ It is either a simple string or an object carrying an explanation:
...
+How a leaf names its entity depends on the schema:
+
+- In the [definition schema](#definition-schema) a leaf is an entity type, for example `service`.
+- In the [resolved schema](#resolved-schema) a leaf is an index into the `refinements.entities` array.
+  Every entity definition in `registry.entities` also has a base refinement, so an index can reach a
+  definition or a refinement of one. An association is resolved during
+  [`weaver registry package`](/docs/usage.md#weaver-registry-package): the named entity is imported
+  from the registry that defines it, and an association that no registry in scope satisfies is an error.
+- In the [materialized schema](#materialized-resolved-schema) a leaf is the id of the entity refinement
+  the index points at.
```
lmolkova, line 397: is this right? in materialized schema we populate actual objects instead of indexes for attributes and dependencies, so we should be populating the whole entity refinement here

crates/weaver_resolver/src/registry.rs
lmolkova: i'd just leave resource for now, at some point we'll deprecate v1 and resource will go away with it.

schemas/semconv-schemas.md
jerbly: Since entities are currently immutable, I left the existing string reference representation after resolution.

lmolkova: wouldn't it be a breaking change to make them immutable if we decide to?

Also, assuming i want to check metric along with associated entities (or generate a dashboard from metric and expected entities), wouldn't i need to do some join shenanigans to join one with another?

crates/weaver_resolved_schema/src/v2/mod.rs
```diff
@@ -258,6 +302,88 @@ pub fn convert_v1_to_v2(
...
+                    }
+                    _ => {
+                        desc_attrs.push(entity::EntityAttributeRef {
+                            base: a,
+                            requirement_level: attr.requirement_level.clone(),
+                        });
+                    }
+                }
+            } else {
+                // TODO logic error!
```
lquerel: Shouldn't we at least display an error message (or, better yet, return an error) when the attribute cannot be resolved?

lquerel (commented):
IMO, there is one error case that should be handled.
Other than that, LGTM.

crates/weaver_resolver/src/dependency.rs
```diff
@@ -109,6 +110,205 @@ impl ResolvedDependency {
...
+        keys.extend(signal_name);
+        // An attribute group also answers to its id without the `registry.`
+        // prefix, which is the older spelling.
+        if g.r#type == GroupType::AttributeGroup {
+            keys.extend(g.id.strip_prefix("registry."));
+        }
+        keys.extend(g.id.strip_prefix(prefix));
+        keys
+    } else {
+        match g.r#type {
```
jsuereth, line 207: Why only these for v1?  Should we do Entity + Span/Event in these too?

crates/weaver_resolved_schema/src/v2/mod.rs
jerbly: That logic error TODO was copied from the other signal blocks, there are five other spots that we need to resolve. I haven't expanded this PR further to cover those.

Turns out it isn't only a logic error: `convert_ref` compared `stability` with `unwrap_or(false)` while the catalog is built with `stability.unwrap_or(Alpha)`, so any attribute without `stability` was silently dropped.

Fixed the lookup so the default applies on both sides, and the entity loop now returns `AttributeNotFound` instead.

I've left the other five existing TODOs.

See: [1e589bb](https://github.com/open-telemetry/weaver/pull/1684/commits/1e589bbc3ec9e407f673c6b16c0b52992f96b9bf)

jerbly:
Decision from the SIG meeting. Putting this back into Draft, breaking down into smaller scoped PRs.

jerbly:
Closing this original PR as it grew too huge. Replaced with a series tracked here: #1692

PR #1693, opened 2026-08-12
Add Resolve config for schema_url overrides

jsuereth:
Helps solve migration friction when moving from V1 where `schema_url` was not a hard requirement to V2 where it is.

This exposes a `[resolver]` section in `.weaver.toml` that can be used to express where to find a specific `schema_url` in the current run of weaver.


jerbly (approved):
Nice. That will be useful.

Issue #1692, opened 2026-08-12
Entity Resolution Tracking Issue


jerbly:
An umbrella issue for a series of small pull requests. Each PR is separately reviewable and separately mergeable. Together they make `entity_associations` resolve, and make the result reachable from code generation, documentation, policies and live-check.

This replaces the single large branch in #1684, which is back in draft.

## The problem

A signal declares the entities it belongs to with `entity_associations`. Today the association is a bare string, and nothing resolves it.

1. A name that no registry defines resolves without a word. A typo survives into the published registry.
2. An entity that a dependency defines is not in the resolved schema. A consumer that must read the definition cannot find it.
3. Every consumer fails in silence. Live-check treats an unknown reference as satisfied. A template gets a string that it cannot resolve. A rego policy gets no entities at all.
4. An `imports` pattern that matches nothing is also silent, so the workaround for problem 2, fails silently.

## The series

Order matters where a PR says so. Each PR gets its own tests.

- [x] **PR 1 #1694** One list of the names an `imports` pattern can match, for every matcher. One entity from a diamond in the dependency graph, not two. A v1 entity answers to its id as well as its name, because a legacy `type: resource` group has no name and cannot be imported at all today. _Blocks PR 3 and PR 4, which both need the shared keys._
- [x] **PR 2 #1695** A signal does not lose an attribute in silence. A missing stability gets the same `Alpha` default in the catalog lookup as it gets when the catalog stores the attribute. The entity conversion returns an error instead of an empty `else` branch. _No dependencies._
- [x] **PR 3 #1701** (`Fix #1` in the notes) A typo in `imports` is visible. Every explicit pattern that named none of the imported groups is reported, at warning severity. Resolution still succeeds. _Needs PR 1._
- [x] **PR 4 #1704** (Part 1 + Part 2) An entry that nothing satisfies fails the resolve. An entry that resolves says where: the leaf becomes `EntityRef { r#type, provenance }`, and it names an entity type or the id of a refinement. A private entity, marked `dependency_resolution.exclude: true`, is out of reach by every route. This is the only schema change in the series, and it bumps the resolved file format. _Needs PR 1._
[...]

PR #1694, opened 2026-08-13
Share the import match keys, and deduplicate imported groups

jerbly:
PR1 for #1692 

The rules for what an `imports` pattern can match a group by were written out twice in `dependency.rs`, once for the explicit matcher and once for the bulk one. This PR puts them in one place. `import_match_keys` returns the strings a pattern can match a group by, and `ImportMatchers` holds one compiled `GlobSet` per `imports` field. The extraction is behaviour-preserving for every group type but one.

That one is the legacy `type: resource` entity, which could never be imported from a dependency. Such a group sets no `name` and holds its entity type in the group id, and the matcher looked only at the name. The pattern therefore had nothing to match, and the import failed in silence. The group id is now a match key too.

The second fix is in the join of the per-dependency results, which kept every copy. A definition that two paths through the dependency graph reach arrived twice, and `check_uniqueness` reported a duplicate that named one location for both copies. The join now keeps the first copy of each definition, bucketed by origin url and group type.

Two tests cover the fixes, and each one fails when its half of the change is removed.

This is the first of a series of small PRs for entity association resolution. The later ones need these shared keys to find the entities that a dependency exports.


crates/weaver_resolver/src/dependency.rs
```diff
@@ -109,6 +110,142 @@ impl ResolvedDependency {
     }
 }
 
+/// The field of an `imports` block that lists a signal type.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub(crate) enum ImportField {
```
jsuereth: Nit: (since you have to resolve merge conflicts anyway) Should we move this into an `imports.rs` or `import_matchers.rs` file?

Importing is starting to become A LOT more complex and may be good to treat it as its own "concept" isolated from dependency resolution.

jerbly: This refactor is in now: [f1d494d](https://github.com/open-telemetry/weaver/pull/1694/commits/f1d494dd6596d5d6bb1f916197c6736ab5335307)

crates/weaver_resolver/src/imports.rs
```diff
@@ -0,0 +1,1155 @@
...
+            GroupType::AttributeGroup => (None, "attribute_group."),
+            GroupType::Span => (name, "span."),
+            GroupType::Event => (name, "event."),
+            GroupType::Metric | GroupType::MetricGroup => (metric_name, "metric."),
+            GroupType::Entity => (name, "entity."),
+            GroupType::Scope | GroupType::Undefined => return vec![],
+        };
+        let mut keys = vec![g.id.as_str()];
+        keys.extend(signal_name);
+        // A pattern can also name an attribute group by its id without the
```
jsuereth, line 99: I wonder if we should tag these kinds of things as "v1 compatibility" features - we may want to *flag* when we see this as a warning to users eventually to get them to move to V2.

crates/weaver_resolver/src/imports.rs
```diff
@@ -0,0 +1,1155 @@
...
+        let mut result: Vec<GroupWithProvenance> = vec![];
+        let mut buckets: HashMap<(String, GroupType), Vec<usize>> = HashMap::new();
+        for dependency in self {
+            for candidate in dependency.import_groups(imports, attribute_catalog, cache_lookup)? {
+                let key = (
+                    origin_url(&candidate.group, &candidate.schema_url).to_string(),
+                    candidate.group.r#type.clone(),
+                );
+                let bucket = buckets.entry(key).or_default();
+                let is_duplicate = bucket
```
jsuereth, line 835: I think in this case we can still have the situation where two *unrelated* dependencies declare a group with the same "id" or "name" and we try to import both and cause a conflict.

Is that covered anywhere in the tests?

jsuereth (commented):
Overall looks like a solid cleanup - One possibly bug - one nit.

crates/weaver_resolver/src/imports.rs
jerbly: I added some TODOs for this: [fd65199](https://github.com/open-telemetry/weaver/pull/1694/commits/fd651998490657883a81e0d534222a98513dc737)

crates/weaver_resolver/src/imports.rs
jerbly: I believe the code is doing the right thing when there are rival definitions available to a top level import. Added a fixture and test to cover this: [fd65199](https://github.com/open-telemetry/weaver/pull/1694/commits/fd651998490657883a81e0d534222a98513dc737)

jsuereth (approved):

PR #1695, opened 2026-08-13
Fix attributes with no stability being dropped in the v1 to v2 conversion

jerbly:
`Catalog::convert_ref` looks up a v1 attribute in the v2 catalog by comparing every field, and the stability clause read `.map(|s| a.common.stability == *s).unwrap_or(false)`. An attribute with no `stability` therefore never matched. The catalog itself is built with `stability.unwrap_or(Alpha)`, so the entry was there all along — the lookup just could not reach it. This PR applies the same `Alpha` default on the lookup side.

A missing stability is only a warning, so registries like this resolve and reach the conversion. Every signal loop calls `convert_ref` and skips the attribute when it returns `None`, so spans, events, metrics and attribute groups all lost it in silence. Entities are where it shows: an entity that loses an identity attribute has no identity, and `registry package --v2` shipped `identity: []` with a zero exit code.

The entity loop also hit an `else { // TODO logic error! }`, which is now a `convert_attribute_ref` helper returning `AttributeNotFound`.

This is the second of a series of small PRs for entity association resolution, and it is independent of the others.

### DIFF
--- a/crates/weaver_resolved_schema/src/v2/catalog.rs
+++ b/crates/weaver_resolved_schema/src/v2/catalog.rs
@@ -69,14 +69,14 @@ impl Catalog {
                         && a.common.note == attribute.note
                         && a.common.deprecated == attribute.deprecated
                         // A v1 attribute can have no stability. The catalog
-                        // stores such an attribute as `Alpha`, so this
-                        // comparison applies the same default. The comparison
-                        // of the annotations below does the same.
+                        // stores such an attribute with the default stability,
+                        // so this comparison applies the same default. The
+                        // comparison of the annotations below does the same.
                         && a.common.stability
                             == *attribute
                                 .stability
                                 .as_ref()
-                                .unwrap_or(&weaver_semconv::stability::Stability::Alpha)
+                                .unwrap_or(&weaver_semconv::stability::Stability::default())
                         && attribute
                             .annotations
                             .as_ref()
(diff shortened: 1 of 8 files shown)


crates/weaver_resolved_schema/src/v2/catalog.rs
```diff
@@ -68,11 +68,15 @@ impl Catalog {
                         && a.common.brief == attribute.brief
                         && a.common.note == attribute.note
                         && a.common.deprecated == attribute.deprecated
-                        && attribute
-                            .stability
-                            .as_ref()
-                            .map(|s| a.common.stability == *s)
-                            .unwrap_or(false)
+                        // A v1 attribute can have no stability. The catalog
```
jsuereth, line 71: Good catch!

jsuereth (approved):

crates/weaver_resolved_schema/src/v2/catalog.rs
```diff
@@ -68,11 +68,15 @@ impl Catalog {
...
+                        // stores such an attribute as `Alpha`, so this
+                        // comparison applies the same default. The comparison
+                        // of the annotations below does the same.
+                        && a.common.stability
+                            == *attribute
+                                .stability
+                                .as_ref()
+                                .unwrap_or(&weaver_semconv::stability::Stability::Alpha)
```
lmolkova: nit, 


```suggestion
                                .unwrap_or(&weaver_semconv::stability::Stability::Development)
```

(even before alpha, this is where all semconv start)

lmolkova (approved):

crates/weaver_resolved_schema/src/v2/catalog.rs
jerbly: This was more than a one line change because this was used in the comparison so it need to change on both sides of that. But, since development is the default this turns out neater anyway. See [2347282](https://github.com/open-telemetry/weaver/pull/1695/commits/234728294c7f69883f4ee5cc60468687077af15c)

PR #1696, opened 2026-08-13
Restore support for dependencies declared by  name  in legacy (v1) manifests

lmolkova:
replace https://github.com/open-telemetry/weaver/pull/1690

fix regression introduced in 0.25.x where we started rejecting definition manifests where dependencies don't list schema_url. this is right behavior for v2, but a breaking change for v1. fixing for v1.

### DIFF
--- a/crates/weaver_resolver/src/loader.rs
+++ b/crates/weaver_resolver/src/loader.rs
@@ -13,6 +13,7 @@ use walkdir::DirEntry;
 use weaver_common::result::WResult;
 use weaver_resolved_schema::v2::ResolvedTelemetrySchema as V2Schema;
 use weaver_resolved_schema::ResolvedTelemetrySchema as V1Schema;
+use weaver_semconv::manifest::Dependency;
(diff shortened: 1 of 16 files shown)


crates/weaver_resolver/src/loader.rs
```diff
@@ -537,13 +533,17 @@ fn load_definition_repository(
...
-    prev_schema_url: &SchemaUrl,
-    schema_url: &SchemaUrl,
-) -> Result<(), Error> {
-    let _ = UseLatestMajorVersion.resolve_conflict(prev_schema_url, schema_url)?;
+/// Checks that two declarations of the same registry can be reconciled.
+///
+/// An unversioned declaration has nothing to compare against, so it must be the only one for
+/// its registry.
+fn check_version_compatibility(prev: &Dependency, dep: &Dependency) -> Result<(), Error> {
+    if !prev.is_versioned() || !dep.is_versioned() {
```
jsuereth, line 541: Nice! probably should have done this first, but good catch and thanks for the extra tests!

jsuereth (approved):

PR #1697, opened 2026-08-13
Add codegen repo to the downstream check mix

lmolkova:
Use otel-rust codegen - check if codegen succeeds and generated code builds


Cargo.toml
```diff
@@ -24,7 +24,10 @@ eula = false
 # Workspace definition ========================================================
 [workspace]
 members = ["crates/*"]
-exclude = ["fuzz"]
+# `target/downstream` holds the repos cloned by `cargo xtask downstream-check`.
+# Several are cargo projects themselves, and without this cargo refuses to build
+# them: they sit under this workspace's root but are not members of it.
+exclude = ["fuzz", "target/downstream"]
```
lmolkova: fixes

```
Error: FAILED (exit 1) - see the 'https://github.com/open-telemetry/opentelemetry-weaver-examples.git: basic - resolve' group above
https://github.com/open-telemetry/opentelemetry-weaver-examples.git: basic - live-check otlp
  --- in target/downstream/open-telemetry_opentelemetry-weaver-examples/basic
  --- $ set -e
  state=$(mktemp -d)
  make build
  '/home/runner/work/weaver/weaver/target/release/weaver' registry live-check --registry ./model --input-source otlp \
    --otlp-grpc-port 4317 --admin-port 4320 --inactivity-timeout 30 \
    --format json --output http >"$state/weaver.log" 2>&1 &
  pid=$!
  trap 'kill $pid 2>/dev/null || true; cat "$state/weaver.log" || true' EXIT
  ready=""
  for _ in $(seq 1 30); do
    if curl -fsS http://127.0.0.1:4320/health >/dev/null 2>&1; then ready=1; break; fi
    sleep 1
  done
  if [ -z "$ready" ]; then echo "live-check never became healthy"; exit 1; fi
  OTEL_EXPORTER_OTLP_ENDPOINT=http://127.0.0.1:4317/ cargo run -- "Hello, World!"
  curl -fsS -X POST http://127.0.0.1:4320/stop -o "$state/report.json"
  python3 -c 'import json,sys
  counts = json.load(open(sys.argv[1]))["statistics"]["advice_level_counts"]
  violations = counts.get("violation", 0)
  print(f"live-check violations: {violations}")
  sys.exit(1 if violations else 0)' "$state/report.json"
  
  cargo build
  error: current package believes it's in a workspace when it's not:
  current:   /home/runner/work/weaver/weaver/target/downstream/open-telemetry_opentelemetry-weaver-examples/basic/Cargo.toml
  workspace: /home/runner/work/weaver/weaver/Cargo.toml
  
  this may be fixable by adding `target/downstream/open-telemetry_opentelemetry-weaver-examples/basic` to the `workspace.members` array of the manifest located at: /home/runner/work/weaver/weaver/Cargo.toml
  Alternatively, to keep it out of the workspace, add the package to the `workspace.exclude` array, or add an empty `[workspace]` table to the package's manifest.
  make: *** [Makefile:4: build] Error 101
```

https://github.com/open-telemetry/weaver/actions/runs/31563764336/job/94011277759#step:11:3886

downstream-check.yaml
```diff
@@ -117,3 +117,31 @@ repos:
...
+    # The weaver image is pinned inline in build.gradle.kts with no override
+    # hook, so it is rewritten to point at ours before generating.
+    # TODO: add a WEAVER_CONTAINER-style override in semantic-conventions-java
+    # and drop the rewrite.
+    checks:
+      - name: generate-and-compile
+        timeout_minutes: 40
+        run: |
+          set -e
+          python3 - {weaver} <<'PY'
```
jsuereth: This one makes me a bit nervous for a few reasons:

- we have a gradlew dependency now *and* we shoudl update our workstream to have a gradlew cache or this build may take a LONG time.
- We're pinning to *main* which may be unstable - I'd prefer if we pinned against a known version for our integration test, otherwise this may be flaky.
- minor: it's python embedded in yaml, so it's harder to read

I 100% agree it's useful to test this in some fashion.  Perhaps we can have this test run their weaver config without actually using gradle to do it (if we're already using nested python scripts)

lmolkova: I switched to otel-rust instead of java so it has the same toolchain but also rust codegen is more diverse than java https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-semantic-conventions/scripts/templates/registry/rust/weaver.yaml, so we cover more with less.


I think testing against main it still ok here: the only thing we expect is that codegen does not fail and code compiles. It should be to assume that code on main compiles.

downstream-check.yaml
```diff
@@ -117,3 +117,28 @@ repos:
       - name: custom_advisor - test
         dir: custom_advisor
         run: make test
+
+  - url: https://github.com/open-telemetry/opentelemetry-rust.git@main
```
jsuereth: Can we pin this to a specific version?

lmolkova: I think testing against main it still ok here: the only thing we expect is that codegen does not fail and code compiles. It should be safe to assume that code on main compiles.

Cargo.toml
jsuereth: I have the same fix in my PR - I didn't realize this PR was load bearing - I thought it was a feature.

downstream-check.yaml
jsuereth: If we weren't customizing the build script and placing an invisible dependency on its structure, i'd agree.  In my past experience - that dependency silently breaks and we don't want it to be chaos if/when it does.  We do refactor builds, and particualrly for gen-ai semconv or codegen I suspect we'll be evolving them rapidly at this point.

jsuereth (approved):

Issue #1668, opened 2026-08-04
live-check: benchmark repeated Rego shared-context cost before optimizing


McGluut:
## Is your feature request related to a problem? Please describe.

Merged PR #1605 made both `resource` and `instrumentation_scope` available beside each signal sample in live-check Rego input. During review, @jerbly noted that each sample may serialize the same shared context again and suggested capturing an issue to measure before optimizing ([discussion](https://github.com/open-telemetry/weaver/pull/1605#discussion_r3678161654)).

`RegoAdvisor::advise` rebuilds the input and calls `Engine::set_input` for each sample. We do not yet have an upstream baseline showing how the cost changes with shared-context size or which part of the path dominates, so selecting caching or reuse now would be premature.

## Describe the solution you'd like

Would maintainers accept a benchmark-only follow-up in `weaver_live_check` with no production-code or public-API change?

The prepared Criterion benchmark has:

- four lanes: `serialize_input`, `set_input`, `policy_evaluation`, and `set_input_and_check`;
- seven fixtures: no context, resource-only with 0/16/128 attributes, and resource-plus-scope with 0/16/128 attributes;
- batch sizes 1, 32, and 256, for 84 cells total;
- `Throughput::Elements(batch_size)`, while every timed iteration loops exactly that count; and
- an untimed positive and name-mismatch control for every fixture before measurement.

A local synthetic feasibility run against current `main` (`f7b63a0`) completed all 84 cells. Representative batch-duration medians for batch 256 are below; parentheses contain the 95% confidence interval.

| Lane | No context | Resource + scope, 128 attributes each |
| --- | ---: | ---: |
| `serialize_input` | 0.616 ms (0.605–0.626) | 119.655 ms (92.639–130.553) |
| `set_input` | 1.551 ms (1.486–1.624) | 270.710 ms (213.445–363.557) |
| `policy_evaluation` | 7.720 ms (7.213–8.286) | 26.562 ms (21.773–36.409) |
| `set_input_and_check` | 13.926 ms (13.275–14.089) | 491.559 ms (336.796–561.189) |

This exposes context-size sensitivity in the synthetic fixture, but it does not establish production impact, a regression, or a materiality threshold. The lanes are separate Criterion measurements, so their medians should not be subtracted as paired observations. The run used Windows x86_64, Rust 1.96.1, and a portable `x86_64-pc-windows-gnullvm`/LLVM-MinGW target; I would treat it as a local baseline rather than a canonical environment result.

[...]

PR #1699, opened 2026-08-13
live-check: benchmark Rego shared context

McGluut:
## Summary

Add a Criterion benchmark for Rego input work with shared resource and instrumentation-scope context.

- Isolates serialization, `set_input`, policy evaluation, and the combined path.
- Covers seven context fixtures across batch sizes 1, 32, and 256.
- Performs fixed matching and nonmatching preflight checks before timing.
- Documents the synthetic-only interpretation boundary.

## Context

Follow-up to #1668. This is a baseline-only microbenchmark; it does not optimize the live-check path or claim production impact.

## Validation

- `cargo fmt --all -- --check`
- Locked/offline Cargo metadata
- Locked/offline gnullvm benchmark compile
- Full 84-cell Criterion run, including preflight
- `git diff --check`

The repository-wide `just` pre-push recipe was attempted, but this local Windows host stopped before the checks began because Git Bash could not create its signal pipe (Win32 error 5). CI remains the authoritative full pre-push validation.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -52,6 +52,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "alloca"
+version = "0.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e5a7d05ea6aea7e9e64d25b9156ba2fee3fdd659e34e41063cd2fc7cd020d7f4"
+dependencies = [
+ "cc",
+]
+
+[[package]]
 name = "allocator-api2"
 version = "0.2.21"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


McGluut:
@jerbly, would you have time to review whether this benchmark-only baseline is useful to land, or suggest the right reviewer? It follows the shared-context measurement question from #1605/#1668 and keeps optimization out of scope.

jerbly:
Sorry for the delay. ~What results did you get?~ I see this on the issue #1668  

We're planning some performance work and this could feed into that. So I think this PR will be parked for a while. Thanks for the work so far though.

McGluut:
Thanks, @jerbly — happy to leave this parked while the performance work takes shape. Is there an issue or design discussion I can follow? I’d like to understand which workload or performance decision you’re targeting, so any follow-up on this baseline is useful to that work.

PR #1704, opened 2026-08-15
Resolve an association, record where it resolved

jerbly:
1. A dangling entity_associations entry now fails the resolve, as does one naming a private (dependency_resolution.exclude) entity. Both were silent.
2. The resolved/2.0 leaf became { type, provenance }, so a published schema says which registry owns each associated entity without copying the entity in.
3. `entity.host` and `host` both become the entity `host`; the second used to replace the first silently. Now reported as a warning, from a single shared id rule that the conversion and the check both use.


crates/weaver_resolved_schema/src/v2/entity.rs
```diff
@@ -39,6 +39,107 @@ pub struct Entity {
...
+    }
+}
+
+/// Turns resolved association expressions back into the authored form, where a
+/// leaf is a name alone.
+///
+/// A v1 group and a markdown table both read a name, so every consumer of the
+/// two needs this.
+#[must_use]
+pub fn to_named_associations(
```
lmolkova, line 123: **`entity_associations` rebind to a local entity after one import hop**

`base` defines the entity:

```yaml
# base/registry.yaml — https://example.com/base/1.0.0
entities:
  - type: host
    identity:
      - ref: host.id
```

`middle` depends on `base`, imports `host`, and associates a metric with it:

```yaml
# middle/registry.yaml — https://example.com/middle/1.0.0
imports:
  entities:
    - host
metrics:
  - name: middle.request.count
    entity_associations:
      - host
```

`derived` depends on `middle`, imports the metric, and defines its own unrelated entity that happens to be called `host`:

```yaml
# derived/registry.yaml
imports:
  metrics:
    - middle.request.count
entities:
  - type: host
    identity:
      - ref: derived.host.uid
```

Expected: a duplicate-id error — `host` already exists in the imported surface.

Actual: resolves clean, no warnings. `weaver registry resolve --v2` on `middle` (trimmed):

```yaml
registry:
  metrics:
  - name: middle.request.count
    entity_associations:
    - host
  entities:
  - type: host
    identity:
    - key: host.id
    provenance:
      source: https://example.com/base/1.0.0
```

and on `derived`:

```yaml
registry:
  metrics:
  - name: middle.request.count
    entity_associations:
    - host
    provenance:
      source: https://example.com/middle/1.0.0
  entities:
  - type: host
    identity:
    - key: derived.host.uid
    provenance:
      path: derived/registry.yaml
```

Base's `host` is not in `derived` at all. The same metric now associates with an entity keyed on `derived.host.uid`.

The import path (`imports.rs`, metric/event/span arms) calls this function, so the association crosses as a bare name and is re-resolved against the importer's entities.

crates/weaver_resolver/src/dependency.rs
```diff
@@ -108,6 +110,80 @@ impl ResolvedDependency {
...
+        Some(EntityLocation {
+            origin: v2_source_url(self, &deps, entity.provenance.source),
+            excluded: is_excluded(&entity.common.annotations),
+        })
+    }
+}
+
+impl EntityLookup for Vec<ResolvedDependency> {
+    fn lookup_entity(&self, name: &str) -> Option<EntityLocation> {
+        self.iter().find_map(|d| d.lookup_entity(name))
```
lmolkova: **Association lookup takes the first dependency holding the name, so manifest order decides the result**

Two dependencies, each with an unrelated `host`. One hides it:

```yaml
# private_base/registry.yaml — https://example.com/base/1.0.0
entities:
  - type: host
    annotations:
      dependency_resolution:
        exclude: true
    identity:
      - ref: host.id
```
```yaml
# public_base/registry.yaml — https://rival.example.com/rival/1.0.0
entities:
  - type: host
    identity:
      - ref: rival.host.uid
```

`derived` depends on both and associates a metric with `host`:

```yaml
# derived/registry.yaml
metrics:
  - name: top.request.count
    entity_associations:
      - host
```

With `private_base` listed first:

```yaml
dependencies:
  - name: base
    schema_url: https://example.com/base/1.0.0
  - name: rival
    schema_url: https://rival.example.com/rival/1.0.0
```
```
× entity 'host' is marked as excluded from dependency resolution and cannot
│ be used by 'metric.top.request.count'
```

Swap the two `dependencies` entries — nothing else changes — and it resolves:

```yaml
registry:
  metrics:
  - name: top.request.count
    entity_associations:
    - host
```

with the association recorded against `https://rival.example.com/rival/1.0.0`.

`find_map` stops at the first dependency that has the name, and `excluded` is then checked only on that one, so a hidden entity in an earlier dependency shadows a visible one later. An excluded entity is not part of the dependency's surface, so the lookup should skip it and keep looking rather than stop there.

crates/weaver_resolver/src/registry.rs
```diff
@@ -436,6 +494,108 @@ fn resolve_dependency_imports<C: crate::SchemaCacheLookup>(
...
+    for g in ureg.groups.iter() {
+        let imported = ureg.registry.entity_association_origins.get(&g.group.id);
+        for name in g
+            .group
+            .entity_associations
+            .iter()
+            .flat_map(|assoc| assoc.referenced_entities())
+            .unique()
+        {
+            if local.contains(name) || imported.is_some_and(|m| m.contains_key(name)) {
```
lmolkova: **Associating a signal with an excluded entity should be an error**

Unless the signal is excluded too — the rule attribute refs already follow: https://github.com/open-telemetry/weaver/blob/d7b979e919993c0feb91a416037b7f6843453746/crates/weaver_resolver/src/attribute.rs#L329-L348 (and `excluded_parent_error` for `extends`/`include`: https://github.com/open-telemetry/weaver/blob/d7b979e919993c0feb91a416037b7f6843453746/crates/weaver_resolver/src/registry.rs#L1107-L1128).

`local.contains(name)` here passes over local names without checking exclusion, so `local_private_assoc` is accepted and exports a public metric referencing an entity no dependent can see. A consumer importing that metric then fails or silently rebinds to its own `host`, depending on whether it happens to define one.

lmolkova (approved):
reported a small final issue to fix, LGTM otherwise

PR #1712, opened 2026-08-19
Add selectable Rustls crypto providers

### LINKED ISSUE #1382: Move SSL-dependency decisions into features
As seen on https://github.com/open-telemetry/weaver/pull/1380 - we have an issue where embedding OpenSSL isn't desired in all consumption use cases for weaver.  Additionally, *all* platforms pay the embedding penalty for OpenSSL even when it isn't necessary because we needed it for our docker build.

Going forward we'd like to create a set of features for our crates where you can pick your RustTLS / SSL dependency as needed and then update per-platform builds to use the appropriate feature set for that platform.  This should be influenced by how other libraries, e.g. gix, reqwest, do the same with features.

lquerel:
Fixes #1382.

## Summary

Move Weaver’s TLS cryptography choice behind Cargo features, following the provider-selection pattern used by `otel-arrow`.

Weaver now uses `ring` by default while allowing custom builds to select AWS-LC, OpenSSL, vendored OpenSSL, or SymCrypt. Reusable crates remain crypto-provider-neutral, and executable entry points install the selected process-wide Rustls provider.

## Crypto provider features

| Feature | Provider | Notes |
| --- | --- | --- |
| `crypto-ring` | ring | Default provider |
| `crypto-aws-lc` | AWS-LC | AWS-LC-backed Rustls provider |
| `crypto-openssl` | OpenSSL | Uses the system OpenSSL installation |
| `crypto-openssl-vendored` | OpenSSL | Builds and statically links OpenSSL |
| `crypto-symcrypt` | SymCrypt | Linux and Windows; requires the system SymCrypt shared library |

Alternative providers are selected by disabling the default features:

```shell
cargo build --release --no-default-features --features crypto-aws-lc
```

Provider features are mutually exclusive. A provider-free build remains possible with `--no-default-features`, although TLS operations then require the embedding process to install a provider.

## Implementation

- Configure `ureq` and the `gix`/`reqwest` transport to use Rustls without  selecting a provider.
- Disable `jsonschema`’s unused HTTP and file resolver defaults, which otherwise introduce an implicit AWS-LC provider.
- Install the selected provider at the beginning of the Weaver CLI.
- Install ring explicitly in `xtask`, since it is a separate TLS-using executable.
- Remove the unconditional vendored OpenSSL dependency from `weaver_common`.
- Keep TLS-using test executables responsible for installing their provider.
- Make the static musl build select ring explicitly.
- Add CI coverage for provider-free and alternative-provider builds.
- Document the available provider features and their platform requirements.

## Downstream note for `otel-arrow`

@jmacd @drewrelmas @cijothomas 

This change will not affect `otel-arrow` until it updates its pinned Weaver revision.

After upgrading the Weaver crates, `otel-arrow` should continue using its own `crypto-ring`, `crypto-aws-lc`, `crypto-openssl`, and `crypto-symcrypt` features to select the process-wide Rustls provider. Weaver’s reusable crates deliberately do not select or install a provider.

The main `df-engine` binary already installs its selected provider early in startup, so no production change should be necessary. Standalone tests, benchmarks, or tools that use Weaver to access an HTTPS Git repository must also
install a provider before invoking those paths: `otap_df_otap::crypto::ensure_crypto_provider();`

The expected result is a smaller dependency graph: Weaver will no longer force `native-tls` or vendored OpenSSL into `otel-arrow`, and its Git/HTTPS operations will use the provider selected by the embedding `otel-arrow` process.


crates/xtask/Cargo.toml
```diff
@@ -19,12 +19,17 @@ gix = { version = "0.85.0", default-features = false, features = [
     "blocking-http-transport-reqwest",
     "max-performance-safe",
     "worktree-mutation",
-    "blocking-http-transport-reqwest-rust-tls",
     "sha1",
 ] }
+gix-reqwest.workspace = true
+rustls = { version = "0.23.40", default-features = false, features = ["ring", "std", "tls12"] }
```
jsuereth: nit: I've been slowly moving to all dependency versions declared in the workspace, and overrides being done on lower-level crates.

cross-arch-build.sh
```diff
@@ -23,5 +23,6 @@ case "${HOST_ARCH}" in
 esac
 
 rustup target add "${RUST_TARGET}"
-cargo build --release --target="${RUST_TARGET}"
+# Keep the container's static musl build independent of system TLS libraries.
+cargo build --release --target="${RUST_TARGET}" --no-default-features --features=crypto-ring
```
jsuereth, line 27: Ah - we can move back to crypto-ring now? great!

jsuereth (approved):

crates/xtask/Cargo.toml
lquerel: Good point. I moved the `rustls`, `rustls-openssl`, and `rustls-symcrypt` versions and shared settings to `[workspace.dependencies]`. The root package and lower-level crates now inherit them with `workspace = true`, keeping only package-specific options such as `optional = true` and the `ring` feature at the consumer level.

Addressed in a83dca01.

cross-arch-build.sh
lquerel: Yes. `gix` and `ureq` now use their provider free Rustls configurations, so the reusable crates no longer select a process wide crypto provider implicitly! This lets the Weaver CLI explicitly select Ring again (including for the static musl build) without the previous provider ambiguity.

lquerel:
@jsuereth This PR should be ready for a new review.

PR #1716, opened 2026-08-20
live-check correct entity lookup

jerbly:
Live-check now follows a v2 `entity_associations` reference into a dependency, or to an entity refinement, neither of which the checker could see before: those entities went unchecked, so a resource missing their required attributes passed clean. The entities of a v2 registry reach the Rego data document too, as `data.entities`.


crates/weaver_live_check/src/live_checker.rs
```diff
@@ -40,6 +40,22 @@ pub struct LiveChecker {
     pub finding_modifier: Option<FindingModifier>,
 }
 
+/// Adds a v2 entity to the name index, unless the name is already taken.
```
jsuereth: Are we using the resolution order here then for conflcits?

Nit: We *may* want to look up entity ref by a *string and schema-url pair*.   Is that an option here?

crates/weaver_live_check/src/advice/type_advisor.rs
```diff
@@ -179,11 +201,46 @@ pub(crate) enum AssocNode<'a, A> {
...
+    fn entity_type(&self) -> &str {
+        self
+    }
+
+    fn lookup<'a>(&self, live_checker: &'a LiveChecker) -> Option<EntityDef<'a>> {
+        live_checker.find_entity(self).map(EntityDef::from)
+    }
+}
+
+impl AssocLeaf for EntityRef {
```
jsuereth: This should have the schema_url attached.

crates/weaver_live_check/src/advice/type_advisor.rs
```diff
@@ -179,11 +201,46 @@ pub(crate) enum AssocNode<'a, A> {
...
+/// definition of that entity lives.
+pub(crate) trait AssocLeaf {
+    /// The entity type, or refinement id, that the leaf names.
+    fn entity_type(&self) -> &str;
+
+    /// The definition, read from the registry the leaf points at.
+    fn lookup<'a>(&self, live_checker: &'a LiveChecker) -> Option<EntityDef<'a>>;
+}
+
+impl AssocLeaf for String {
```
jsuereth: When do we only have raw strings?

Is this only possible in V1?

In which case, maybe we only do the fallback "search everything" in the V1 lookup flow, but we lookup *by schema id* in V2 since you've remembered better where it came from.

jsuereth (approved):
Based on offline discussion - approving assuming we can clean up V2 path past V1 deprecation

jerbly:
Popping this back into draft while I rework it for only v2 support.

crates/weaver_live_check/src/live_checker.rs
```diff
@@ -165,9 +168,26 @@ impl LiveChecker {
...
+    /// Find the entity that a v2 association reference names
+    ///
+    /// The reference says which registry defines the entity, so this reads that
+    /// one registry. `None` means no registry in scope answers the reference, or
+    /// the registry under check is v1 and holds no such reference.
+    #[must_use]
+    pub fn lookup_entity(&self, entity_ref: &EntityRef) -> Option<&V2Entity> {
+        match self.registry.as_ref() {
+            VersionedRegistry::V2(registry) => registry.lookup_entity(entity_ref).ok(),
+            VersionedRegistry::V1(_) => None,
```
jsuereth, line 189: QQ - since I can't follow the logic super fast -

Does this mean *anyone* using entity_associations in v1 will get an error?

crates/weaver_live_check/src/advice/type_advisor.rs
```diff
@@ -138,7 +160,7 @@ pub(crate) fn check_entity_resource_attributes(
     };
 
     match entity {
-        VersionedEntity::V1(group) => {
+        EntityDef::V1(group) => {
```
jsuereth, line 163: I'd expect "V1" to be a no-op - unless this code only runs in `--v2` then I think it's fine...

jsuereth (approved):
Approving because V2 looks good - but question on V1 behavior - not sure if I'm interpreting it right as the logic to supress v1 checks may be outside the diffs

defaults/jq/advice.jq
```diff
@@ -21,6 +21,31 @@
     reduce .[] as $template ({}; . + {($template): true})
   ),
   
+  # The schema url of the registry under check. An association leaf omits its
+  # provenance when the entity is defined here, so a policy resolves the absence
+  # against this.
+  "schema_url": .registry.schema_url,
```
lmolkova, line 27: nit: do we have any docs about rego input schemathat need to be updated?

crates/weaver_live_check/data/policies/entity_advice/entities.rego
```diff
@@ -0,0 +1,69 @@
...
+		[attr.value, attr.name, prefix, assoc.type, source],
+	)
+}
+
+make_advice(advice_type, advice_level, advice_context, message) := {
+	"type": "advice",
+	"advice_type": advice_type,
+	"advice_level": advice_level,
+	"advice_context": advice_context,
+	"message": message,
```
lmolkova, line 68: should we add signal type signal name here as for other common advices ?

lmolkova (approved):

crates/weaver_live_check/src/live_checker.rs
jerbly: No. lookup_entity is only called from v2 logic.

crates/weaver_live_check/src/advice/type_advisor.rs
jerbly: We still have support for best-effort entity_association checks in live-check when you're running v1 - that was previously released code a few versions ago so I've kept that existing functionality.

crates/weaver_live_check/data/policies/entity_advice/entities.rego
jerbly: The rego advisor fills those in for you so you don't need to. Only if you want to override.

defaults/jq/advice.jq
jerbly: We should have something more formal for this, but we don't. There's a small sectino in the live-check readme which I've added to here: [1424695](https://github.com/open-telemetry/weaver/pull/1716/commits/14246953fbbb3766f8613b2f9172b696edd16039)

PR #1718, opened 2026-08-21
Add lookup_entity jinja function for v2 generate

jerbly:
Adds a lookup_entity function for jinja templates for v2 registries only. You give it a leaf node from the entity associations tree and it calls the lookup function on the registry to find that entity by name and provenance.

Included doc updates, and a test + example macro for walking the tree and grabbing the entities.

(This is the penultimate PR for #1692)


crates/weaver_forge/src/lib.rs
```diff
@@ -886,13 +903,14 @@ pub(crate) fn install_weaver_extensions(
...
     comment_flag: bool,
+    v2_registry: Option<Arc<ForgeResolvedRegistry>>,
 ) -> Result<(), Error> {
     code::add_filters(env, config, comment_flag)?;
     ansi::add_filters(env);
     case::add_filters(env);
     otel::add_filters(env);
     util::add_filters(env, config);
-    util::add_functions(env);
+    util::add_functions(env, v2_registry);
```
jsuereth, line 913: It's interesting we don't have general lookup/helper utils that keep the full registry for you even if the "env" / "context" we send in is filtered with JQ.  I kinda like this pattern going forward for some other helpers we have/use.

jsuereth (approved):

crates/weaver_forge/src/lib.rs
jerbly: I like it too - I think it opens up a lot.

PR #1721, opened 2026-08-23
Live-check for v2 semconv using Matchers

jerbly:
This PR contains a document describing an approach for live-check to work with v2. It's design to cover use cases that need to be very strictly signal-based and use cases that are more relaxed. The way you configure how live-check matches a sample to a signal and optionally attribute_groups is the secret sauce.

The key item to review at this stage is [crates/weaver_live_check/docs/live-check-v2-semconv.md](https://github.com/open-telemetry/weaver/compare/main...jerbly:weaver-cel?expand=1#diff-238a004a86b96ad5cd090855912ab4c0e7ebf831919481678f4baa9215b7fb69)

The rest of the files in this PR are examples/tests showing the usage of CEL, which is the recommendation here, for the examples in the document.

The document is in three parts:
1. The use cases and the problem to solve.
2. The spec for Matchers
3. A Bonus showing how CEL can also be used to make `conditionally_required` programmatically readable and enforceable in live-check in some circumstances.

----

## Implemented Matchers for live-check (not `conditionally_required` that's a separate item)

  - Added `[[live-check.matchers]]`, which picks the v2 signal and attribute groups a sample is checked against with a CEL expression. v2 registries only.
  - A matcher's `attribute_groups` says which attributes are permitted on a sample: their definitions are used for the attribute checks, but one missing from the sample is not reported. Name a group in the new `strict_attribute_groups` to enforce its requirement levels; a signal's own attributes are always enforced.
  - New findings: `kind_mismatch` and `unexpected_attribute`. A v2 metric or log raises `unexpected_attribute` against its natural signal or against the matcher's where one sets `signal`.
  - By default a v2 registry compares an attribute with the signal and attribute groups its match holds. However, when `search_all_attributes` is used, it searches the whole registry and its dependencies. v1 is unchanged.
  - Every sample's result holds a `match_info` giving the signal, the attribute groups and what each matcher that applied contributed.
  - Statistics count what each matcher matched and errored on, and a matcher that matched nothing is reported as a warning.
  - Added `-D`/`--param` and `--params` to pass parameters to the output template. The ansi format reads `show_finding_id`, which labels a finding with its id rather than its level.
  - Added [Matchers](crates/weaver_live_check/docs/matchers.md), a guide with a worked example for each sample type, and corrected the config sections in the live-check and config READMEs from `[live_check]` to `[live-check]` — the underscore form was silently ignored.


crates/weaver_live_check/docs/live-check-v2-semconv.md
```diff
@@ -0,0 +1,483 @@
...
+A scope also carries a `schema_url`, which is the schema the telemetry claims to follow. We ingest it today and hand it to policies, but we do nothing else with it. It is worth comparing it with the `schema_url` of the registry we are checking against, because if the two disagree then the samples were built against a different version of the schema and some of the findings that follow are explained by that. A finding at information level when they differ would tell you so straight away.
+
+---
+# Matcher specification
+
+A Matcher gives live-check an identifier for the samples that do not have one. It can also add attributes to the set that a sample is compared with. A matcher never changes the checks themselves, it only decides what a sample is compared with.
+
+## Where matchers are defined
+
+Matchers describe the telemetry you emit, not the schema you define, so they belong in `.weaver.toml` and not in the registry. They are an array of tables and they are evaluated in the order you write them:
```
lmolkova, line 304: TBH I can see them making it into the schema too in a different form:

e.g. I want to say that all logs my app emits are going to have this set of attributes or all HTTP server spans are enriched with `user.id`.

I don't think it's a problem though - weaver.toml seems like a good place to start and maybe some of the features will become available in the definition schema too.

crates/weaver_live_check/docs/live-check-v2-semconv.md
```diff
@@ -0,0 +1,483 @@
...
+These are the kind that are worth writing:
+
+| Sentence | Expression |
+|---|---|
+| If and only if `task.arn` is populated. | `"aws.ecs.task.arn" in attributes` |
+| If `server.address` is set. | `"server.address" in attributes` |
+| if and only if k8s.hpa.metric.type is ContainerResource | `"k8s.hpa.metric.type" in attributes && attributes["k8s.hpa.metric.type"] == "ContainerResource"` |
+| Required if `exception.type` is not set, recommended otherwise. | `!("exception.type" in attributes)` |
+
+Some conditions are about the attribute they apply to. `server.port` is `conditionally_required: If not default (443).` When the port is missing there is no value to compare with 443. You cannot express a condition like that, so it keeps the sentence alone.
```
lmolkova, line 484: this is where I think we should have definition schema to support it.

If we `group_ref: server` (public group) in definition schema, we should preserve some indication of it in the resolved schema and then we can auto-match to the group without anything in the weaver.toml.

It'd be useful beyond life checks. 

This is just to say that we probably should solve some of these problems in other ways, but this is a great step forward anyway!

crates/weaver_live_check/docs/live-check-v2-semconv.md
```diff
@@ -0,0 +1,483 @@
...
+| A `when` today | 46 | 20% |
+| About the outcome of a metric or an event, which has no status | 36 | 16% |
+| One clause you can express, one you cannot | 13 | 6% |
+| About the value of the attribute it applies to | 24 | 10% |
+| On an attribute group that spans and metrics share | 4 | 2% |
+| Not about anything live-check can see | 107 | 46% |
+
+Three things come out of the scan.
+
+The largest group is about failure. 47 conditions set `error.type`, or something like it, when the operation failed. `status.code == "error"` expresses that, and it is the only way to express it: a `when` of `"error.type" in attributes` is true exactly when the attribute is there, so it can never report the attribute missing.
```
lmolkova, line 503: for this I was also thinking about a bit more expressive schema like

```
   requirement_level:
     conditionally_required: on_success | on_failure | free-form note
```

A lot of attributes would become checkable with something like this

lmolkova (commented):
this is awesome!

crates/weaver_live_check/src/sample_attribute.rs
```diff
@@ -174,18 +174,34 @@ impl LiveCheckRunner for SampleAttribute {
...
         let mut result = LiveCheckResult::new();
-        // find the attribute in the registry
-        let semconv_attribute = {
-            if let Some(attribute) = live_checker.find_attribute(&self.name) {
-                Some(attribute)
-            } else {
-                live_checker.find_template(&self.name)
-            }
-        };
+        // v1 searches the whole registry; v2 only when asked.
```
lmolkova, line 181: When `search_all_attributes` is false, `SampleAttribute::run_live_check` emits `missing_attribute` (`Violation`) for any attribute not in the matched signal or attribute groups, even if the attribute exists in the registry. This duplicates the `unexpected_attribute` (`Improvement`) finding already raised on the signal and promotes an improvement into a violation, breaking `--fail-on violation`.

```rust
#[test]
fn unexpected_attribute_does_not_raise_missing_attribute_violation() {
    let mut live_checker = v2_live_checker_full(
        Vec::new(),
        vec![span_attribute(
            "myapp.checkout.id",
            RequirementLevel::Basic(BasicRequirementLevelSpec::Required),
        )],
        Vec::new(),
        vec![Box::new(TypeAdvisor)],
    );
    live_checker
        .set_matchers(&matcher_configs(MATCHERS))
        .expect("valid config");
    let mut stats =
        LiveCheckStatistics::Cumulative(CumulativeStatistics::new(&live_checker.registry));
    let mut span = checkout_span(); // contains declared "myapp.checkout.id" and undeclared "myapp.checkout.stage"
    let parent = Sample::Span(span.clone());
    span.run_live_check(&mut live_checker, &mut stats, None, &parent)
        .unwrap();

    let stage_attr = &span.attributes[1];
    let attr_findings = stage_attr
        .live_check_result
        .as_ref()
        .map(|r| r.all_advice.as_slice())
        .unwrap_or_default();

    assert!(
        !attr_findings.iter().any(|f| f.id == "missing_attribute"),
        "unexpected attribute should not be reported as missing_attribute violation: {attr_findings:?}"
    );
    assert_eq!(
        stats.max_level(),
        Some(FindingLevel::Improvement),
        "unexpected attribute should remain an improvement, not escalate to violation"
    );
}
```

Skip emitting `missing_attribute` when the parent signal match already checked expected attributes, or only emit `missing_attribute` when the full registry was searched.

crates/weaver_live_check/src/cel.rs
```diff
@@ -0,0 +1,739 @@
...
+}
+
+impl Bindings for SampleLog {
+    fn bind(&self, referenced: &Referenced, context: &mut Context<'_>) {
+        if referenced.wants("event_name") {
+            context.add_variable_from_value("event_name", self.event_name.as_str());
+        }
+        // An absent field is left unbound, so reading it errors rather than
+        // testing a default.
+        if referenced.wants("severity_text") {
```
lmolkova, line 161: Optional variables (`body`, `severity_text`, `severity_number`, `resource`, `instrumentation_scope`) are left unbound in CEL context when absent. Because CEL does not support testing whether a top-level identifier is declared, expressions cannot guard with `!= null` and fail evaluation with `Undeclared reference`.

```rust
#[test]
fn absent_optional_fields_can_be_guarded_with_null_check() {
    let log: SampleLog = parse(
        r#"{ "event_name": "myapp.order.placed", "attributes": [], "live_check_result": null }"#,
    );

    // Expressions guarding optional fields should evaluate cleanly to false when absent
    let matched = evaluate(r#"body != null && body.contains("declined")"#, &log)
        .expect("guarding against null should evaluate without error");
    assert!(!matched);
}
```

Bind absent optional fields as `Value::Null` so expressions can safely guard with `!= null`.

crates/weaver_live_check/src/advice/type_advisor.rs
```diff
@@ -807,6 +867,59 @@ impl Advisor for TypeAdvisor {
...
+                    return Ok(Vec::new());
+                };
+                let VersionedSignal::Span(span) = &*semconv_span else {
+                    return Ok(Vec::new());
+                };
+                let mut advice_list =
+                    check_attributes(&span.attributes, &sample_span.attributes, parent_signal);
+                if sample_span.kind != v2_span_kind_to_v1(span.kind) {
+                    advice_list.push(PolicyFinding {
+                        id: FindingId::KindMismatch.into(),
```
lmolkova, line 881: nit: `KindMismatch` context contains `expected_value` but omits `actual_value`. Add `ACTUAL_VALUE_ADVICE_CONTEXT_KEY: enum_name(&sample_span.kind)` to match `type_mismatch`.

lmolkova (approved):
some edge cases and nits found by AI, LGTM otherwise!

crates/weaver_live_check/src/sample_attribute.rs
jerbly: The current behaviour is by-design:
1. Always provide "unexpected attribute" advice at the signal/parent level, when the sample attribute was not a member of the matched signal or attribute_group sets.
2. Always provide all advice for every attribute.

"unexpected" is a finding on the signal/parent.
"missing_attribute" is a finding on the attribute itself.

`--search-all-attributes` influences "unexpected" to provide a hint that the attribute exists somewhere in your registry-tree. It affects the analysis at the attribute level since that attribute could (if found) be checked.

I believe this behaviour is correct.

crates/weaver_live_check/src/advice/type_advisor.rs
jerbly: fixed in https://github.com/open-telemetry/weaver/pull/1721/commits/a7bef1a8164c0ec8eb845eb2e80c89e5767dc14e

crates/weaver_live_check/src/cel.rs
jerbly: fixed in https://github.com/open-telemetry/weaver/pull/1721/commits/a7bef1a8164c0ec8eb845eb2e80c89e5767dc14e

crates/weaver_cel/src/expression.rs
```diff
@@ -0,0 +1,327 @@
...
+
+use cel::{Context, Env, Program, Value};
+
+use crate::matches::literal_patterns;
+use crate::{free_variables::free_variables, Bindings, Error};
+
+thread_local! {
+    /// The CEL standard environment, built once per thread because
+    /// `Context::default` rebuilds the whole function table.
+    static STDLIB: Arc<Env> = Arc::new(Env::stdlib());
```
jsuereth: Does this mean evaluate is not re-entrant?

I think that's probably fine.

crates/weaver_cel/src/expression.rs
```diff
@@ -0,0 +1,327 @@
...
+
+impl Expression {
+    /// Compiles an expression.
+    pub fn compile(source: &str) -> Result<Self, Error> {
+        let program = Program::compile(source).map_err(|error| Error::CompileFailed {
+            expression: source.to_owned(),
+            error: error.to_string(),
+        })?;
+        for pattern in literal_patterns(program.expression()) {
+            let _ = regex::Regex::new(pattern).map_err(|error| Error::BadPattern {
```
jsuereth: Why are you doing this and ignoring the error? Can you at least comment on that?

crates/weaver_cel/src/free_variables.rs
```diff
@@ -0,0 +1,78 @@
+// SPDX-License-Identifier: Apache-2.0
+
+//! Collects the variables an expression reads from its environment.
+
+use std::collections::BTreeSet;
+
+use cel::common::ast::{EntryExpr, Expr, IdedExpr};
+
+/// The variables an expression reads that nothing in it binds.
```
jsuereth, line 9: Is this just for errors?

crates/weaver_cel/README.md
```diff
@@ -0,0 +1,73 @@
...
+use std::collections::HashMap;
+use weaver_cel::{Bindings, Context, Expression, Referenced, Value};
+
+struct Span {
+    name: String,
+    attributes: HashMap<String, Value>,
+}
+
+impl Bindings for Span {
+    fn bind(&self, referenced: &Referenced, context: &mut Context<'_>) {
```
jsuereth, line 38: Are we doing it this way to avoid cloning?

Can we bind by reference to context?

jsuereth (commented):
First round of review - questions on CEL design

crates/weaver_forge/src/output_processor.rs
```diff
@@ -154,6 +154,28 @@ impl OutputProcessor {
...
         templates_path: Option<PathBuf>,
         output: OutputTarget,
+    ) -> Result<Self, Error> {
+        Self::with_params(
+            format,
+            prefix,
+            embedded_templates,
+            templates_path,
+            output,
+            Params::default(),
```
jsuereth: Should we just remove this constructor for the new one?

crates/weaver_live_check/src/matcher.rs
```diff
@@ -0,0 +1,2333 @@
...
+
+    /// The signal and attribute groups to compare a sample with.
+    ///
+    /// The first matcher with a `signal` overrides the natural match; a later
+    /// one is listed in [`SampleMatch::conflicts`]. Attribute groups accumulate
+    /// in declaration order, first mention winning.
+    pub fn match_for(
+        &self,
+        sample_type: SampleType,
+        bindings: &dyn Bindings,
```
jsuereth: I think my only concern here is the API usage, lifetimes, etc.

I think we can clean this up BUT if we publsih on crates.io it would be breaking

crates/weaver_cel/src/expression.rs
jerbly: https://github.com/open-telemetry/weaver/pull/1721/commits/d4c536db40434be1e674a64cdf689ea7c81d9348 - It was re-entrant: STDLIB.with(Arc::clone) only copied an Arc out of the thread-local and nothing was held during evaluation. But turns out there's no reason for it to be per-thread. Replaced with a process-wide OnceLock<Arc<Env>>, and Scope::new now goes through Context::with_env instead of building Context::Root by hand, so a new field in a cel release won't break us.

crates/weaver_cel/src/expression.rs
jerbly: https://github.com/open-telemetry/weaver/pull/1721/commits/d4c536db40434be1e674a64cdf689ea7c81d9348 - It wasn't ignored: the ? propagated the error and let _ only dropped the successfully compiled Regex. Agreed it read badly. Rewritten as if let Err(error)... with a comment explaining the check exists because the cel interpreter only compiles a pattern at run time.

crates/weaver_cel/src/free_variables.rs
jerbly: https://github.com/open-telemetry/weaver/commit/d4c536db40434be1e674a64cdf689ea7c81d9348 - No, it's for performance. It produces the Referenced set that a Bindings impl uses, so an expression that only reads `name` never builds the attribute map, and Scope binds the union for all matchers once per sample instead of once per matcher. Expanded the doc comment to say this.

crates/weaver_cel/README.md
jerbly: Yes, it's to avoid cloning what an expression never reads, and no, cel can't bind by reference.

Claude explains this better than I can:

> The lifetime on Context<'a> is only for a child scope's parent pointer and an optional &'a dyn VariableResolver. The variables themselves are BTreeMap<String, Box<dyn Val>>, and Val: Any + Send + Sync, so every bound value has to be 'static. In practice that means Value::String(Arc<String>) and Value::Map(Arc<HashMap<Key, Value>>) with Arc<String> keys: binding an attribute map allocates per key and per string value, per sample.
> 
> Since cel forces the copy, the lever is to copy less. Referenced records which variables an expression reads, so a when of name.startsWith("checkout") never builds the attribute map, and Scope builds the union once per sample rather than once per matcher. The live-check impls pass as_str() rather than cloning first; the README example's .clone() overstates the cost, I'll switch it to as_str(). https://github.com/open-telemetry/weaver/pull/1721/commits/16cfebe3613659c765063e7c8eb0be4eeb9e86b8
> 
> cel does have a VariableResolver for lazy lookup, but it still returns owned Values and is consulted on every occurrence of an identifier, so it only wins when an expression names a variable and then short-circuits before reading it. Not worth the caching it would need.
> 
> The genuine bind-by-reference route is to satisfy 'static by sharing: hold sample attributes in Arcs and implement Val over them so attributes["k"] reads through without building a HashMap. That's a redesign of the live-check sample structs to use Arc everywhere, not a weaver_cel change, and I'd rather leave it until profiling shows attribute binding is the bottleneck.
> 

Should we start an issue to collect performance refactoring opportunities?

crates/weaver_forge/src/output_processor.rs
jerbly: https://github.com/open-telemetry/weaver/pull/1721/commits/3377bd864fd1149a1665350286de5a05d880fc37

crates/weaver_live_check/src/matcher.rs
jerbly: I think publishing live_check on crates.io is a long way off. However, I have improved the API and lifetimes around the Matchers in this commit: https://github.com/open-telemetry/weaver/pull/1721/commits/4a471aeea20e0d0279bae9d200ff53e758faba75

1. `Matchers::compile` now takes the `LiveChecker` and resolves every signal and attribute group to registry Rcs up front.
2. `Scope` lost its lifetime parameter. We only ever build root contexts with no resolver, so it holds `Context<'static>`.
3. `match_for` takes one `&dyn Matchable` instead of a `SampleType` plus `&dyn Bindings`. `Matchable` extends `Bindings` with `sample_type()`, so the type can't disagree with the sample.

crates/weaver_cel/README.md
jsuereth: Yes - what I'm worried about is the API will dramatically break due to exposing lifetimes.

I think the desire for the interface to be "simple" means claude is doing something wonky here.

I'd expect to have "Binding" actually just be a trait against the *live* `Sample`, and we have a `'a` lifetime that encompasses the lifetime of the `*Sample` here.

I'm not sure what kind of rules claude thinks it has to abide by here, but it's locked itself into a wierd pattern and I don't think it's correct.

crates/weaver_live_check/src/matcher.rs
jsuereth: Ah - Why does Context need the whole resolver? Is it pulling anything from there?  I would have expected to use something like [this](https://docs.rs/cel/latest/cel/context/trait.VariableResolver.html) - but possibly implement it using `serde` derivation - similar to how Rego or Jinja is working.  I.e. we can define a VariableResolver for any Serde serializable type.

What I wanted to see what something like:

```rust
// The entire need for a cel helper
fn create_context<'a, T: Serialize>(value: &'a T) -> Context<'a> {
   let mut ctx = Context::empty();
   ctx.set_variable_resolver(SerdeResolverHelper::new(value));
   ctx
}

// somewhere early - in livecheck
let cel_expression =  Program::compile("foo && size(foo) > 0")?;

// then inside livecheck
let my_context = create_context(&my_sample);
let result = cel_expression.execute(&my_context)?;
```

I think Claude basically invented a bunch of magic/krap here - but I'm not sure it lines up with how we do other things, and it feels WAY off to me. I'm sure it works, but maybe *don't* create a CEL crate and just hit it all private in live-check for now until we untangle it.

Otherwise this PR looks great - usage / integration to live check is good in my book.

jerbly: OK. I think I'll take this back to basics, remove the weaver_cel crate entirely (which was an outcome of early optimization considering reuse for `conditionally_required`) and make the lightest integration. I over engineered it.

jerbly:
@jsuereth - I've reworked this and removed all my premature optimizations and custom linting - they likely don't make a big difference anyway in hindsight. I think, we should do some measuring to bring in optimizations. And for the linting, I guess wait and see if runtime versus compile-time errors are annoying??

I deleted the weaver_cel crate and its custom AST walking, free-variable analysis, scoped binding and regex pre-checks. weaver_live_check now uses the cel crate directly. A `when` is a plain `cel::Program`. Each sample type implements a small `Matchable` trait beside its other impls, binding its variables through the cel crate's serde support. Got rid of ~1000 lines.

crates/weaver_live_check/src/sample_metric.rs
```diff
@@ -425,7 +447,102 @@ impl LiveCheckRunner for SampleMetric {
...
+            self.instrumentation_scope.as_deref(),
+            context,
+        )
+    }
+}
+
+impl SampleMetric {
+    /// The attributes that have the same value on every data point. A key
+    /// whose value differs between points is not included.
+    fn agreed_attributes(&self) -> AttributeMap<'_> {
```
jsuereth, line 476: Ah - yeah - this is where the *rust* implementation of CEL had some hard choices and we're one of the poor outcomes for how it models Values - specifically collections of values..

jsuereth (approved):
This looks good to merge now to me.

PR #1726, opened 2026-08-24
Changelog row for #1704 and a test that resolution keeps v2 signal names

lmolkova:
Follow-up to #1704.

- Add a changelog row noting #1704 as a v2-only breaking change: the `entity.` / `span.` prefix is no longer stripped from a signal type, so an entity authored as `type: entity.test` keeps that type instead of becoming `test`.
- Add a `weaver_resolver` test (with a v2 fixture) asserting the materialized span/metric/entity/event names match what was authored, so this behavior can't silently regress.


jerbly (approved):
Thanks for the fix!

PR #1729, opened 2026-08-26
Document generated-file header comment as a template best practice

### LINKED ISSUE #440: Add Comment in Generated Code Indicating Source and Generated Status as Best Practice in Jinja Templates

adedmla:
Resolves #440.

Adds a "Mark generated files" tip to `docs/codegen.md` recommending that templates emit a short header comment noting (1) that the file is generated and (2) where it was generated from. This helps downstream users avoid hand-editing generated artifacts and makes it easy to trace output back to its source template.

Supersedes #1085, which was closed only for an unsigned CLA. Docs-only change.


lmolkova (approved):
Thanks! would you mind checking if we have it documented and follow this guidance in https://github.com/open-telemetry/opentelemetry-weaver-packages where we intend to keep common generation templates? Thanks

adedmla:
Thanks for lmk! I didn't find it written down as authoring guidance. It's not in `CONTRIBUTING.md`, `AGENTS.md`, or `templates/docs/markdown/README.md`. Right now it seems to be followed by copying existing templates. This tip in the repo would be the place that codifies what those templates are already doing.

lmolkova:
> Thanks for lmk! I didn't find it written down as authoring guidance. It's not in `CONTRIBUTING.md`, `AGENTS.md`, or `templates/docs/markdown/README.md`. Right now it seems to be followed by copying existing templates. This tip in the repo would be the place that codifies what those templates are already doing.

@adedmla thanks! would you be interested in contributing it to templates there so it's actually populated on everything generated with templates?

adedmla:
Yes I would!

PR #1730, opened 2026-08-27
Fix forge dependency tree

jerbly:
**UPDATE**
Following a slack discussion the decision was to separate the list of registries from the graph representing the dependency relationships:

`dependencies` is now a map keyed by schema url, holding every registry this one depends on directly or indirectly, materialized once. The new `dependency_graph` gives the direct dependencies of each as schema urls only, where a registry with no entry has unknown dependencies rather than none.

The edges come from direct dependencies that `WeaverResolver` records as each registry resolves, so the urls are the ones actually used after version arbitration. For a registry consumed as an already-resolved artifact, they come from its manifest, carried on the new `LoadedSemconvRegistry::Resolved`/`ResolvedV2` struct variants and read through `SchemaResolver::direct_dependencies`.

----

ForgeResolvedRegistry::try_from_resolved_schema built the materialized dependencies from ResolvedTelemetrySchema::dependencies. That field is a flat list of every registry the schema was built from — the ones the manifest names and everything reached through them — not the direct edges. The code then recursed with the same function, so each registry expanded its own full list again.

For root → middle → sub → leaf:
```
produced                 expected

root                     root
├── leaf                 └── middle
├── middle                   └── sub
│   ├── leaf                     └── leaf
│   └── sub
│       └── leaf
└── sub
    └── leaf
```
A registry two hops away sat next to the one that actually depends on it, and leaf was materialized four times.

The flat list can't simply be trimmed to the direct edges: provenance.source (DependencyRef) is an index into it.

The fix

Record the direct edges separately during resolution, and build the tree from those.

- WeaverResolver keeps direct_dependencies: HashMap<SchemaUrl, Vec<SchemaUrl>>, filled in as each registry resolves. These are the URLs actually used, so version arbitration is already applied.
- LoadedSemconvRegistry::Resolved/ResolvedV2 become struct variants carrying direct_dependencies read from the registry's manifest. That covers a registry consumed as an already-resolved artifact, where the publication manifest is the only record of what it depends on.
- New SchemaResolver::direct_dependencies(&self, url) -> Option<&[SchemaUrl]>, defaulting to None, which means unknown rather than empty.
- ForgeResolvedRegistry::build(schema, resolver, expand): with the edges known, nest them and keep going; without them, list every dependency at one level and nest nothing, so at least nothing is duplicated.

ResolvedTelemetrySchema::dependencies is unchanged.


crates/weaver_forge/src/v2/registry.rs
```diff
@@ -39,7 +39,11 @@ pub struct ForgeResolvedRegistry {
     pub registry: Registry,
     /// The set of refinments defined in this registry.
     pub refinements: Refinements,
-    /// The resolved dependencies of this registry.
+    /// The registries this one depends on directly, as its manifest declares them,
+    /// each carrying its own dependencies in turn.
+    ///
+    /// Where the manifest is unknown, every direct and transitive dependency is
```
lmolkova: `defaults/jq/advice.jq` still assumes the flat closure, so live-check Rego policies silently lose every transitive registry from `data.entities`.

It maps one level:
```jq
[(.registry.dependencies // [])[], .registry]
```
and its comment ("The dependency list is the whole closure, so one level covers every entity") is now wrong. A leaf whose `provenance.source` is a transitive registry resolves to nothing, and the policy just doesn't fire. Needs a recursive descent, e.g. `[.registry | recurse(.dependencies[]?)]`.

Fails on this branch, passes before it:
```rust
#[test]
fn probe_rego_data_holds_transitive_entities() {
    const TOP_URL: &str = "https://example.com/top/1.0.0";
    const DEP_URL: &str = "https://example.com/base/1.0.0";
    const CORE_URL: &str = "https://example.com/core/1.0.0";
    let transitive = v2_assoc_registry(CORE_URL, vec![], vec![v2_entity("service", "core.service.name")], vec![], vec![]);
    let dependency = v2_assoc_registry(DEP_URL, vec![], vec![v2_entity("host", "host.id")], vec![], vec![transitive]);
    let (live_checker, _stats) = v2_live_checker(v2_assoc_registry(TOP_URL, vec![], vec![], vec![], vec![dependency]));
    let data = rego_data(&live_checker);
    let of_core = rego_entities_of(&data, CORE_URL);
    assert_eq!(of_core["service"]["identity"][0]["key"], "core.service.name");
}
```
```
no entities for https://example.com/core/1.0.0 in {"entities":{"https://example.com/base/1.0.0":{...},"https://example.com/top/1.0.0":{}}, ...}
```

crates/weaver_forge/src/v2/registry.rs
```diff
@@ -110,6 +110,17 @@ impl ForgeResolvedRegistry {
             })
     }
 
+    /// Finds a registry in the dependency tree by schema URL, at any depth.
+    fn find_dependency(&self, schema_url: &SchemaUrl) -> Option<&ForgeResolvedRegistry> {
```
lmolkova: The Rego counterpart of this walk is missing. `defaults/rego/semconv.rego` `defining_registry` still searches one level:
```rego
defining_registry(ref) := dependency if {
	some dependency in input.dependencies
	dependency.schema_url == ref.provenance.source
}
```
so `data.semconv.lookup_entity` is undefined for a leaf naming a transitive dependency, and a policy built on it silently skips the entity instead of checking it.

Fails on this branch, passes before it (`weaver_checker`):
```rust
#[test]
fn rego_lookup_entity_from_a_transitive_dependency() {
    let input = serde_json::json!({
        "schema_url": "https://example.com/main/1.0.0",
        "registry": {
            "entities": [],
            "metrics": [{"name": "m", "entity_associations": [{
                "type": "service",
                "provenance": {"source": "https://example.com/core/1.0.0"}}]}],
        },
        "refinements": {"entities": []},
        "dependencies": [{
            "schema_url": "https://example.com/dep/1.0.0",
            "registry": {"entities": []},
            "refinements": {"entities": []},
            "dependencies": [{
                "schema_url": "https://example.com/core/1.0.0",
                "registry": {"entities": [{"type": "service", "brief": "core service"}]},
                "refinements": {"entities": []},
            }],
        }],
    });
    assert_eq!(resolved_entities(&input), vec!["service -> core service"]);
}
```
```
left: []
right: ["service -> core service"]
```

lmolkova (commented):
do we still want rego to look up entities across all dependencies? If so I believe we need to update it too

crates/weaver_forge/src/v2/registry.rs
jerbly: Fixed in [32c6d95](https://github.com/open-telemetry/weaver/pull/1730/commits/32c6d95c178a4c286427d7e473ad62a531de4d12)

crates/weaver_forge/src/v2/registry.rs
jerbly: Fixed in [32c6d95](https://github.com/open-telemetry/weaver/pull/1730/commits/32c6d95c178a4c286427d7e473ad62a531de4d12)

crates/weaver_forge/src/v2/registry.rs
```diff
@@ -39,9 +40,46 @@ pub struct ForgeResolvedRegistry {
...
+    /// The direct dependencies of each registry, this one included. A registry
+    /// with no entry has unknown dependencies, not none.
+    #[serde(default, skip_serializing_if = "BTreeMap::is_empty")]
+    pub dependency_graph: BTreeMap<SchemaUrl, Vec<SchemaUrl>>,
+}
+
+/// A dependency of a [`ForgeResolvedRegistry`], stored under its schema url.
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, JsonSchema)]
+#[serde(deny_unknown_fields)]
+pub struct ForgeDependency {
```
jsuereth, line 56: Ah - I like this instead of re-using the other structure.

jsuereth (approved):
Overall looks good!

Thanks for making this massive refactor.


One thing I've been thinking -

We should record the issue we ran into and *why* the design is the way it is somewhere.  We don't really capture markdown design docs yet, but I'd love to have this written somewhere that we can repeatedly find when we touch this code.

Issue #970, opened 2025-10-06
Define span links in yaml definition


thompson-tomo:
**Is your feature request related to a problem? Please describe.**
I want my definitions of spans to also indicate what links may/should/must have

**Describe the solution you'd like**
Introduction of a new array property on the span definition called links which is an array of objects.

Each object would contain the following:

- Ref: what type of span it should be linked to & is required
- brief/note
- Attributes: defaults to empty list and can only be defined if link type is other
- link type: parent, child or other which is default

**Describe alternatives you've considered**

Defining links as first class signal and re-using but limited benefit especially with attribute groups available

**Additional context**
- https://github.com/open-telemetry/semantic-conventions/issues/1057

PR #1731, opened 2026-08-27
Add span links to the v2 span definition

vesari:
This is the first step to address #970.

A v2 span can now declare links to span types, including a requirement level, description, and expected link attributes:

```yaml
spans:
  - type: consumer.op
    kind: consumer
    brief: A local op span with links.
    links:
      - ref: consumer.batch
        requirement_level: required
        brief: The batch this op belongs to.
      - ref: linkserver.produce
        attributes:
          - ref: linkserver.job.id
            requirement_level: required
```

In the resolved schema, links contain the target span type, a normalized requirement level (defaulting to `recommended`), link attributes resolved to catalog references, and provenance for the registry that declared the link.

### Behaviour

- Span refinements inherit links from their base span (see follow-ups for spans from dependencies). They cannot declare their own links yet.
- Imported spans preserve their links, including spans reconstructed during dependency version upgrades.
- Locally declared link targets must name a span defined in or imported into the registry. Inherited links skip this validation because their targets may be defined in a dependency.
- Invalid link targets and unsupported link-attribute data produce fatal errors (`SpanLinkTargetNotFound`, `UnsupportedSpanLinkAttribute`) instead of being silently dropped.
- Link attributes must be plain attribute references: attribute-group references are rejected at parse time.
- Link attributes must already exist in the registry catalog.

### Design

This is a v2-only feature. Weaver carries links through its existing hidden v1 intermediate representation, alongside other v2-only fields such as `span_name`. Should the resolver become independent of v1 and v2, the links would migrate together with that family of fields.

Links use attribute requirement levels because links, unlike signals, may be required. They do not have a `link_type` field because runtime span links do not carry a type.

### Not in this PR

- Forge/template exposure; this will be implemented in a follow-up PR.
- Refinement semantics for links inherited from dependency spans.
- Live-check validation of links (presence and attribute checks need only the observed span itself; checking a target's type also needs the span it links to).

### Tests

Tests cover parsing, negative validation cases, conversion, local and imported links, refinement inheritance, link attributes, and dependency version-upgrade regressions.


vesari:
N.B. After resolving the merge conflicts arising from #1763, I aligned the links carrier with the #1763 pattern (i.e. how a v2-only field crosses the resolver): the hidden span_links carrier now holds the v2 type directly, with no v1 duplicate and no conversion. This PR originally carried links the old way, but this adoption doesn't affect behaviour (the end-to-end expected file is identical). Let me know if you want it made differently.

crates/weaver_resolver/data/registry-test-v2-dep/span_links_registry/published/resolved.yaml
```diff
@@ -0,0 +1,64 @@
...
+    name:
+      note: Process span name.
+    links:
+    - ref: linkserver.produce
+      requirement_level: opt_in
+      brief: One link per processed message.
+      attributes:
+      - base: 0
+        requirement_level: required
+        sampling_relevant: true
```
lmolkova, line 32: I think sampling relevancy should not be a link attribute ref property - link can be provided before or after span start and links themselves are not sampled on their own.

But maybe we can explore having sampling-relevancy on the whole link (i.e. should someone provide it at start time) - this should probably be a follow up

crates/weaver_resolver/data/registry-test-v2-dep/span_links_registry/registry/registry.yaml
```diff
@@ -0,0 +1,37 @@
...
+    name:
+      note: Op span name.
+    brief: A local op span with links.
+    stability: stable
+    attributes:
+      - ref: linkserver.job.id
+        requirement_level: recommended
+    links:
+      - ref: consumer.batch
+        requirement_level: required
```
lmolkova, line 27: do we need requirement level on links?

I'm thinking - e.g. I receive a message from the queue and it does not have context. Do I still provide the link with invalid context (and let's say I don't have any link attributes)? 

Maybe we should drop requirement level on links for now? we can add it back once we know the use-case, we'd find some non-breaking way.

schemas/semconv.schema.v2.json
```diff
@@ -1589,7 +1642,7 @@
       ]
     },
     "SpanRefinement": {
-      "description": "A refinement of an existing span.",
+      "description": "A refinement of an existing span.\n\nA refinement inherits the base span's links during resolution.\nA refinement cannot declare, replace, or extend links yet.",
```
lmolkova, line 1645: do we need to update this? the clarification is nice in the docs, but we don't describe refinement behavior fully here.

crates/weaver_resolved_schema/src/error.rs
```diff
@@ -64,6 +65,26 @@ pub enum Error {
         field: &'static str,
     },
 
+    /// A span link references a span type that does not exist in the registry.
+    #[error("Span link target '{link_ref}' (group: {group_id}) not found among span definitions")]
```
lmolkova, line 69:
```suggestion
    #[error("Span '{link_ref}' (linked by {group_id}) not found among span definitions")]
```

maybe?

crates/weaver_resolved_schema/src/error.rs
```diff
@@ -64,6 +65,26 @@ pub enum Error {
...
+    SpanLinkTargetNotFound {
+        /// Group id of the span declaring the link.
+        group_id: String,
+        /// The link's target span type.
+        link_ref: String,
+    },
+
+    /// A span link attribute uses a feature that is not supported yet.
+    #[error("Unsupported span link attribute in group {group_id} (link: {link_ref}): {reason}")]
```
lmolkova, line 78: when would we return this error? not sure I understand what this message means

crates/weaver_resolved_schema/src/convert.rs
```diff
@@ -232,6 +232,85 @@ fn convert_attribute_ref<'a>(
...
+        link_ref: link.r#ref.to_string(),
+        reason,
+    };
+    if ar.base.brief.is_some()
+        || ar.base.note.is_some()
+        || ar.base.examples.is_some()
+        || !ar.base.annotations.is_empty()
+    {
+        return Err(unsupported(format!(
+            "attribute '{}' carries overrides; only requirement_level and sampling_relevant are supported on link attributes",
```
lmolkova, line 255: I think to prevent this we should define a new type of attributeref specific to links and should not reuse an existing one. Also I don't see a reason why we wouldn't support 

    if ar.base.brief.is_some()
        || ar.base.note.is_some()
        || ar.base.examples.is_some()
        || !ar.base.annotations.is_empty()

on the attributes on links. Why not?

crates/weaver_resolved_schema/src/convert.rs
```diff
@@ -232,6 +232,85 @@ fn convert_attribute_ref<'a>(
...
+            ar.base.r#ref
+        )));
+    }
+    let Some((root, _)) = c.root_attribute(&ar.base.r#ref) else {
+        return Err(unsupported(format!(
+            "attribute '{}' not found in the catalog",
```
lmolkova, line 261: there is a special error for this, can we reuse it instead of unsupported?

crates/weaver_resolved_schema/src/convert.rs
```diff
@@ -232,6 +232,85 @@ fn convert_attribute_ref<'a>(
...
+            ar.base.r#ref
+        )));
+    };
+    let Some(base) = v2_catalog.convert_ref(root) else {
+        return Err(unsupported(format!(
+            "attribute '{}' could not be mapped to the v2 catalog",
```
lmolkova, line 267: why would it happen? doesn't convert_ref return an error of some sort we can pass directly?

crates/weaver_resolved_schema/src/convert.rs
```diff
@@ -232,6 +232,85 @@ fn convert_attribute_ref<'a>(
...
+    c: &V1Catalog,
+    v2_catalog: &V2CatalogBuilder,
+    provenance: &V2Provenance,
+) -> Result<Vec<span::SpanLink>, crate::error::Error> {
+    let mut links = Vec::new();
+    for link in g.span_links.iter() {
+        // Only locally declared links are validated: refinement links can
+        // be inherited from a dependency, so their targets may live elsewhere.
+        if validate_targets && !span_types.contains(&link.r#ref) {
+            return Err(crate::error::Error::SpanLinkTargetNotFound {
```
lmolkova, line 293: nit:

Do we need a special error here? Maybe we can reuse error that's used when span refinement can't find a base span?

crates/weaver_semconv/src/v2/span.rs
```diff
@@ -397,6 +401,37 @@ pub fn split_span_attributes_and_groups(
...
+    /// "opt_in") because a link, unlike a signal, can be required.
+    /// Defaults to 'recommended' when omitted.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub requirement_level: Option<RequirementLevel>,
+    /// Refines the brief description of the link.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub brief: Option<String>,
+    /// Refines the more elaborate description of the link.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub note: Option<String>,
```
lmolkova, line 426:
```suggestion
    /// The brief description of the link.
    #[serde(skip_serializing_if = "Option::is_none")]
    pub brief: Option<String>,
    /// The more elaborate description of the link.
    #[serde(skip_serializing_if = "Option::is_none")]
    pub note: Option<String>,
```

SpanLink declares the link rather than refining an existing link, so the doc comments should say "The brief description..." rather than "Refines...".

crates/weaver_resolved_schema/src/convert.rs
```diff
@@ -232,6 +232,85 @@ fn convert_attribute_ref<'a>(
...
+    validate_targets: bool,
+    c: &V1Catalog,
+    v2_catalog: &V2CatalogBuilder,
+    provenance: &V2Provenance,
+) -> Result<Vec<span::SpanLink>, crate::error::Error> {
+    let mut links = Vec::new();
+    for link in g.span_links.iter() {
+        // Only locally declared links are validated: refinement links can
+        // be inherited from a dependency, so their targets may live elsewhere.
+        if validate_targets && !span_types.contains(&link.r#ref) {
```
lmolkova, line 292: `validate_targets` is enabled for all non-refinement spans, which includes spans imported from dependencies. If an imported span links to a span in its source dependency that was not imported into this registry, resolution will fail with `SpanLinkTargetNotFound`.

Here is a test demonstrating the issue:

```rust
#[test]
fn test_imported_span_link_target_in_dependency_not_imported() {
    let dep_url = SchemaUrl::new("https://dep.example.com/schemas/1.0.0".to_owned()).unwrap();
    let mut imported_span = span_group_with_links(
        "dep.consumer",
        vec![link_to("dep.producer")], // target is defined in dependency, not imported locally
    );
    imported_span.lineage = Some(weaver_semconv::v1::group::GroupLineage::new(
        weaver_semconv::provenance::Provenance::new(dep_url.clone(), "dep.yaml"),
    ));

    let v1_catalog = crate::v1::catalog::test_utils::CatalogBuilder::default().build();
    let v1_registry = V1Registry {
        registry_url: "https://my.schema.url".to_owned(),
        entity_association_origins: Default::default(),
        groups: vec![imported_span],
    };
    let mut deps = BTreeSet::new();
    deps.insert(dep_url);

    let result = convert_v1_to_v2(v1_catalog, v1_registry, deps);
    assert!(
        result.is_ok(),
        "Imported span linking to an unimported dependency span must not fail target validation: {:?}",
        result.err()
    );
}
```

Suggestion: Only validate link targets for locally declared spans (`provenance.source.is_none()`).

crates/weaver_resolver/data/registry-test-v2-dep/span_links_registry/registry/registry.yaml
```diff
@@ -0,0 +1,37 @@
...
+  # references the same attribute so it is present in the catalog.
+  - type: consumer.op
+    kind: consumer
+    name:
+      note: Op span name.
+    brief: A local op span with links.
+    stability: stable
+    attributes:
+      - ref: linkserver.job.id
+        requirement_level: recommended
```
lmolkova, line 24: Resolving link attributes via `c.root_attribute` requires the attribute to already exist in the catalog. If a link uses an attribute from an imported dependency, it fails to resolve unless the span redundantly declares that attribute in its own `attributes` list (as done here on `consumer.op` and noted in comments above).

If we remove this redundant declaration:

```rust
#[test]
fn test_span_link_attribute_resolves_without_span_attribute_declaration() {
    let registry_yaml = r#"
file_format: definition/2
imports:
  spans:
    - linkserver.*
spans:
  - type: consumer.op
    kind: consumer
    name:
      note: Op span name.
    brief: A local op span with links.
    stability: stable
    links:
      - ref: linkserver.produce
        attributes:
          - ref: linkserver.job.id
            requirement_level: required
"#;

    // Resolving this against the published linkserver dependency currently fails with:
    // UnsupportedSpanLinkAttribute: "attribute 'linkserver.job.id' not found in the catalog"
}
```

Suggestion: Link attributes should be collected and resolved from the dependency catalog during registry resolution without requiring the span to declare them as span attributes.

lmolkova (commented):
Thanks a lot!! It looks great, a few comments on the datamodel and some resolution edge cases

jerbly:
I guess we should have a follow-up PR to add support to live-check and UI?

crates/weaver_resolver/data/registry-test-v2-dep/span_links_registry/published/resolved.yaml
vesari: Sure, removed it. Now the dedicated `LinkAttributeRef` type has no `sampling_relevant`. I agree on the whole-link sampling to be a follow-up.

crates/weaver_resolver/data/registry-test-v2-dep/span_links_registry/registry/registry.yaml
vesari: Ok, I dropped it. Requirement levels on link attributes remain, and an omitted level now falls back to the referenced attribute's level, per the `AttributeRef` contract.

schemas/semconv.schema.v2.json
vesari: Removed.

crates/weaver_resolved_schema/src/error.rs
vesari: Done.

vesari: Done.

crates/weaver_resolved_schema/src/error.rs
vesari: I removed it. Unknown names now fail in the resolver with the standard `UnresolvedAttributeRef` and the conversion keeps only `SpanLinkAttributeNotFound`, for the case where the catalog entry the resolver placed cannot be found.

crates/weaver_resolved_schema/src/convert.rs
vesari: Sure thing :) I've done both; link attributes now have a dedicated `LinkAttributeRef` type, and overrides are supported, i.e. link attributes now resolve like any other attribute reference, so an override produces its own catalog entry.

crates/weaver_resolved_schema/src/convert.rs
vesari: Done, unknown names now fail during resolution with the standard `UnresolvedAttributeRef`.

crates/weaver_resolved_schema/src/convert.rs
vesari: `convert_ref` returns an Option, so there is no inner error to pass through. That case now fails with the same `SpanLinkAttributeNotFound` as the lookup miss, and after the resolver change, it should not occur for a normally resolved registry.

crates/weaver_resolved_schema/src/convert.rs
vesari: I kept a dedicated error (with your message text) because it names the link target, which `RefinementBaseNotFound` cannot, but if you prefer I can merge them.

crates/weaver_semconv/src/v2/span.rs
vesari: Thanks for your suggestion, I applied it.

crates/weaver_resolved_schema/src/convert.rs
vesari: Oh yes, you're right. I followed your advice, and the target validation now runs only for spans the registry declares itself, using your `provenance.source.is_none()` predicate. Your test is in the suite, and it failed before the fix but it passes now.

crates/weaver_resolver/data/registry-test-v2-dep/span_links_registry/registry/registry.yaml
vesari: I followed your suggestion also here, and link attributes now resolve like any other attribute reference, including from the dependency catalog, so I got rid of the redundant declaration. The test registry reflects the failing case you mentioned, in that `consumer.op` does not declare `linkserver.job.id` as a span attribute anymore: the attribute now enters the catalog through the link reference alone.

vesari:
> I guess we should have a follow-up PR to add support to live-check and UI?

Yes, absolutely. I had already planned for a forge PR, and I guess UI can join it? As for live-check, maybe it will need yet another PR.

vesari:
One more note after syncing with main: #1748 has since moved the `span_name` carrier back to a v1 type with a conversion, while `span_links` still follows the #1763 pattern (v2 type carried directly). I can convert links to the split pattern too if that is the direction now, thanks for letting me know!

PR #1732, opened 2026-08-27
V1 to v2 refactoring

jsuereth:
Pull the trigger and *fully* duplicate all types for V2.

Post this PR we can do the following:

- Fix deprecations / definition issues in V1 since we aren't directly sharing types.
- "Drop" V1 completely once we've officially dropped support (this will be wellllll in the future, but planning now)
- Explicitly understand where any component in our system has a V1 or V2 dependency because the import is easy to grep for.


crates/weaver_semconv/src/v2/attribute.rs
```diff
@@ -3,18 +3,444 @@
...
+use crate::v2::{signal_id::SignalId, CommonFields};
+use crate::YamlValue;
+use weaver_common::ordered_float::OrderedF64;
+
+/// The different roles for attributes in groups.
+#[derive(
+    Serialize, Deserialize, Debug, Default, Clone, Eq, PartialEq, Hash, JsonSchema, PartialOrd, Ord,
+)]
+#[serde(rename_all = "snake_case")]
+pub enum AttributeRole {
```
lmolkova: do we need it in v2?

crates/weaver_semconv/src/convert.rs
```diff
@@ -0,0 +1,1811 @@
...
+    span::{
+        Span, SpanAttributeOrGroupRef, SpanAttributeRef, SpanKindSpec as V2SpanKindSpec,
+        SpanRefinement,
+    },
+    Imports as V2Imports, SemConvSpecV2,
+};
+
+/// Converts a V2 primitive or array type to V1.
+#[must_use]
+pub fn v2_primitive_or_array_type_to_v1(
```
lmolkova: which of these actually need to be public? 

AI tells 26 of the 40 `pub fn` have no caller outside this module today, and `v1_attribute_group_visibility_to_v2` (L943) has none at all.

lmolkova (approved):
Nice!

crates/weaver_semconv/src/convert.rs
```diff
@@ -0,0 +1,1811 @@
...
+#[must_use]
+pub fn v1_enum_entry_to_v2(e: V1EnumEntriesSpec) -> V2EnumEntriesSpec {
+    V2EnumEntriesSpec {
+        id: e.id,
+        value: v1_value_to_v2(e.value),
+        brief: e.brief,
+        note: e.note,
+        stability: e.stability,
+        deprecated: e.deprecated,
+        annotations: Default::default(),
```
jerbly:
```suggestion
        annotations: e.annotations,
```

crates/weaver_resolved_schema/src/convert.rs
```diff
@@ -0,0 +1,1640 @@
...
+            })
+        });
+        let mut lookup: BTreeMap<String, Vec<usize>> = BTreeMap::new();
+        for (idx, attr) in attributes.iter().enumerate() {
+            lookup.entry(attr.key.clone()).or_default().push(idx);
+        }
+        Self { attributes, lookup }
+    }
+
+    fn convert_ref(&self, attribute: &V1Attribute) -> Option<V2AttributeRef> {
```
jerbly, line 54: `convert_ref` matches a v1 attribute that has no `stability` against a catalog entry stored with `Stability::default()` (line 82 below). If that match ever fails, the lookup returns `None` and the attribute is dropped from the v2 signal with only a `log::info!` — silent loss rather than an error.

Three tests that guarded exactly this were removed here, and I couldn't find replacements:

- `v2/catalog.rs::test_lookup_defaults_missing_stability`
- `v2/mod.rs::test_convert_span_keeps_attribute_without_stability`
- `v2/mod.rs::test_convert_entity_keeps_attribute_without_stability`

The new tests in this file can't cover that path — 21 occurrences of `stability: Some(...)`, none of `stability: None`.

Two others also went without replacement:

- `test_span_refinement_name_note_propagates`, which covered the span kind/name hoisting this PR performs
- `test_convert_public_attribute_group_carries_requirement_level`

Deliberate, or fallout from moving code between modules? For a refactor this size the passing suite is most of the safety argument, so I'd like to see these ported.

crates/weaver_semconv/src/semconv.rs
```diff
@@ -416,18 +312,10 @@ impl SemConvSpecWithProvenance {
 
 #[cfg(test)]
 mod tests {
```
jerbly, line 314: A group of tests in this module were removed. I grepped for their assertion strings and they don't appear anywhere else:

- `test_error_message_bad_format`
- `test_error_message_invalid_v1`
- `test_error_message_invalid_unversioned`
- `test_error_message_invalid_format_2`
- `test_semconv_spec_with_provenance_from_file_v2`, which asserted the synthetic `registry.<provenance-name>` group id
- the 7-error validation assertion inside `test_semconv_spec_from_file_2`

`test_versioned_semconv` replaces them but only checks that a v1 and a v2 spec parse, so the error-path wording is no longer covered and `provenance_path_to_name` now has no test at all.

crates/weaver_live_check/src/lib.rs
```diff
@@ -136,10 +138,12 @@ impl VersionedAttribute {
 
     /// Get the type of the attribute
     #[must_use]
-    pub fn r#type(&self) -> &AttributeType {
+    pub fn r#type(&self) -> AttributeType {
```
jerbly: This went from `-> &AttributeType` to an owned `AttributeType`, so both arms now clone. For `AttributeType::Enum { members }` that deep-clones the whole `Vec<EnumEntriesSpec>` — ids, briefs, notes, annotations — on every call.

All three call sites are per-sample-attribute:

- `advice/type_advisor.rs:623`
- `advice/enum_advisor.rs:37`
- `sample_attribute.rs:209`

The last is the most wasteful: it clones the members vector only to evaluate `matches!(..., AttributeType::Template(_))`.

Live-check is the throughput-sensitive path, so `Cow<'_, AttributeType>` looks worth it here — borrow for v1, convert only for v2.

(`instrument()` below has the same shape, but `InstrumentSpec` is `Copy`, so that one is free.)

crates/weaver_semconv_gen/allowed-external-types.toml
```diff
@@ -5,13 +5,10 @@
...
     "weaver_common::*",
-    "weaver_forge::error::Error",
+    "weaver_forge::*",
     "miette::protocol::Diagnostic",
     "serde_core::ser::Serialize",
-    "weaver_resolved_schema::v2::ResolvedTelemetrySchema",
-    "weaver_forge::v2::registry::ForgeResolvedRegistry",
-    "weaver_forge::output_processor::OutputProcessor",
-    "weaver_resolved_schema::ResolvedTelemetrySchema",
+    "weaver_resolved_schema::*",
```
jerbly, line 11: This widens the guard rather than updating it: `weaver_resolved_schema::v2::ResolvedTelemetrySchema` and `weaver_resolved_schema::ResolvedTelemetrySchema` became `weaver_resolved_schema::*`, and `weaver_forge::error::Error` plus two others became `weaver_forge::*`. The file's own header asks for as few exceptions as possible; with wildcards, `cargo-check-external-types` will no longer flag a future leak from either crate.

The rename only needed a `v1::` inserted into the existing paths — `crates/weaver_emit/allowed-external-types.toml` in this same PR does exactly that.

crates/weaver_live_check/allowed-external-types.toml
```diff
@@ -8,7 +8,7 @@ allowed_external_types = [
     "miette::protocol::Diagnostic",
     "weaver_forge::*",
     "serde_json::value::Value",
-    "weaver_resolved_schema::attribute::Attribute",
+    "weaver_resolved_schema::*",
```
jerbly, line 11: Same as on `weaver_semconv_gen`: `weaver_resolved_schema::attribute::Attribute` became `weaver_resolved_schema::*`, where `weaver_resolved_schema::v1::attribute::Attribute` would have kept the guard intact.

jerbly (commented):
Four medium-severity notes from a review pass. None of them are about the approach — the v1/v2 split itself looks clean, and I diffed the moved v1 files line-by-line without finding behaviour changes. They're about coverage that came out with the move, one hot-path regression, and the external-types guards.

Checked locally on 946f3df: `cargo fmt --check` clean, `cargo clippy --all-targets --all-features` clean (no new warnings), `cargo nextest run --all` 1054/1054 green, and all eight CI-checked JSON schemas regenerate byte-identical.

crates/weaver_semconv_gen/allowed-external-types.toml
jsuereth: This was conscious choice on my part.  Make everything in resolved schema public - and consider the entire crate a transitive dependency.

I can trim back down - it's a decision we need to make as maintainers.

I think avoiding exposing unstable crates we use in our public APIs is good. However - in this case the two crates are always simultaneously released...

crates/weaver_live_check/allowed-external-types.toml
jsuereth: See comment above

crates/weaver_live_check/src/lib.rs
jsuereth: Fixed

crates/weaver_resolved_schema/src/convert.rs
jsuereth: Updated.

I removed one of the tests because we no longer error with "logic failure".

crates/weaver_semconv/src/semconv.rs
jsuereth: These tests moved into version specific counterparts.

crates/weaver_resolved_schema/src/convert.rs
```diff
@@ -0,0 +1,2553 @@
...
+        attributes.sort_by(|a, b| {
+            a.key
+                .cmp(&b.key)
+                .then_with(|| a.common.brief.cmp(&b.common.brief))
+                .then_with(|| a.common.note.cmp(&b.common.note))
+                .then_with(|| a.common.stability.cmp(&b.common.stability))
+                .then_with(|| a.common.deprecated.cmp(&b.common.deprecated))
+                .then_with(|| a.provenance.path.cmp(&b.provenance.path))
+                .then_with(|| a.provenance.source.cmp(&b.provenance.source))
+        });
```
jerbly: This sort became non-deterministic.

The chain covers `key, brief, note, stability, deprecated, provenance.path, provenance.source`, but not `r#type`, `examples`, or `annotations`.

The input is a `HashSet<V2Attribute>` (line 322), whose iteration order is randomised per process by `RandomState`, and `sort_by` is stable. So two attributes that tie on all seven fields above but differ in `examples` or `type` — e.g. a v1 `ref:` that overrides only `examples` — keep their randomised input order and can land either way round on any given run. That moves `attribute_catalog` entries and renumbers every `AttributeRef` in the emitted `resolved/2.0` output, which should surface as intermittent fixture flakes.

The `v2::catalog::Catalog` this replaces sorted by `(key, hash(attr))` over the whole attribute. `DefaultHasher` has fixed keys, so that was a total order and deterministic despite the `HashSet` input.

The ref renumbering already visible in `data/registry-test-v2-dep/deep_registry/expected-schema.yaml` is consistent with this.

Extending the `then_with` chain with `r#type`, `examples` and `annotations` would fix it, or keep hashing the whole attribute as a final tiebreak.

jerbly (approved):

crates/weaver_resolved_schema/src/convert.rs
jsuereth: Ok - using autogenerated sort now.

Issue #1744, opened 2026-09-02
weaver-installer.sh fails to detect x86_64-unknown-linux-gnu platform despite binary being available


lmolkova:
## Issue

The `weaver-installer.sh` script fails to install Weaver on Linux x86_64 runners with the error:

```
ERROR: there isn't a download for your platform x86_64-unknown-linux-gnu
```

However, the `weaver-x86_64-unknown-linux-gnu.tar.xz` binary **does exist** in the release artifacts for v0.26.0 and likely other versions.

## Environment

- **Platform**: Linux x86_64 (GitHub Actions Ubuntu runner)
- **Weaver Version**: v0.26.0
- **Setup Method**: Using `open-telemetry/weaver/.github/actions/setup-weaver@f8e1b42fbd1e09df5263c26b7df6011eaafd454d` action

## Steps to Reproduce

1. Run a GitHub Actions workflow on Ubuntu with the setup-weaver action
2. The action attempts to install Weaver v0.26.0 on Linux x86_64
3. The `weaver-installer.sh` script fails with platform detection error

## Root Cause

The installer script does not properly recognize or handle the `x86_64-unknown-linux-gnu` platform, even though the binary is available at:
```
https://github.com/open-telemetry/weaver/releases/download/v0.26.0/weaver-x86_64-unknown-linux-gnu.tar.xz
```

## Expected Behavior

The installer should successfully detect the Linux x86_64 platform and download the correct binary.

## Actual Behavior

The installer exits with:
```
ERROR: there isn't a download for your platform x86_64-unknown-linux-gnu
```

## Workaround

Direct binary download:
```bash
VERSION="v0.26.0"
ARCH="x86_64-unknown-linux-gnu"
URL="https://github.com/open-telemetry/weaver/releases/download/${VERSION}/weaver-${ARCH}.tar.xz"

mkdir -p ~/.cargo/bin
cd /tmp
curl -LsSf "$URL" | tar -xJ
mv weaver ~/.cargo/bin/
chmod +x ~/.cargo/bin/weaver
echo "$HOME/.cargo/bin" >> $GITHUB_PATH
```

## Impact

This breaks CI/CD workflows using the setup-weaver action on Ubuntu/Linux runners.

lmolkova:
This seems to be a regression introduced in https://github.com/open-telemetry/weaver/pull/1727


v0.26.0 run [33664316162](https://github.com/open-telemetry/weaver/actions/runs/33664316162):

[...]

PR #1747, opened 2026-09-02
ci: smoke-test the generated installers before publishing a release

lmolkova:
Adds a smoke-test-installers job that runs the freshly generated installers before a release is published, and gates host on it.

Motivated by #1744: v0.26.0 shipped an installer that knew only the Windows target, so every Unix install failed. Every job in that run was green - nothing checks that the installer we publish can install anything.

Between build-global-artifacts and host, on ubuntu x64/arm64, macOS arm64/x64 and Windows, the job serves the just-built artifacts over localhost, runs the real installer against them via the WEAVER_DOWNLOAD_URL override, and asserts weaver --version matches the tag being released. host now requires it, so a broken installer stops the release instead of shipping.

Verified by running the full release workflow on a fork: [all five jobs pass](https://github.com/lmolkova/weaver/actions/runs/33689073940).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -244,6 +244,87 @@ jobs:
           path: |
             ${{ steps.cargo-dist.outputs.paths }}
             ${{ env.BUILD_MANIFEST_NAME }}
+  # Run the generated installers against the freshly built artifacts, served from
+  # localhost, so a broken installer fails the release instead of shipping.
+  smoke-test-installers:
+    needs:
+      - build-local-artifacts
+      - build-global-artifacts
+    permissions:
+      "contents": "read"
+    timeout-minutes: 10
+    strategy:
+      fail-fast: false
+      matrix:
+        runner: ["ubuntu-24.04", "ubuntu-24.04-arm", "macos-15", "macos-15-intel", "windows-2025"]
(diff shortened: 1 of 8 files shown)


.github/workflows/release.yml
jsuereth, line 1: I think we need to move this into a config fragment and use it in cargo-dist:

jsuereth: https://axodotdev.github.io/cargo-dist/book/ci/customizing.html#hand-editing-releaseyml

jsuereth: Ok we probably want to use this feature - https://axodotdev.github.io/cargo-dist/book/ci/customizing.html#custom-jobs

lmolkova: thanks! I did add a custom job, but it still requires some manual patching because it only adds `needs: [plan, build-local-artifacts]`, but we also need `build-global-artifacts` to test generated artifacts.

It also injects `secrets: inherit` unconditionally, which makes zizmor unhappy.

These are now updated in xtask post-generation

jsuereth (approved):
Wow - we *need* to make some patches to cargo-dist I think.

Good work on the fix though! I think the maintenance pain is worth the protection.  Hopefully we dont' have to change dist workflows for a while.

PR #15, opened 2024-02-11
Fix up contributing document.

jsuereth:
I added information about how to get in touch/join the community directly.  Additionally, I copied otel-rust's contribuing gudieliens with a few TODO sections for us to fill out as we go and infrastructure comes along.

Issue #13, opened 2024-02-11
[Project Tracking] Prototype and Proposal for Milestone 1 - Fully resolved registry / telemetry schema


jsuereth:
From our [OTEP plan](https://github.com/open-telemetry/oteps/blob/main/text/0243-app-telemetry-schema-vision-roadmap.md#oteps), we have the first milestone to achieve:

>  Introduce a registry section with a list of fully resolved attributes, metrics, and other telemetry signals from the existing semantic convention registry. Upon implementation, the OpenTelemetry project's telemetry schema will include a registry of all standardized attributes, metrics, and signals as defined by the semantic conventions.

This is a meta issue tracking progress towards a prototype and proposal:

## Tasks

- [ ] #10 
- [ ] #9
- [ ] #8 
- [ ] #3

lquerel:
@jsuereth Yes!

PR #1748, opened 2026-09-03
More v1<->v2 splits

### LINKED ISSUE #1624: v2: remove deprecated from stability
https://github.com/open-telemetry/weaver/blob/3b517a1aa9129342d996c32d4992f33bdca1ea22/schemas/semconv.schema.v2.json#L1684-L1692

jsuereth:
Did some more thorough audits - as we had some "non v1" tagged structs that were not split in the initial PR.

- Stability
- Deprecation
- Manifest (new design)
- SignalRequirementLevel
- SchemaURL (not moved)

There's some interesting decisions to be had here on some of this, but at a minimum we can start removing fields in V2 we do not want, e.g. #1624

### Manifest Design

- `registry_manifest.yml` file is ALWAYS a v1 manifest - and resolves with v1 and allows legacy name/version fields which synthesize a schema_url.
- `manifest.yml` file ALWAYS means one of the two v2 manfiests (definition or published).   If `file_format` is missing, we assume it's a definition manifest.

### UI breakage

Also fixes an issue where API / UI was broken on deprecated field changes.

Fixes #1624


crates/weaver_resolver/src/lib.rs
```diff
@@ -428,7 +425,10 @@ impl WeaverResolver {
             }
         }
 
-        let manifest = repo.manifest().cloned();
+        let manifest = match repo.v2_manifest().transpose() {
+            Ok(m) => m,
+            Err(e) => return WResult::FatalErr(e.into()),
+        };
```
lmolkova: In `resolve_internal`, we are producing a V1 `ResolvedTelemetrySchema`. Converting `repo.v2_manifest()` (V1 -> V2) and then mapping `Into::into` back to V1 (V2 -> V1) is unnecessary and fails if a V1 manifest has deprecated stability. Using `repo.v1_manifest()` directly is simpler and avoids the roundtrip.

```suggestion
        let manifest = repo.v1_manifest();
```

crates/weaver_resolver/src/lib.rs
```diff
@@ -497,7 +497,7 @@ impl WeaverResolver {
             instrumentation_library: None,
             dependencies,
             versions: None,
-            registry_manifest: manifest,
+            registry_manifest: manifest.map(Into::into),
```
lmolkova:
```suggestion
            registry_manifest: manifest,
```

crates/weaver_semconv/src/v2/manifest.rs
```diff
@@ -349,14 +296,19 @@ impl RegistryManifest {
...
-        nfes.extend(
-            deserialization_warnings
-                .iter()
-                .map(|w| DeprecatedSyntaxInRegistryManifest {
+        for w in deserialization_warnings {
+            if w.starts_with("Missing 'file_format' field") {
+                nfes.push(MissingManifestFileFormat {
+                    path: manifest_path_buf.clone(),
+                    expected_format: DEFINITION_MANIFEST_FILE_FORMAT.to_owned(),
+                });
```
lmolkova: String matching on formatted error messages (`w.starts_with("Missing 'file_format' field")`) to reconstruct `MissingManifestFileFormat` is brittle. Consider passing structured warnings or errors directly into `nfes` instead of formatting into `deserialization_warnings: Vec<String>` and parsing it back.

crates/weaver_semconv/src/v1/manifest.rs
```diff
@@ -0,0 +1,565 @@
...
+            let legacy = file_name == LEGACY_REGISTRY_MANIFEST;
+            legacy
+        } else {
+            false
+        };
+
+        if is_legacy {
+            nfes.push(LegacyRegistryManifest {
+                path: manifest_path_buf.clone(),
+            });
```
lmolkova: Note: `LegacyRegistryManifest` (defined in `crates/weaver_semconv/src/lib.rs`, outside this diff) advises users: `Please rename it to manifest.yaml`. Since `manifest.yaml` is now strictly treated as a V2 manifest, simply renaming a V1 manifest file will break if it uses V1 properties (`semconv_version`, `schema_base_url`, or `name:` dependencies).

Consider updating the diagnostic message in `crates/weaver_semconv/src/lib.rs`:
`The registry manifest at {path:?} is using the legacy v1 filename. To migrate to v2, rename it to manifest.yaml and add 'file_format: definition_manifest/2.0'.`

crates/weaver_semconv/src/v2/manifest.rs
```diff
@@ -334,7 +279,9 @@ impl RegistryManifest {
 
         // Check if this is a legacy manifest file
         let is_legacy = if let Some(file_name) = manifest_path_buf.file_name() {
-            file_name == LEGACY_REGISTRY_MANIFEST
+            #[allow(deprecated)]
+            let legacy = file_name == LEGACY_REGISTRY_MANIFEST;
+            legacy
         } else {
             false
```
lmolkova: `ManifestPath::find` routes `LEGACY_REGISTRY_MANIFEST` directly to `ManifestPath::LegacyPath`, which invokes `V1RegistryManifest::try_from_file`. `V2RegistryManifest::try_from_file` will never be called on `registry_manifest.yaml` during resolution, making this legacy check dead code in V2.

crates/weaver_resolver/src/lib.rs
lmolkova: can it be simplified into

```suggestion
        let manifest = repo.v1_manifest();
```

?
(and line 500 also can be simplified)

crates/weaver_semconv/src/v2/manifest.rs
lmolkova: String matching on formatted error messages (`w.starts_with("Missing 'file_format' field")`) to reconstruct `MissingManifestFileFormat` is brittle. Consider passing structured warnings or errors directly into `nfes` instead of formatting into `deserialization_warnings: Vec<String>` and parsing it back.

crates/weaver_semconv/src/v2/manifest.rs
lmolkova: do we ever call this method with legacy file name now? seems to be a dead code.

crates/weaver_semconv/src/registry_repo.rs
```diff
@@ -5,50 +5,229 @@
...
+
+    /// Returns a reference to the V2 manifest, if this is a V2 manifest.
+    #[must_use]
+    pub fn as_v2(&self) -> Option<&V2RegistryManifest> {
+        match self {
+            VersionedManifest::V1(_) => None,
+            VersionedManifest::V2(v2) => Some(v2),
+        }
+    }
+
```
lmolkova: `as_v1` and `as_v2` appear to be unused across the repo. Can these be removed?

crates/weaver_semconv/src/registry_repo.rs
```diff
@@ -5,50 +5,229 @@
...
+        match self {
+            ManifestPath::LegacyPath(p) | ManifestPath::RegistryPath(p) => Some(p),
+            ManifestPath::None => None,
+        }
+    }
+
+    /// Attempts to load the manifest from this [`ManifestPath`].
+    pub fn try_into_manifest(&self, nfes: &mut Vec<Error>) -> Result<VersionedManifest, Error> {
+        VersionedManifest::try_from_file(self, nfes)
+    }
```
lmolkova: `try_into_manifest` appears to be unused across the repo.

crates/weaver_semconv/src/convert.rs
```diff
@@ -330,6 +345,133 @@ pub fn v2_span_name_to_v1(s: crate::v2::span::SpanName) -> V1SpanName {
...
+    fn from(d: V2Dependency) -> Self {
+        v2_dependency_to_v1(d)
+    }
+}
+
+/// Converts a V2 definition registry manifest to V1.
+#[must_use]
+pub fn v2_definition_manifest_to_v1(
+    m: V2DefinitionRegistryManifest,
+) -> V1DefinitionRegistryManifest {
```
lmolkova: Manifest and association conversion helpers (such as `v2_definition_manifest_to_v1`, `v1_definition_manifest_to_v2`, `v2_dependency_to_v1`, `v1_entity_association_to_v2`, `v1_signal_requirement_level_to_v2`) can be `pub(crate)` like the other internal helpers in this module.

crates/weaver_resolved_schema/src/convert.rs
```diff
@@ -185,6 +186,15 @@ fn fix_span_group_id(group_id: &str) -> SignalId {
     fix_group_id("span.", group_id)
 }
 
+fn convert_v1_stability(
```
jerbly: _Comment from Claude_

**A v1 registry with `stability: deprecated` on any attribute now fails every v1→v2 conversion.**

`.ok()` swallows the `Deprecated` error and `unwrap_or_default()` turns it into `Development`, so line 337 stores that attribute in the v2 catalog as `Development`. But `V2CatalogBuilder::convert_ref` (L54) matches a v1 attribute by converting the catalog entry's stability *back* to v1 and comparing (L68-77). `Development != Deprecated`, so the filter never matches, `convert_ref` returns `None`, and `convert_attribute_ref` (L241) raises `Error::AttributeNotFound`.

Reproduced on this branch with a throwaway unit test:
```
v1 `deprecated` is stored in the v2 catalog as: Development
convert_ref returned: None
```
`stability: deprecated` is still accepted by v1 validation (`v1/group.rs` L466/632/653 only warn) and was the normal pre-v2 spelling, so this is not an exotic input. It breaks `registry serve`, the MCP service, v2 forge templates, and any v2 registry depending on a v1 registry. Previously both sides were the same `Stability` type and `Deprecated` matched itself.

Same helper also silently rewrites group/signal `deprecated` → `development` at L400, 478, 514, 550, 622, 665.

Worth a regression test — the current suite is green because nothing covers a v1 registry carrying `stability: deprecated`.

crates/weaver_resolved_schema/src/convert.rs
```diff
@@ -68,11 +68,12 @@ impl V2CatalogBuilder {
                         && a.common.brief == attribute.brief
                         && a.common.note == attribute.note
                         && a.common.deprecated == attribute.deprecated
-                        && a.common.stability
-                            == *attribute
-                                .stability
-                                .as_ref()
-                                .unwrap_or(&weaver_semconv::stability::Stability::default())
+                        && weaver_semconv::v1::stability::Stability::from(
```
jerbly: _Comment from Claude_

This is the comparison side of the `AttributeNotFound` bug flagged at L189: the round-trip `v1 -> v2 -> v1` is lossy for `Deprecated`, so this equality can never hold for such an attribute.

Separately, converting `attribute.stability` once outside the closure (or a `PartialEq<V1Stability> for V2Stability`) would keep the predicate readable — it currently converts per candidate.

crates/weaver_semconv/src/convert.rs
```diff
@@ -945,27 +1138,175 @@ pub fn v1_span_kind_to_v2(k: V1SpanKindSpec) -> V2SpanKindSpec {
...
+pub fn v1_stability_to_v2(s: V1Stability) -> Result<V2Stability, Error> {
+    match s {
+        V1Stability::Stable => Ok(V2Stability::Stable),
+        V1Stability::Development => Ok(V2Stability::Development),
+        V1Stability::Alpha => Ok(V2Stability::Alpha),
+        V1Stability::Beta => Ok(V2Stability::Beta),
+        V1Stability::ReleaseCandidate => Ok(V2Stability::ReleaseCandidate),
+        // TODO: Should we consider this *STABLE* for v1->v2, since
+        // that was how things were treated in semconv prior to v2?
+        #[allow(deprecated)]
```
jerbly, line 1217: _Comment from Claude_

This `TODO` is an open design question shipping in merged code, and it's the root of the `AttributeNotFound` failure I flagged in `weaver_resolved_schema/src/convert.rs:189`.

Worth settling before merge: does v1 `deprecated` map to `Stable` (how semconv treated it pre-v2), or stay an error? If it stays an error, the three callers need to agree — right now one errors (manifest), one silently defaults to `Development` (catalog), and one silently drops to `None` (enum entries, L1050).

crates/weaver_semconv/src/convert.rs
```diff
@@ -854,7 +1047,7 @@ pub(crate) fn v1_enum_entry_to_v2(e: V1EnumEntriesSpec) -> V2EnumEntriesSpec {
         value: v1_value_to_v2(e.value),
         brief: e.brief,
         note: e.note,
-        stability: e.stability,
+        stability: e.stability.and_then(|s| v1_stability_to_v2(s).ok()),
```
jerbly: _Comment from Claude_

`.ok()` here turns an enum member declared `stability: deprecated` into `stability: None` with no diagnostic, while the sibling manifest conversions surface the same input as an error. The deprecation signal is lost silently.

crates/weaver_semconv/src/registry_repo.rs
```diff
@@ -5,50 +5,206 @@
...
+/// Finds the path to the manifest file, which could be:
 /// - directly the path to the manifest file, or
-/// - either `manifest.yaml` or `registry_manifest.yaml` in the given directory.
-/// - None otherwise.
-fn find_manifest_path(registry_path: &Path) -> Option<PathBuf> {
-    // First check to see if we're pointing at a manifest.
+/// - `manifest.yaml` in the given directory (`RegistryPath`), or
+/// - `registry_manifest.yaml` in the given directory (`LegacyPath`), or
+/// - `None` otherwise.
+fn find_manifest_path(registry_path: &Path) -> ManifestPath {
```
jerbly, line 174: _Comment from Claude_

**Following weaver's own rename advice breaks a v1 registry.**

The manifest schema version is now derived from the *filename*: only `registry_manifest.yaml` reaches the v1 parser. A v1 manifest renamed to `manifest.yaml` — exactly what the new `LegacyRegistryManifest` deprecation warning instructs — hits the v2 parser, which has no `semconv_version`/`schema_base_url` and rejects `name`-based dependencies.

Reproduced using this repo's own `data/v1-dependency-syntax/main` fixture, renamed to `manifest.yaml`:
```
x The registry manifest at ".../manifest.yaml" is invalid. missing required field 'schema_url'
```
Same applies to any file passed directly as `--registry path/to/x.yaml` (L184).

crates/weaver_semconv/src/v1/manifest.rs
```diff
@@ -0,0 +1,552 @@
...
+    #[serde(default)]
+    stability: Stability,
+}
+
+impl RawManifestFields {
+    fn into_manifest(
+        self,
+        path: &std::path::Path,
+        nfes: &mut Vec<Error>,
+    ) -> Result<RegistryManifest, Error> {
```
jerbly, line 148: _Comment from Claude_

Rejecting *any* `file_format` here closes the last migration path. Combined with the filename-based routing in `registry_repo.rs:172`, all three moves a user might make are hard errors:

1. Keep `registry_manifest.yaml`, add the `file_format` the new `MissingManifestFileFormat` warning asks for -> `Unknown file_format 'definition_manifest/2.0'. Expected no file_format for a definition manifest.`
2. Rename to `manifest.yaml`, keep v1 syntax -> `missing required field 'schema_url'`
3. Keep `registry_manifest.yaml` holding a publication manifest (parsed fine on `main`) -> `Unknown file_format 'manifest/2.0'.`

All three reproduced on this branch. If (3) is intended per the design note, the message should at least say "rename this file to `manifest.yaml`" rather than calling the file invalid.

crates/weaver_semconv/src/v1/manifest.rs
```diff
@@ -0,0 +1,552 @@
...
+
+/// Parses a dependency declaration, discriminating on which identifying field it declares.
+/// A v1 manifest may declare a dependency by `name`, which then gets a placeholder, unversioned schema URL minted
+/// from that name.
+fn parse_dependency(value: serde_yaml::Value) -> Result<Dependency, String> {
+    #[derive(Deserialize)]
+    #[serde(deny_unknown_fields)]
+    struct NamedDependency {
+        name: String,
+        registry_path: VirtualDirectoryPath,
```
jerbly, line 104: _Comment from Claude_

Dead branch: in the `else` arm `name` is `None` by construction, so `subject` is always `"a dependency"` and the `|name| format!("dependency '{name}'")` arm is unreachable — the error can never name the entry (e.g. `- name: 123`).

The v2 copy has no `if name.is_some()` early return, so the same `map_or_else` works as intended there.

crates/weaver_semconv/src/v1/manifest.rs
```diff
@@ -0,0 +1,552 @@
...
+        let file = std::fs::File::open(path).map_err(|e| InvalidRegistryManifest {
+            path: manifest_path_buf.clone(),
+            error: e.to_string(),
+        })?;
+        let reader = std::io::BufReader::new(file);
+        let raw: RawManifestFields =
+            serde_yaml::from_reader(reader).map_err(|e| InvalidRegistryManifest {
+                path: manifest_path_buf.clone(),
+                error: e.to_string(),
+            })?;
```
jerbly, line 231: _Comment from Claude_

Minor: `manifest_path_buf.file_name().is_some_and(|n| n == LEGACY_REGISTRY_MANIFEST)` says this in one line — and that's exactly the idiom `registry_repo.rs:170` uses for the same test. `ManifestPath` has also already classified this as `LegacyPath` before `try_from_file` is called, so the check is re-derived.

jerbly (commented):
_Comment from Claude_

Review of the Rust and UI changes on this branch, at `merge-base(main, HEAD)`. I ran the project gates locally: `cargo fmt --all --check` clean, `cargo clippy --all-targets --all-features` clean, `cargo nextest run --all` **1237 passed / 1 skipped**, and all 8 generated schemas plus `docs/usage.md` regenerate byte-identical. The pre-existing `Stability::Deprecated` clippy warning is gone, since this PR removes the variant.

The split itself looks thorough and the manifest design is coherent. Two things I'd want resolved before merge:

1. **`crates/weaver_resolved_schema/src/convert.rs:189`** — a v1 registry with `stability: deprecated` on any attribute now fails every v1→v2 conversion with `AttributeNotFound`. Reproduced locally. The green suite is misleading here: nothing covers that input.
2. **The manifest migration has no working path** — all three moves a user might make (`registry_manifest.yaml` + `file_format`, rename to `manifest.yaml` keeping v1 syntax, or an existing `registry_manifest.yaml` holding a publication manifest) are hard errors, and none of the messages says what to do. Details on `registry_repo.rs:172` and `v1/manifest.rs:149`.

Both touch the "v1 is frozen" rule in `CLAUDE.md`.

Smaller note that has no diff line to hang off: `schemas/semconv-syntax.v2.md:60` still says dependencies are declared in `registry_manifest.yaml`, which is now the v1-only name.

On `unwrap()` in the new tests (CLAUDE.md #5 asks for `expect()`): there are 42 new occurrences, all inside `#[cfg(test)]`. Flagging for completeness rather than as a change request — `main` already carries ~795, so this matches local precedent.

crates/weaver_semconv/src/registry_repo.rs
```diff
@@ -5,50 +5,206 @@
...
+        nfes: &mut Vec<Error>,
+    ) -> Result<Self, Error> {
+        match manifest_path {
+            ManifestPath::LegacyPath(path) => {
+                V1RegistryManifest::try_from_file(path, nfes).map(VersionedManifest::V1)
+            }
+            ManifestPath::RegistryPath(path) => {
+                V2RegistryManifest::try_from_file(path, nfes).map(VersionedManifest::V2)
+            }
+            ManifestPath::None => Err(Error::RegistryManifestNotFound {
```
jerbly: _Comment from Claude_

`PathBuf::new()` here surfaces as `The registry manifest at "" was not found`. `try_from_path` has the real path in hand and could pass it through.

crates/weaver_semconv/src/registry_repo.rs
```diff
@@ -5,50 +5,206 @@
...
+
+    /// Returns the dependencies converted to V2 dependencies.
+    #[must_use]
+    pub fn dependencies(&self) -> Vec<Dependency> {
+        match self {
+            VersionedManifest::V1(v1) => v1
+                .dependencies()
+                .iter()
+                .cloned()
+                .map(crate::convert::v1_dependency_to_v2)
```
jerbly: _Comment from Claude_

Nit (CLAUDE.md #15, `use` imports over qualified paths): `crate::convert::*` inline here and at L126/L133; L358 spells out `crate::v2::manifest::DefinitionRegistryManifest` though L10 already imports from that module.

crates/weaver_semconv/src/v2/manifest.rs
```diff
@@ -231,71 +191,52 @@ impl RawManifestFields {
...
 
-/// A registry manifest that can be either a definition or a publication manifest.
+/// A registry manifest that can be either a definition or a publication manifest (version 2).
 ///
 /// The `file_format` field is the discriminator:
 /// - `"manifest/2.0"` → [`PublicationRegistryManifest`]
-/// - absent → [`DefinitionRegistryManifest`]
-#[derive(Debug, Clone, JsonSchema)]
+/// - `"definition_manifest/2.0"` → [`DefinitionRegistryManifest`]
+#[derive(Serialize, Deserialize, Debug, Clone, JsonSchema)]
```
jerbly, line 255: _Comment from Claude_

This derived `Deserialize` is unused (`try_from_file` goes through `RawManifestFields`) and would be wrong if used: `#[serde(untagged)]` tries variants in order, and `DefinitionRegistryManifest` has no `deny_unknown_fields`, so a publication manifest deserializes cleanly into `Definition` with `file_format: "manifest/2.0"` and `resolved_registry_uri` silently dropped. Same shape in `v1/manifest.rs:198`.

Related: the `#[serde(alias = "resolved_schema_uri")]` on `PublicationRegistryManifest` (L334) is only reachable through this derive, and would accept the deprecated field *without* the warning that `into_manifest` (L166-178) raises — two mechanisms for one migration, one dead and silent.

Suggest dropping the derive, or ordering `Publication` first with `deny_unknown_fields`.

crates/weaver_semconv/src/convert.rs
```diff
@@ -324,19 +341,163 @@ pub fn v2_span_kind_to_v1(k: V2SpanKindSpec) -> V1SpanKindSpec {
...
+    match s {
+        V2Stability::Stable => V1Stability::Stable,
+        V2Stability::Development => V1Stability::Development,
+        V2Stability::Alpha => V1Stability::Alpha,
+        V2Stability::Beta => V1Stability::Beta,
+        V2Stability::ReleaseCandidate => V1Stability::ReleaseCandidate,
+    }
+}
+
+impl From<V2Stability> for V1Stability {
```
jerbly, line 389: _Comment from Claude_

CLAUDE.md #13 (prefer `From`/`Into` over standalone conversion fns) looks half-applied: this file adds ~10 `From`/`TryFrom` impls that each just delegate to the existing free fn, and keeps both. Production callers still use the free fns (`registry_repo.rs:115/126/133`, `imports.rs`), so the trait impls are exercised only by tests.

Either make the impl the definition and drop the free fn, or skip the impls — as-is it's two public spellings of each conversion.

crates/weaver_live_check/src/live_checker.rs
```diff
@@ -1038,6 +1038,7 @@ mod tests {
                 dependency_graph: Default::default(),
             }))
         } else {
+            use weaver_semconv::v1::stability::Stability;
```
jerbly: _Comment from Claude_

There are seven of these function-local `use ...v1::stability::Stability;` inside `else` branches (L1041, 1286-1287, 1495, 2095, 2727, 4072), so `Stability` names a different type in different branches of the same function.

`convert.rs` solves the identical v1/v2 collision with module-level `as V1Stability` / `as V2Stability` aliases — worth matching that here.

crates/weaver_forge/src/lib.rs
```diff
@@ -2079,4 +2079,140 @@ mod tests {
             );
         }
     }
+
+    #[test]
```
jerbly: _Comment from Claude_

These two new tests are ~95% identical — the 30-line `root_schema` literal and the `v1_schema` literal differ only in two URL strings. A shared builder taking the schema URL would cut most of it (CLAUDE.md #1).

Also L2210-2215: `assert!(result.is_fatal())` is subsumed by the `if let ... else { panic!() }` that follows; `assert!(matches!(result, WResult::FatalErr(Error::SchemaError(_))))` says it once.

ui/src/components/StabilityBadge.tsx
```diff
@@ -54,3 +53,48 @@ export function StabilityDot({ stability }: StabilityBadgeProps) {
...
+  }
+
+  return (
+    <span className={`badge badge-error${sizeClass}`} title={title}>
+      Deprecated
+    </span>
+  )
+}
+
+export function DeprecatedDot({ deprecated }: DeprecatedBadgeProps) {
```
jerbly: _Comment from Claude_

`DeprecatedDot` is exported but never imported anywhere in `ui/src` or `ui/e2e`. Its sibling `StabilityDot` is used by `NamespaceTree.tsx`; this one looks like dead code.

ui/src/routes/search.tsx
```diff
@@ -75,9 +80,16 @@ const parseStabilityFilter = (value: string | null): StabilityFilter =>
 const parseViewMode = (value: string | null): ViewMode => (value === 'tree' ? 'tree' : 'list')
 
 // Deprecated items are hidden by default, so the URL only carries a param
-// when the user opts in to showing them - same "only serialize non-default
-// state" convention as the rest of this file's filters.
-const parseHideDeprecated = (value: string | null): boolean => value !== 'show'
+// when the user opts in to showing them or filtering exclusively for them.
```
jerbly, line 98: _Comment from Claude_

This maps legacy `?stability=deprecated` bookmarks onto `deprecated=only` (correctly — `parseStabilityFilter` returns `null` for it now that it's out of `stabilityOptions`), but the same PR deletes `render.attr.stability_deprecated` from the e2e fixture, so nothing exercises the path.

src/serve/handlers.rs
```diff
@@ -272,12 +272,17 @@ pub async fn search_registry(
 ) -> impl IntoResponse {
     // Convert Option<String> to Option<&str> for search
     let query = params.q.as_deref();
+    let deprecated = params.deprecated.or(if params.hide_deprecated {
```
jerbly: _Comment from Claude_

Nit: reads more directly as `params.deprecated.or_else(|| params.hide_deprecated.then_some(false))`. Also `hide_deprecated`'s doc in `src/serve/types.rs:41` still reads as the primary knob — worth marking it the legacy shim.

ui/e2e/smoke.spec.ts
```diff
@@ -41,28 +41,61 @@ test('searching and clicking a result opens a detail page', async ({ page }) =>
...
 
-test('deprecated items are hidden by default and revealed by the eye toggle', async ({
+test('deprecated items are hidden by default and controlled by hide/show/only buttons', async ({
   page,
 }) => {
   await page.goto('/search')
   await expect(page.locator('a.card').first()).toBeVisible()
 
+  // Verify the stability dropdown does not include Deprecated.
+  const stabilitySelect = page.locator('select').nth(1)
```
jerbly: _Comment from Claude_

Nit: `page.locator('select').nth(1)` is positional and breaks if another `<select>` is added ahead of it. A role/label locator would be stable.

crates/weaver_resolved_schema/src/convert.rs
lmolkova: I think it's working as intended. WE deprecated deprecated stability even before weaver and it should not leak into v2. 

it's not in the schema or docs

https://github.com/open-telemetry/weaver/blob/bab3a7afdc730a9fb930f27668c1cf5520c9cf60/schemas/semconv.schema.json#L670
https://github.com/open-telemetry/weaver/blob/main/schemas/semconv-syntax.md#syntax

it's not in semconv, we even warn on it https://github.com/open-telemetry/weaver/blob/bab3a7afdc730a9fb930f27668c1cf5520c9cf60/crates/weaver_semconv/src/v1/group.rs#L379-L386 and fail with `--future`

crates/weaver_resolver/src/lib.rs
jsuereth: Should be fixed

crates/weaver_semconv/src/v2/manifest.rs
jsuereth: I think you're right, should be fixed

crates/weaver_semconv/src/registry_repo.rs
jsuereth: Right - too much shuffling of code, should be cleaned up.

crates/weaver_semconv/src/v2/manifest.rs
jsuereth: These are now ordered.  I'm not sure I believe Claude fully here - or don't believe the risk (perhaps it is a risk with more LLM code).

TL;DR; I'd rather have these types support serialization and do it right. I don't think folks will be interacting directly with this type, but with our publicly exposed resolution algorithms and entry points.

However - I also think not having this derive deserialize means some logic in those other palces becomes problematic.

In any case, re-ordered here.

crates/weaver_semconv/src/convert.rs
jsuereth: Fixed

crates/weaver_semconv/src/convert.rs
jsuereth: Settled - fixed.

crates/weaver_semconv/src/registry_repo.rs
jsuereth: This is intentional.

The rename advice is part of moving to "v2" manifest file definition format.

You can use a v2 manifest with v1 definitions right now.  Upgrade is purely incremental file-by-file

ui/src/routes/search.tsx
jsuereth: Added new tests - also manually tested and I think the new flows/options are better for how stability works in practice in semconv.

crates/weaver_semconv/src/v2/manifest.rs
jsuereth: Right - that was ugly, sorry I thought I had pushed my fix to clean it up, but it was sitting in my repo.  Fixed - this doesn't try to do any "vec of warnings" shenigans inside structs anymore.

crates/weaver_semconv/src/registry_repo.rs
jsuereth: They can be - removed

crates/weaver_semconv/src/convert.rs
jsuereth: They're now "From" types so you can use `.into()` or `.from()` like canonical rust.

crates/weaver_resolved_schema/src/convert.rs
jsuereth: 100% - I think claude is overreacting a bit -  or at least doesn't understand the state of V1 vs. V2.

crates/weaver_forge/src/lib.rs
jsuereth: fixed

ui/src/components/StabilityBadge.tsx
jsuereth: Fixed

src/serve/handlers.rs
jsuereth: Fixed

ui/e2e/smoke.spec.ts
jsuereth: Fixed

jerbly:
I think this is fine apart from the new sorting feature in the UI:

The sort runs on the client, over the current page. In list view the API returns 50 results per page, so "Deprecated first", "Name" and "Stability" reorder that page only, not the whole result set. On a real registry, page 1 of "Deprecated first" can show no deprecated items even when later pages have them. The e2e test passes only because the render-registry fixture fits on one page.

Also:
- The dropdown shows in tree view but does nothing there.
- "Name" and "Stability" sorting have no e2e coverage.

jerbly (approved):
LGTM

PR #1749, opened 2026-09-03
Record OTLP context on live-check samples

manolama:
Populate OTLP identity and timing fields directly on live-check
    samples and serialize them in reports by default. Document the
    report format and cover default propagation, finding output,
    and distinct resource/scope ordering.

### DIFF
--- a/crates/weaver_live_check/README.md
+++ b/crates/weaver_live_check/README.md
@@ -287,6 +287,34 @@ To provide your own custom templates use the `--templates` option.
 
 As mentioned, the exit-code is set non-zero if any `violation` finding is provided in the output. This can be used in tests and/or CI to fail builds for example.
 
+### OTLP context in reports
+
+When live check receives OTLP, it records the full telemetry fields on the
+corresponding live-check samples in the report:
+
+```sh
+weaver registry live-check --format json --output ./outdir
+```
+
+Spans include `trace_id`, `span_id`, `parent_span_id`, `trace_state`,
+`start_time`, and `end_time`. Span events include `timestamp`, span links
+include their linked `trace_id` and `span_id`, logs include their trace and
+span IDs plus `timestamp`, and metric data points include `start_time` and
+`end_time`. Timestamps use RFC 3339 format.
+
+Resource and instrumentation-scope data is emitted as `resource` and
+`instrumentation_scope` samples before the signals that share it, rather than
+being repeated on every span, log, or metric. The serialized signal samples do
+not contain a separate reference to those entries, so consumers that need this
+provenance must retain that ordering and grouping.
+
+This is Weaver's live-check report format (JSON, YAML, or JSON Lines), not
+OTLP JSON or OTLP protobuf, and it is not directly ingestible by an OTLP
+backend. It is intended to supplement validation findings with enough source
+context for report consumers. To inspect findings in a standard OpenTelemetry
+backend, enable [`OTLP Log Record Emission`](#otlp-log-record-emission), which
+emits the findings themselves as OTLP log records.
+
 ### Statistics
 
 A statistics entity is produced when the input is closed like this snippet:
--- a/crates/weaver_live_check/src/otlp_logger.rs
+++ b/crates/weaver_live_check/src/otlp_logger.rs
@@ -2,8 +2,8 @@
 
 //! OTLP logger provider for emitting policy findings as log records.
 
-use opentelemetry::logs::{AnyValue, Logger, LoggerProvider, Severity};
-use opentelemetry::{Key, KeyValue};
+use opentelemetry::logs::{AnyValue, LogRecord as _, Logger, LoggerProvider, Severity};
+use opentelemetry::{Key, KeyValue, SpanId, TraceId};
 use opentelemetry_otlp::WithExportConfig;
 use opentelemetry_sdk::logs::SdkLoggerProvider;
 use opentelemetry_sdk::resource::ResourceDetector;
@@ -136,6 +136,10 @@ impl OtlpEmitter {
             signal_type.as_ref(),
         );
 
+        if let Some((trace_id, span_id)) = trace_context_from_sample(parent_signal) {
+            log_record.set_trace_context(trace_id, span_id, None);
+        }
+
         logger.emit(log_record);
     }
 
@@ -152,6 +156,13 @@ impl OtlpEmitter {
             error: format!("Failed to shutdown OTLP log provider: {e}"),
         })
     }
+}
+
+fn trace_context_from_sample(sample: &Sample) -> Option<(TraceId, SpanId)> {
+    let (trace_id, span_id) = sample.trace_context_ids()?;
+    TraceId::from_hex(trace_id)
+        .ok()
+        .zip(SpanId::from_hex(span_id).ok())
 }
 
 impl From<&FindingLevel> for GeneratedFindingLevel {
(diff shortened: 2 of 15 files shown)


jerbly:
I had a quick look. Before going more in depth I think the design here is not quite right, but just because of the way live-check has grown from it's initial implementation. Rather than create a separate object for this data we can add the missing fields to the sample objects - the fields on them today were fairly arbitrarily chosen.

Secondly, do we need to repeat the resource and scope on every signal sample? Is there a lighter way to achieve this?

And finally, you mention in your proposal a few times that weaver is not a tee?? Is there some other solution here entirely that would be an even better fit?

manolama:
Thanks for the check @jerbly. Sorry, my Claude pushed before I could clean it up some more.
Now it appends the data in same live-check report alongside the samples, but in a separate object to avoid conflict with existing fields.
I would like to keep the resource scope on each sample since we can have trace scenarios where we validate telemetry from different systems, containers, etc. That way we keep the original lined up with the raw telemetry. Otherwise it would need a dictionary approach that may be too cumbersome at this point as I wouldn't expect the scenarios to be so huge it would matter.
The tee aspect (yanked that from the PR) was from the original attempt I took at teeing the data in the https://github.com/open-telemetry/semantic-conventions-conformance repo since I wasn't sure ya'll would want it in Weaver itself. I do agree with @lmolkova that it's more useful here.

jerbly:
1. Rather than make a new SampleContext struct, just populate all the Sample* structs with the missing fields.
2. For Resource and InstrumentationScope there are some options:
  a. Today they land in the JSON output in the order they've arrived and would not be interleaved with other incoming OTLP from other sources, so your parser would be able to associate them correctly.
  b. Restructure the Sample JSON so the nesting is preserved:
```
Resource
-- InstrumentionScope
---- Span
---- Span
```
3. Use the OTLP emitted findings instead of the JSON report. Once you have done step (1) above we could associate the findings with the span_id, trace_id. An Observability backend would then give you the UI with waterfall view etc. and weaver's findings in place. (This is a feature I wanted regardless for people doing continuous evaluation with live-check in staging environments).

docs/usage.md
```diff
@@ -522,6 +522,10 @@ Includes: Flexible input ingestion, configurable assessment, and template-based
 
   Possible values: `true`, `false`
 
+* `--capture-telemetry <CAPTURE_TELEMETRY>` — Capture raw OTLP telemetry into the report, for consumers that would like to visualize the original data used to generate the report. Off by default: report output is unchanged unless this is set. [default: false]
```
kamphaus: Is there any doc on where to find the captured telemetry?

kamphaus: In what format is the raw OTLP telemetry written?
Is it [OTLP in JSON encoding](https://opentelemetry.io/docs/specs/otlp/#json-protobuf-encoding) or the OTLP binary protobuf encoding?
If it is a custom encoding, then it would be more difficult to visualize in any standard tool, requiring some custom post-processing.

manolama:
> 1. Rather than make a new SampleContext struct, just populate all the Sample* structs with the missing fields.

Ack, done in this split commit.

> 2. For Resource and InstrumentationScope there are some options:
a. Today they land in the JSON output in the order they've arrived and would not be interleaved with other incoming OTLP from other sources, so your parser would be able to associate them correctly.

That's fine then, using the ordering.

> 3. Use the OTLP emitted findings instead of the JSON report. Once you have done step (1) above we could associate the findings with the span_id, trace_id. An Observability backend would then give you the UI with waterfall view etc. and weaver's findings in place. (This is a feature I wanted regardless for people doing continuous evaluation with live-check in staging environments).

Trace and span IDs are now correctly propagated with the Weaver OLTP logs.

I still really want to have that capture flag so that telemetry is serialized on the final JSON report. The reason being, Trask had a nice, simple little UI on his original conformance site. It just read parsed the JSON report without needing a full telemetry backend. I'd like to do the same with the official repo https://github.com/open-telemetry/semantic-conventions-conformance where folks can point to a specific scenario to see how traces (and eventually logs and metrics) look as _if_ they were in a real backend. Does that part make sense?

docs/usage.md
manolama: Got catch, updated.

manolama: Updated the docs too but the idea here is that the data is serialized in full in the final report samples. There was already some serialization of fields but mostly focused on the attributes. Now the whole payload can be recorded in the JSON report.

crates/weaver_config/src/live_check.rs
```diff
@@ -110,6 +110,10 @@ pub struct LiveCheckConfig {
     /// Disable statistics accumulation. Useful for long-running live-check sessions.
     pub no_stats: bool,
 
+    /// Capture raw OTLP context (trace/span identity, timestamps, and the
+    /// resource/scope each sample belongs to) into the report.
+    pub capture_telemetry: bool,
+
```
jerbly: Let's just remove this option. This PR is just about adding the missing fields that should have been on the Sample* structs from the beginning. Keeps it really simple.

crates/weaver_live_check/src/otlp_logger.rs
```diff
@@ -154,6 +158,13 @@ impl OtlpEmitter {
     }
 }
 
+fn trace_context_from_sample(sample: &Sample) -> Option<(TraceId, SpanId)> {
+    let (trace_id, span_id) = sample.trace_context_ids()?;
+    TraceId::from_hex(trace_id)
+        .ok()
+        .zip(SpanId::from_hex(span_id).ok())
+}
+
```
jerbly, line 167: Nice! People that use weaver in a continuous mode can now have findings attached to the signal that produced it. This is a really nice feature.

jerbly:
> I still really want to have that capture flag so that telemetry is serialized on the final JSON report.

Let's not have the flag. Live-check should provide all this info as standard. This will simplify the code significantly too.

This is a really great improvement, thanks. Nearly there...

crates/weaver_config/src/live_check.rs
manolama: done, thanks!

crates/weaver_live_check/README.md
```diff
@@ -287,6 +287,35 @@ To provide your own custom templates use the `--templates` option.
...
+
+When live check receives OTLP, it records the full telemetry fields on the 
+corresponding live-check samples in the report:
+
+```sh
+weaver registry live-check --format json --output ./outdir
+```
+
+The fields are inline on the relevant item in the report's `samples` list;
+there is no separate telemetry section. Spans include `trace_id`, `span_id`,
```
jerbly:
```suggestion
Spans include `trace_id`, `span_id`,
```

jerbly (approved):
Please add a Changelog entry and then I think this is good to go. Thanks!

src/registry/otlp/conversion.rs
```diff
@@ -181,6 +181,22 @@ fn otlp_data_to_data_points(data: &Option<Data>) -> Option<DataPoints> {
...
+) -> (Option<String>, Option<String>) {
+    (
+        optional_unix_nanos_to_utc(start_time_unix_nano),
+        optional_unix_nanos_to_utc(time_unix_nano),
+    )
+}
+
+/// Builds a span's raw context (trace/span/parent identity, tracestate,
+/// start/end time). Resource and instrumentation scope are filled in by the
+/// caller, which holds them as `Rc`s shared with every other span in the
```
lmolkova: the sentence looks incomplete

lmolkova (approved):
Thanks!

Could you please also update the PR title and description to match the implementation?

kamphaus (approved):

PR #1760, opened 2026-09-08
feat(live-check): support virtual-directory format for --advice-policies and --advice-data

### LINKED ISSUE #1751: Support the virtual-directory format for --advice-data and --advice-policies
`--registry` and `--policy` take `source[@refspec][[sub_folder]]`, so they can come from a git repo or a remote archive. `--advice-data` and `--advice-policies` are local paths only.

They should go through `VirtualDirectory` too.

lmolkova:
Resolves #1751

Support virtual directory formats (archives, git repos, etc.) and auth for `--advice-policies` and `--advice-data` in `live-check` and `mcp` commands.

### DIFF
--- a/crates/weaver_common/src/vdir.rs
+++ b/crates/weaver_common/src/vdir.rs
@@ -322,9 +322,10 @@ static REGISTRY_REGEX: Lazy<Regex> = Lazy::new(|| {
 /// Paths may optionally specify:
 /// - A sub-folder within the archive or repository via `[sub_folder]`
 /// - [Not Yet Implemented] A specific Git refspec (branch, tag, or commit) via `@refspec`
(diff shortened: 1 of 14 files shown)


jerbly:
Should we solve #1362 first?

lmolkova:
> Should we solve #1362 first?

I don't think they are related to each other. This is to avoid hacks in conformance tests where I'd like to leverage conformance repo custom policies and data from semconv. I don't depend on default policies (which validate what weaver registry check would do on definitions). 

I'm not sure default policies are helpful in live-check.

jerbly (approved):
> I'm not sure default policies are helpful in live-check.

Ha! Seems like we'll never agree on this one. :)

This PR just needs a changelog then it's good.

PR #1763, opened 2026-09-10
Formal span name definition for validation and code generation in semconv v2

### LINKED ISSUE #819: Define span names in yaml

lmolkova:
Fixes #819 

Adds formal span name templates to semantic conventions v2, enabling live-check validation and code generation.

Design document: [`docs/specs/span-naming/span_naming.md`](https://github.com/open-telemetry/weaver/blob/979bc9b0754495f4465ae1bd8cb45987dc7453d9/docs/specs/span-naming/span_naming.md)

Key changes:
- Adds `templates` to v2 span names: a list of patterns with literals and placeholders for attributes
- Validates placeholder syntax, attribute presence on span in rust code, sampling-relevance
- allowed types (strings, ints), exhaustiveness is for rego policies

Example:
```yaml
file_format: "definition/2"

spans:
  - type: http.client.request
    kind: client
    brief: Client HTTP span
    name:
      templates:
        - pattern: "{http.request.method} {url.template}"
        - pattern: "{http.request.method}"
        - pattern: "HTTP"   # can always fallback to span type in codegen
    attributes:
      - ref: http.request.method
      - ref: url.template
```

this covers pretty much all spans we have so far in semconv and semconv-genai.

It's still possible to set `note` - free form text about span name. Codegen, when it does not see templates, would need to generate helpers that require span name. Otherwise codegen can generate it from start-time attirbutes.

### DIFF
--- a/docs/specs/span-naming/span_naming.md
+++ b/docs/specs/span-naming/span_naming.md
@@ -1,0 +1,284 @@
+# Span Name Templates
+
+Status: Proposal
+
+## Introduction
+
+Semantic conventions currently describe span names via free-form text in the `note` property or in external markdown documentation. This document specifies a formal, declarative mechanism in the Weaver semantic convention schema to define span names using ordered templates.
+
+## Analysis of Existing Conventions
+
+A review of span naming across semantic conventions reveals consistent patterns across domains:
+
+1. **GenAI** (including MCP): Consistently defines span names as an ordered sequence of patterns based on attribute availability:
+   - `{gen_ai.operation.name} {gen_ai.request.model}`
+   - `{gen_ai.operation.name} {gen_ai.data_source.id}`
+   - `{gen_ai.operation.name}`
+   - `create_agent {gen_ai.agent.name}`
+   - `invoke_agent {gen_ai.agent.name}` -> `invoke_agent`
+   - `execute_tool {gen_ai.tool.name} {gen_ai.skill.name} {gen_ai.skill.resource.name}` -> `execute_tool {gen_ai.tool.name} {gen_ai.skill.resource.name}` -> `execute_tool {gen_ai.tool.name} {process.executable.name}` -> `execute_tool {gen_ai.tool.name}`
+   - `invoke_workflow {gen_ai.workflow.name}`
+   - `plan {gen_ai.agent.name}` -> `plan`
+   - `{mcp.method.name} {gen_ai.tool.name}` -> `{mcp.method.name} {gen_ai.prompt.name}` -> `{mcp.method.name}`
+
+2. **HTTP**:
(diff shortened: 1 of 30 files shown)


docs/specs/span-naming/span_naming.md
```diff
@@ -0,0 +1,284 @@
...
+```
+
+And for HTTP client:
+
+```yaml
+spans:
+  - type: http.client
+    kind: client
+    name:
+      templates:
```
lmolkova, line 79: @open-telemetry/specs-semconv-maintainers want to get your thoughts. 

It's naive, but simple and covers pretty much all we have in semconv today.

docs/specs/span-naming/span_naming.md
```diff
@@ -0,0 +1,284 @@
...
+
+When `templates` are not specified on a span, code generation cannot infer how to construct the span name automatically. In this case, codegen makes the span name a required parameter that the caller must supply explicitly (e.g. `start_span(name: &str, ...)`).
+
+### 4. Live-Check Validation (Planned)
+
+In `weaver_live_check` (planned for follow-up PR), validating incoming spans against the formal specification:
+1. Find the first template in `templates` where all attributes are present on the sample span (and not equal to `_OTHER` for enum attributes).
+2. Format that template with the sample's attribute values -> `expected_name`.
+3. If a template matched: verify `sample.name == expected_name`.
+4. If mismatched, emit an advisory with actual vs expected span name.
```
jerbly, line 167:
```suggestion
4. If mismatched, find the next template in `templates` where all attributes are present on the sample span (and not equal to `_OTHER` for enum attributes) and goto (2)
5. if all templates are exhausted, emit an advisory with actual vs expected span name.
```

Essentially, we need to walk down the templates matching each one. We may have attrs A,B and C on the sample but the templates in priority order could be:
- {A}
- {A} {B}
- {A} {C} {B}
- {A} {B} {C}

The sample's name could be any one of those combos so we don't want to fail on the first one.

lmolkova: updated the text here - since order is important, if A is present, we should pick A regardless of other attributes being present. 

So it's not the best or the most complete of, it's the first in the order of priority and priority is defined by semantic conventions authors who should probably put `{A} {B} {C}` higher than `{A}`.

jsuereth (approved):

jerbly (approved):

Issue #565, opened 2025-01-22
Support user-provided JQ modules


goedelsoup:
**Is your feature request related to a problem? Please describe.**

I frequently have to replicate filters between numerous Weaver configurations in order to support providing our customers with a consistent SDK across languages. It would be great to consolidate the filtering logic alongside my registry definitions and templates.

**Describe the solution you'd like**

Forge currently loads a [default module](../tree/main/defaults/jq/semconv.jq) while provides the helper functions authors would typically use for their template generation. It would be great if we could supply additional modules to load via the `weaver generate` command.

**Describe alternatives you've considered**

Many warts encountered when dealing with deprecated elements could be worked around by adding a `exclude_attribute` and `exclude_metric` option on the relevant functions, however, user-defined JQ modules would provide better composition long-term under the multi-registry support model.

lquerel:
Supporting the addition of a JQ prelude alongside the one already in place doesn’t seem very complicated. We could leverage 1) the existing override mechanism used for building the weaver.yaml configuration (see this [doc](https://github.com/open-telemetry/weaver/blob/main/docs/weaver-config.md#configuration-file-loading-order-and-overriding-rules)), and 2) the introduction of a new field in the WeaverConfig structure that allows adding one or more JQ prelude files. This would enable the predefined prelude to be supplemented with one or more custom JQ preludes shared across multiple template directories.

Question: Would you prefer to have the option to completely bypass the default JQ prelude, or would the ability to add additional preludes be sufficient?

If someone is interested in implementing this request, I can provide more details.

goedelsoup:
My initial idea was just to be able to supply modules in addition to the prelude, as they utilize some of the prelude functions. I personally don't see any reasonable need to _exclude_ the prelude; even in scenarios where one may want to lint/validate that only certain functions were used, I'd imagine that'd be a positive assertion.

PR #1768, opened 2026-09-14
Support user-provided JQ modules in filters and conditions

LJimmy007:
Addresses #565 by allowing template packages and projects to share JQ functions through `jq_modules`. Definitions are available to filters and `when`, alongside the built-in helpers.

```yaml
# weaver.yaml
jq_modules: [jq/helpers.jq]
templates:
  - template: output.md.j2
    filter: custom_attributes
    when: should_generate
    application_mode: single
```

Relative paths resolve against the declaring configuration; nested imports resolve against the importing file. Diagnostics include module locations. Unconfigured filters retain disabled file loading.

**Compatibility and trust:** Existing `FilterErrorDetail` literals and no-module error JSON are preserved. Rust callers constructing `TemplateConfig` exhaustively must supply the new field or use `..Default::default()`. Modules are trusted code: absolute paths and `..` are allowed, with no filesystem sandbox. Modules reload and compile on each evaluation.

**Local validation:** Windows 11, Rust 1.96.1 GNU, based on `e95314d684f8425094baf3ec736fa9fac7588d56`. Focused config/Forge/API tests and CLI generation pass, as do formatting, focused strict Clippy, Rustdoc, validation and 20 history checks. Full workspace: 1,227 passed, one existing Windows symlink-fixture failure, two ignored. Full Clippy encounters upstream's Windows `Sighup` warning. The complete `just` chain and cross-platform matrix were not run; GitHub CI remains to be evaluated.

**Performance:** Identical release builds/dependencies, Ryzen 9 7945HX, 21 measured batches after warmup. Small no-module calls: 1.466 → 1.599 ms median (+8.66% paired); 10,000-attribute calls: 58.047 → 57.939 ms, effectively unchanged. Candidate-only medians: one module 1.908 ms, eight modules 3.420 ms, repeated module condition 1.945 ms. These are synthetic JQ API measurements, not end-to-end generation results.

Implementation and verification used AI assistance.


jsuereth:
This looks really cool.

One big ask though - Can you make sure there's some kind of integration test that shows *pulling* a JQ module using the code you have from a package?  I see there's a lot of unit tests - but I'd like to see at least one integration test where a directory is in as close to a "template with JQ module" as possible.

LJimmy007:
Thanks! I added a real template-package fixture containing weaver.yaml, a JQ module, and a template. The integration test runs weaver registry generate from outside the package directory and verifies that module-defined functions work in both the filter and when paths. The filter also uses Weaver’s built-in semconv_attributes helper.

I also updated the Windows-sensitive provenance assertion to check the full canonical missing-module path.

Commit: 7f5dde91370e29fdaf9aa8e0ac629f9a580f2179

The new workflows are currently awaiting approval before any jobs can run. Could you approve them so the Windows, Ubuntu, and macOS checks can execute?

PR #1769, opened 2026-09-14
Cache Git registries on disk across invocations

### LINKED ISSUE #1636: Cache version-pinned Git registries on disk across invocations

Every `weaver registry check` / `resolve` / `live-check` re-clones a remote Git registry — and every remote `dependencies[]` entry of a manifest — into a throwaway temporary directory that is deleted when the command exits. There is no persistent, ref-keyed cache: `VirtualDirectory::create_tmp_repo` (`crates/weaver_common/src/vdir.rs`) creates a fresh `tempfile::TempDir` per call and removes it on drop. (The in-process resolver cache from #1504 helps within a single run, but nothing persists across invocations.)

This is fine for one interactive run, but costly when many invocations resolve the *same* pinned registry:

- **CI live-check**: running `weaver registry live-check` in one container per integration-test suite pays a full cold clone of `semantic-conventions` in every container before any validation — few seconds each, ×N suites.
- **Codegen fleets**: repos that `registry generate` against upstream semconv re-clone on every generate.

We ran to this In [OBI](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation) integration tests. and ended up solving it by:

- [scripts/fetch-upstream-semconv.sh](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/scripts/fetch-upstream-semconv.sh) pre-downloads the pinned semconv `model/` tree into `schemas/obi/.deps/upstream-v<ver>/model`
- `schemas/obi/manifest.yaml` is rewritten so its dependency reads `registry_path: .deps/upstream-v1.41.0[model]` (a local path) instead of the canonical semconv Git URL — the script even has to cope with the manifest pointing at *either* form.
- CI stages `.deps` in a prep job and restores it into every sharded test job so weaver never hits the network.

With this feature OBI could delete the fetch script and the CI `.deps` plumbing, restore the manifest to the self-describing `…/semantic-conventions.git@v1.41.0[model]`, and just pass `--registry-cache-dir <cached-dir>` (plus `--offline` in CI). The cache dir becomes an ordinary `actions/cache` keyed on the pinned version.

An **opt-in, ref-keyed on-disk cache** for version-pinned Git registries, so a pinned registry is cloned once and reused by later invocations.

- **Only pinned sources are cached.** A source with an explicit `@<refspec>` (tag, branch, or commit) is cacheable; a bare URL tracking a moving default branch is never cached, so stale content can't be silently served.
- **Keyed by `(url, refspec)`.** Registries that differ only by `[sub_folder]` share one clone. A stable, platform/toolchain-independent hash keeps the key reproducible so it doubles as a CI cache key.
- **Concurrency-safe population.** The clone is staged in a private directory, then atomically `rename`d into place — parallel weaver processes never observe a half-populated entry, and a lost race just reuses the winner's clone.
- **Configured via global CLI flags**, following the existing convention for cross-cutting toggles (`--allow-git-credentials`, `--future`: global flags bridged to `weaver_common` at startup):
  - `--registry-cache-dir <PATH>` — cache root; providing it enables the cache. Omitting it keeps today's throwaway-clone behavior (fully backward compatible).
  - `--offline` — a cache miss for a pinned source is a hard error instead of a network fetch (deterministic CI).
  - `--registry-cache-refresh` — re-fetch and atomically replace a cached entry even on a hit (for a moving tag/branch); ignored under `--offline`.

```bash
weaver --registry-cache-dir /path/to/cache \
  registry check -r "https://github.com/open-telemetry/semantic-conventions.git@v1.41.0[model]"
weaver --registry-cache-dir /path/to/cache --offline \
  registry live-check --registry "https://github.com/open-telemetry/semantic-conventions.git@v1.41.0[model]"
```

The OBI solution

**Additional context**

Scope: limited to the Git source path in the virtual-directory layer; local folders/archives and remote archives are unaffected.

If this change is wanted I would love to work on this

NimrodAvni78:
Fixes https://github.com/open-telemetry/weaver/issues/1636

## Summary

Every `registry check` / `resolve` / `live-check` re-clones a remote Git registry into a temp dir and throws it away on exit. This PR adds an opt-in on-disk cache, so a registry pinned to immutable content is cloned once and reused.

Example usecase: CI running one weaver container per test suite, each cold-cloning semantic-conventions before it can validate anything. Locally, a pinned `registry check` drops from 4.2s to 0.18s on the second run.

## Usage

The flags are global, and the cache stays off unless you pass a directory.

| Flag | Effect |
| --- | --- |
| `--registry-cache-dir <PATH>` | Cache root. Providing it enables the cache. |
| `--registry-cache-offline` | A cache miss is an error instead of a fetch. Requires `--registry-cache-dir`. |
| `--registry-cache-refresh` | Re-fetch and replace an entry even on a hit. Requires `--registry-cache-dir`. |

```bash
weaver --registry-cache-dir /path/to/cache \
  registry check -r "https://github.com/open-telemetry/semantic-conventions.git@v1.41.0[model]"
```

## What gets cached

Only a refspec that resolves to immutable content: a commit SHA or a tag. A branch, or a URL with no refspec, is re-cloned every run, so the cache cannot serve a stale snapshot of a moving ref. Entries are keyed by `(url, refspec)`, so registries differing only by sub-folder share one clone.

Population is concurrency-safe — staged in a private directory, then atomically renamed — so parallel processes never see a half-written entry. There is no lock, though: on a cold cache all N processes still clone, and the race is settled after the fact.

Two things worth knowing. `--registry-cache-offline` covers this cache only; remote archives, remote files, and a `dependencies[]` entry resolved from its `schema_url` are still downloaded. And a hit is served without verifying its contents, so point the cache somewhere only your own builds can write — weaver executes any `*.rego` policy it finds in a resolved registry.

## Compatibility and testing

Without `--registry-cache-dir` nothing changes: git sources clone into a throwaway temp dir as before. The new public API is two variants on the already-`#[non_exhaustive]` `Error` enum, plus the flags.

Workspace tests, clippy and fmt are clean, and `docs/usage.md` is regenerated. The new tests drive the cache dispatch itself, so a branch or unpinned source that failed to bypass the cache would fail the suite.


crates/weaver_common/src/vdir.rs
```diff
@@ -294,6 +294,128 @@ fn is_commit_sha(s: &str) -> bool {
...
+
+/// Derives a filesystem-safe cache directory name for a pinned Git source.
+///
+/// The key is a function of `(url, refspec)` only — the sub-folder is applied
+/// after checkout, so registries that differ only by sub-folder share one clone.
+/// A short human-readable slug is prefixed for debuggability; a 64-bit FNV-1a
+/// hash (rendered as hex) guarantees uniqueness and is stable across platforms
+/// and toolchain versions, so the same key is reproducible in CI caches.
+fn git_cache_key(url: &str, refspec: &str) -> String {
+    const FNV_OFFSET: u64 = 0xcbf2_9ce4_8422_2325;
```
jsuereth, line 356: If you want specific hashing - make it clear you're computing a hash.

This looks super suspicous as -is, just use an existing hash library.

crates/weaver_common/src/vdir.rs
```diff
@@ -294,6 +294,128 @@ fn is_commit_sha(s: &str) -> bool {
...
+            .trim_end_matches(".git"),
+    );
+    let ref_slug = sanitize(refspec);
+
+    format!("{repo_slug}-{ref_slug}-{hash:016x}")
+}
+
+/// Strips a `user[:password]@` component from `url`, returning `None` when the
+/// URL carries no credentials.
+fn url_without_userinfo(url: &str) -> Option<String> {
```
jsuereth, line 395: We already havea URL library if you want to use it here instead of re-inventing.

crates/weaver_common/src/vdir.rs
```diff
@@ -652,33 +774,233 @@ impl VirtualDirectory {
...
         vdir_path: String,
+    ) -> Result<Self, Error> {
+        Self::try_from_git_url_with_cache(url, sub_folder, refspec, vdir_path, &git_cache_config())
+    }
+
+    /// Dispatches a Git source to the cache or to a throwaway clone.
+    ///
+    /// A source without a refspec tracks the remote default branch, so it is
+    /// never cacheable and never consults `cache`.
+    fn try_from_git_url_with_cache(
```
jsuereth, line 800: Why not jut have `tmp`/`cache` be the *same* thing - just in one case we create a tmp directory for cache, in the other we use a stable directory?

crates/weaver_common/src/vdir.rs
NimrodAvni78: Switched to the SHA-1 hasher from gix which is already included

crates/weaver_common/src/vdir.rs
NimrodAvni78: Removed it. seems like it wasn't even used (the registry parser splits at the first @, so a URL with user:token@ never gets as far as the clone)

crates/weaver_common/src/vdir.rs
NimrodAvni78: Done: there's now one handling for both, returning either a cache entry or a temp dir.

